PR #157, opened 2020-06-17
Use consistent marker of inaccessible ACLs

Vinnl:
A LitDataset can have an `acl` object attached with the keys
`resourceAcl` and `fallbackAcl`, which are undefined and null
respectively if they are not found, the cause of which is usually
lack of Control access for the current user.

That will also result in the original Resource not even advertising
an ACL. In that case, the `acl` object as a whole was set to
`null`, meaning that the result could no longer be passed to
getAgentAccessModesOne/All without casting, even though that is
perfectly capable of dealing with inaccessible ACLs (it will just
return null).

(Specifically, I want Pod Manager to be able to get rid of [this cast](https://github.com/inrupt/pod-manager/blob/7e3ae1303587b637c5798c5cf69964f3a178cf4b/src/lit-solid-helpers/index.ts#L223).)

So now, if we can deduce from a Resource's metadata that neither
ACL file is going to be accessible anyway, we mark that in the same
way as we would mark it if we had tried to fetch those files and
_then_ found out that they were inaccessible.

### DIFF
--- a/src/litDataset.test.ts
+++ b/src/litDataset.test.ts
@@ -346,7 +346,7 @@ describe("fetchLitDatasetWithAcl", () => {
     ]);
   });
 
-  it("returns null for the ACL if the fetched Resource does not include a pointer to an ACL file", async () => {
+  it("does not attempt to fetch ACLs if the fetched Resource does not include a pointer to an ACL file, and sets an appropriate default value.", async () => {
     const mockFetch = jest.fn(window.fetch);
 
     mockFetch.mockReturnValueOnce(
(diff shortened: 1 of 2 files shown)


Vinnl:
I verified using 

    npm install @solid/lit-pod@fixacl-status

that this does properly allow me to remove the cast in Pod Manager.

src/litDataset.test.ts
```diff
@@ -346,7 +346,7 @@ describe("fetchLitDatasetWithAcl", () => {
     ]);
   });
 
-  it("returns null for the ACL if the fetched Resource does not include a pointer to an ACL file", async () => {
+  it("does not attempt to fetch ACLs if the fetched Resource does not include a pointer to an ACL file", async () => {
```
NSeydoux, line 349:
```suggestion
  it("does not attempt to fetch ACLs if the fetched Resource does not include a pointer to an ACL file, and sets an appropriate default value.", async () => {
```

PR #166, opened 2020-06-17
Enable reading resource metadata without fetching the resource

NSeydoux:
# New feature description

This makes public the function previously called `internal_fetchResourceInfo`. This functions returns metadata associated to a resource (e.g. IRI, ACL), be it a LitDataset or a plain file.

In particular, you'll notice that `ResourceInfo` was renamed `ResourceWithInfo`, because otherwise the type returned by `fetchResourceInfo` required to be read in the following manner: 
```typescript
const myInfo = fetchResourceInfo(url);
const info = myInfo.resourceInfo;
```
which is a little counter-intuitive.

# Checklist

- [ ] Relevant documentation, if any, has been written/updated. (working on this prompted a cascade of modifications, the documentation should be updated in upcoming PRs)


CHANGELOG.md
```diff
@@ -14,3 +14,4 @@ First release! What's possible with this first release:
 - Manipulate data in datasets.
 - Inspect a user's permissions w.r.t. a given Resource. (Experimental.)
 - Retrieve, delete and/or write any file (including non-RDF) from/to a Pod.
+- Fetch information about a Resource (file or dataset) without fetching the resource itself
```
Vinnl, line 17: I don't mind too much, but given that this is an initial release, I think I'd just lump this under "Manipulate data in datasets" (or maybe make that "Read and manipulate").

.prettierignore
```diff
@@ -1 +1,2 @@
-resources/
\ No newline at end of file
+resources/
+*.snap
```
Vinnl, line 2: Hmm, what happens without this? I thought Prettier didn't operate on snap files in the first place?

src/interfaces.ts
```diff
@@ -104,32 +104,34 @@ type unstable_WacAllow = {
...
-     * @see https://github.com/solid/solid-spec/blob/cb1373a369398d561b909009bd0e5a8c3fec953b/api-rest.md#wac-allow-headers
-     * @see https://github.com/solid/specification/issues/171
-     */
-    unstable_permissions?: unstable_WacAllow;
-  };
+export type ResourceWithInfo = {
+  resourceInfo: ResourceInfo;
 };
 
+export type ResourceInfo = {
```
Vinnl, line 115: For everything that we `export`, we need to consider whether we want to expose it to developers. If we do, we should also export it from `index.ts`; if not, we need to explicitly exclude it from the API docs using `@hidden` or `@ignore`.

I'm not actually sure whether we want to hide it from developers; it could be useful when they're e.g. mocking Resources in tests? Then again, `ResourceWithInfo` already _is_ public.

```suggestion
/** @hidden Developers shouldn't create this object manually. */
export type ResourceInfo = {
```

By the way, in case you didn't know: as an alternative to splitting this out, the return type of `fetchResourceInfo` could also have been `Promise<ResourceWithInfo['resourceInfo']>`. Whatever you prefer :)

src/litDataset.ts
```diff
@@ -84,15 +85,20 @@ export async function fetchLitDataset(
...
 }
 
 /**
- * @internal
+ * Retrieve the information about a resource (e.g. access permissions) without
+ * fetching the resource itself.
+ *
+ * @param url URL to fetch a [[LitDataset]] from.
+ * @param options Optional parameter `options.fetch`: An alternative `fetch` function to make the HTTP request, compatible with the browser-native [fetch API](https://developer.mozilla.org/docs/Web/API/WindowOrWorkerGlobalScope/fetch#parameters).
+ * @returns Promise resolving to a [[LitDataset]] containing the data at the given Resource, or rejecting if fetching it failed.
```
Vinnl, line 99: This doesn't seem right - it doesn't return a LitDataset, right?

```suggestion
 * @param url URL to fetch Resource metadata from.
 * @param options Optional parameter `options.fetch`: An alternative `fetch` function to make the HTTP request, compatible with the browser-native [fetch API](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/fetch#Parameters).
 * @returns Promise resolving to the metadata describing the given Resource, or rejecting if fetching it failed.
```

(Btw, one other error that I sneaked in earlier: the "Parameters" heading on the MDN docs page has an ID of `#Parameters`, not `#parameters`. Don't know how that got in my browser history. Also, it'll likely be called otherwise in different languages, so I did include the locale in the URL here - though I'm not sure whether it's better to end up at the top of the page in your own language, or at the right section in English...)

src/litDataset.ts
```diff
@@ -254,16 +262,17 @@ export async function saveLitDatasetAt(
...
-      changeLog: { additions: [], deletions: [] },
-      resourceInfo: resourceInfo,
-    }
-  );
+  const storedDataset: LitDataset &
+    ChangeLog &
+    ResourceWithInfo = Object.assign(litDataset, {
+    changeLog: { additions: [], deletions: [] },
+    resourceInfo: resourceInfo,
+  });
```
Vinnl, line 275: Go home Prettier, you're drunk.

.prettierignore
NSeydoux: I initially experienced a prettier failure that looked like it was caused by this, but actually that did not fix it. I removed this in 962df4e313e82526e937a111f987c975ba85da91

src/litDataset.ts
NSeydoux: I'd say the right section is more appropriate, and our whole documentation is in english anyways, so including the locale doesn't seem problematic.

CHANGELOG.md
NSeydoux: Resolved in 49a62169510b1fb6c2cacb3a544d10b73d401be3

src/interfaces.ts
NSeydoux: I wasn't familiar with the `ResourceWithInfo['resourceInfo']` notation. I saw you using it a bunch of times, but not having used it myself I didn't think about it. It seems to be the simplest in this case, instead of exporting two very similar types

NSeydoux:
I also changed `ResourceWithInfo` to `WithResourceInfo` (and the same for `WithChangelog`), I agree it reads better.

Issue #152, opened 2020-06-17
Add ability to get parsed file metadata


Vinnl:
**Search terms you've used**

File metadata, ACL, resource ACL.

**Is your feature request related to a problem? Please describe.**
If I want to know what permissions the user has w.r.t. a plain file, I currently have to manually parse the Response Headers. Whereas when I fetch an RDF resource, I can use both the ACL-reading methods, or read the parsed WAC-Allow statements.

**Describe the solution you'd like**
For `getFile` to return an object that has the blob as a property, but also `LitDatasetInfo`.

**Describe alternatives you've considered**
A separate `fetchFileInfo` (or perhaps `fetchResourceInfo`) would also work, that does a `HEAD` request (and possibly `fetchFileInfoWithACL`). This could also work for Resource of which we don't need to inspect the contents.

**Additional context**
See [this PR](https://github.com/inrupt/pod-manager/pull/18/files#diff-3ac371871dc7cf96a9cf19e4c3868e09R161) of Pod Manager where they had to write the WAC-Allow parsing code themselves, meaning they'll now manually have to keep up with the spec. ([Direct link to source code.](https://github.com/inrupt/pod-manager/blob/7e3ae1303587b637c5798c5cf69964f3a178cf4b/src/lit-solid-helpers/index.ts#L161-L195))

Vinnl:
Note also that we already have `internal_fetchDatasetInfo` (`fetchResourceInfo` after #156), so we could just expose that.

Vinnl:
Release 0.4.0 [introduced `getResourceInfo`](https://github.com/inrupt/solid-client-js/blob/main/CHANGELOG.md#040---2020-09-15).

PR #167, opened 2020-06-17
Functions reading access permissions no longer require the Resource

### LINKED ISSUE #164: Make it possible to read access permissions without fetching a Resource
**Search terms you've used**

ACL, Resource, HEAD

**Is your feature request related to a problem? Please describe.**

Currently, the functions reading access permissions expect a resource as a parameter (e.g. 
```typescript
unstable_getAgentAccessModesOne(
  dataset: LitDataset & ResourceWithInfo & unstable_Acl,
  agent: WebId
): unstable_AccessModes | null
```
while really only the resource information and the ACL is required here, and not the Resource itself. 

Make it unnecessary to provide a Resource (be it a `Blob` or a `LitDataset`) when reading access permissions.

Mock up the Resource: `unstable_getAgentAccessModesOne({acl: myAcl, resourceInfo: MyResourceInfo}`.

**Additional context**

Now that we have implemented features to access only resource information, without the resource itself, these functions should evolve to support new use cases.

NSeydoux:
Fixes #164 

Note that the type signature of `unstable_getAgentAccessModesOne` and `unstable_getAgentAccessModesAll` changed to only expect the ACL, as opposed to the resource with the ACL attached. The `ResourceInfo` (e.g. origin and ACL IRI) also have been removed from the type signature of these functions, since only the ACL data is required. If they had been added initially for future-proofing and in anticipation of API changes, 686585af4fbc48bdb47bd8253e4cc8fd6b283b77 should be reverted.

# Checklist

- [x] Relevant documentation, if any, has been written/updated. This piece of documentation will likely evolve with #152. 


src/acl.ts
```diff
@@ -126,97 +126,97 @@ function getContainerPath(resourcePath: string): string {
...
  *
- * @param dataset A [[LitDataset]] that might have an ACL attached.
+ * @param acl A Resource that might have an ACL attached.
  * @returns Whether `dataset` has an ACL attached.
  */
-export function unstable_hasResourceAcl<Dataset extends unstable_Acl>(
-  dataset: Dataset
-): dataset is Dataset & {
-  acl: { resourceAcl: Exclude<unstable_Acl["acl"]["resourceAcl"], undefined> };
+export function unstable_hasResourceAcl<Acl extends unstable_WithAcl["acl"]>(
```
Vinnl, line 141: It kinda undermines your entire PR (sorry!), but... I think I'd feel more comfortable keeping this function as accepting the full Resource, i.e.:

```suggestion
export function unstable_hasResourceAcl<ResourceWithAcl extends unstable_WithAcl>(
```

The reason for this is that generally, I _think_ we don't want to require developers to first extract the ACL file from the Resource, and only _then_ be able to work with it. I'm afraid that that will lead to situations where developers are going to be required to pass _both_ around all over the place.

I'm not 100% sure about this - it might very well be that, once we get to writing ACLs, there's no avoiding having the developer manually set it. But I'd prefer to postpone this refactor until we really can no longer avoid it, and until then try to maintain an API where the external-facing methods just operate on the Resource itself.

I suppose I should have added an E2E test for just the reading of ACLs already... I'll make sure I'll do that in for the "writing" PR.

src/interfaces.ts
```diff
@@ -144,7 +144,7 @@ export type WithChangeLog = {
 /**
  * @hidden Developers should use [[unstable_getResourceAcl]] and [[unstable_getFallbackAcl]] to access these.
  */
-export type unstable_Acl = {
+export type unstable_WithAcl = {
   acl: {
     resourceAcl?: unstable_AclDataset;
```
Vinnl, line 149: I saw you changed this to use `null` instead of `undefined` elsewhere? I initially did it this way because it's likely that the Resource ACL does not exist, whereas the fallback ACL should always exist - but since it's perfectly possible that the fallback ACL is not accessible to the current user, this definitely also makes more sense:

```suggestion
    resourceAcl: unstable_AclDataset | null;
```

src/acl/agent.ts
```diff
@@ -37,59 +37,59 @@ import {
...
 /**
- * Find out what Access Modes have been granted to a given Agent specifically for a given LitDataset.
+ * Find out what Access Modes have been granted to a given Agent specifically for a given Resource.
  *
  * Keep in mind that this function will not tell you what access the given Agent has through other ACL rules, e.g. public or group-specific permissions.
  *
  * Also, please note that this function is still experimental: its API can change in non-major releases.
  *
- * @param dataset The LitDataset to which the given Agent may have been granted access.
+ * @param resourceInfo Informations about the Resource to which Agents may have been granted access.
```
Vinnl, line 52:
```suggestion
 * @param resourceInfo Information about the Resource to which the given Agent may have been granted access.
```

(Why did you change this to "Agent_s_" - this function operates on a single Agent, right?

src/acl/agent.ts
```diff
@@ -37,59 +37,59 @@ import {
...
 /**
- * Find out what Access Modes have been granted to specific Agents for a given LitDataset.
+ * Find out what Access Modes have been granted to specific Agents for a given Resource.
  *
  * Keep in mind that this function will not tell you what access arbitrary Agents might have through other ACL rules, e.g. public or group-specific permissions.
  *
  * Also, please note that this function is still experimental: its API can change in non-major releases.
  *
- * @param dataset The LitDataset to which Agents may have been granted access.
+ * @param resourceInfo Informations about the Resource to which Agents may have been granted access.
```
Vinnl, line 82:
```suggestion
 * @param resourceInfo Information about the Resource to which Agents may have been granted access.
```

Is "information" countable in French? I think I've seen this more often with French speakers :P

website/docs/tutorials/managing-access.md
```diff
@@ -81,7 +95,7 @@ import {
 const litDatasetWithAcl = await unstable_fetchLitDatasetWithAcl(
   "https://example.com"
 );
-const accessByAgent = unstable_getAgentAccessModesAll(litDatasetWithAcl);
+const accessByAgent = unstable_getAgentAccessModesAll(litDatasetWithAcl.acl);
```
Vinnl, line 98: So if we do this, we lose the ability to eventually make the ACL data part of the Resource itself. Not sure if that's desirable... (As in, I think Pat might have something to say about that :P )

website/docs/tutorials/managing-access.md
```diff
@@ -44,6 +44,17 @@ A Resource's or Container's ACL is only accessible to your app if the following
 Access can be granted to individual [agents](../glossary#agent), to groups, or even to everyone.
 We currently only support reading what access has been granted to individual agents specifically.
 
+### Fetching access information
```
Vinnl, line 47: OK, with this, I think I see where you're coming from: it feels weird to just fetch the metadata, only to have it wrapped in an object with a single property.

That said, if we want to be able to represent that metadata as if it's part of the dataset itself in the future, then that property shouldn't be exposed to the developer anyway - they should be accessing it through our functions.

website/docs/tutorials/managing-access.md
```diff
@@ -44,6 +44,17 @@ A Resource's or Container's ACL is only accessible to your app if the following
...
+
+There are two main ways of checking the access information of a Resource:
+
+1. Fetch the entire Resource, data and information at once. To do so, use the functions
+   `unstable_fetchFile` or `unstable_fetchLitDatasetWithAcl`. The returned value includes both the Resource
+   data (e.g. your profile or friend list), the `ResourceInfo`, and the ACL containing the associated
+   access information.
+2. Fetch only the information about a Resource, without fetching the Resource itself
+   with `fetchResourceInfo`, and then fetching the associated ACL (if any).
```
Vinnl, line 56: Also, we don't currently expose a function to explicitly fetch a Resource's ACL (because we're pretending it's part of the Resource). So maybe what we need to satisfy Pod Manager's use case, is an `unstable_fetchResourceInfoWithAcl` that does the same as `unstable_fetchLitDatasetWithAcl`, but without fetching the contents of the Resource itself.

(I also wonder how esoteric their use case is - i.e. do we even need to mention it in this tutorial? Maybe we should add some point in the future add an "Advanced" section that has tutorials such as "Traversing Containers in a Pod"?)

src/acl.ts
NSeydoux: I see your point. I was thinking about this change feature-wise, but design-wise, I think you are right. My motivation was that when you only fetch a resource's information and ACL, you don't have the resource itself, but rather either the resource info or the ACL, and I was concerned that these two being separate pieces of data, it would be burdensome to not take one of these as a parameter, instead of a Resource-like object with an `acl` or `resourceInfo` member. 

But I think that this could not be an issue if we design the functions to only fetch metadata accordingly. I'm not sure 
```typescript
async function fetchResourceInfo(
  url: UrlString,
  options: Partial<typeof defaultFetchOptions> = defaultFetchOptions
): Promise<WithResourceInfo["resourceInfo"]>
```
should change to return `Promise<WithResourceInfo>`, because `const myInfo = (await fetchResourceInfo("someUrl")).resourceInfo` doesn't look right (IMHO), but I wouldn't mind having 
```typescript
async function fetchResourceInfoAndAcl(
  url: UrlString,
  options: Partial<typeof defaultFetchOptions> = defaultFetchOptions
): Promise<WithResourceInfo & unstable_WithAcl>
```
With the latter, the fetched object can be passed to `unstable_hasResourceAcl`, so it still holds to the motivation for this PR, which is to make it easy to work without fetching the body of the resource

src/acl/agent.ts
NSeydoux: It is, actually both "information" and "data" are, so if I'm not extra careful this happens stuck_out_tongue

website/docs/tutorials/managing-access.md
NSeydoux: I should have read all your comments before answering to the first one, but yes, this is the motivation, but I agree that as long as our API is designed to return types that make sense and work well with each other, it should be okay, we just have to provide the right functions.

website/docs/tutorials/managing-access.md
NSeydoux: 100% agree on the need for `unstable_fetchResourceInfoWithAcl`. And good point, I don't expect this to be the most common use case, so it may be better not to mention it at all here.

src/acl/agent.ts
NSeydoux: That's an oversight, I'm not sure how I ended up rephrasing this, thanks for catching it.

src/acl/agent.ts
NSeydoux: Yup, both 'information' and 'data' are countable in French, so if I'm not careful this happens ^^

src/acl.ts
Vinnl: > `const myInfo = (await fetchResourceInfo("someUrl")).resourceInfo` doesn't look right (IMHO)

Agreed, so hopefully by keeping that an internal method we can sidestep this issue :P

(I also don't mind whether it's `fetchResourceInfoAndAcl` or `fetchResourceInfoWithAcl`, but make sure that it's consistent with `fetchLitDatasetWithAcl`, i.e. rename that if necessary.)

src/acl/agent.ts
Vinnl: Really, "data" is countable? I mean, in Dutch it can serve as the plural version of "date", in which case it _is_ countable, but for the concept of data that is really weird. I wonder what I'd get if I asked a French person for one information or one data :P

Vinnl:
To log it where it's relevant: the issue there was that `options` isn't passed along for `fetchAcl`, so it doesn't call the mock fetcher provided in the tests.

website/docs/tutorials/managing-access.md
NSeydoux: Fixed in b18586db675fc622a332bc4cb576712a0d675d75

src/litDataset.ts
```diff
@@ -115,10 +115,72 @@ export async function fetchResourceInfo(
 
   const resourceInfo = parseResourceInfo(response);
 
-  return resourceInfo;
+  return { resourceInfo };
```
Vinnl, line 118: Heh. I kinda expected this function to stay as-is, but I guess this does make sense given our earlier conversation. However, shouldn't `fetchAcl` then also return `unstable_WithAcl` rather than `unstable_WithAcl['acl']`?

(This is also a breaking change, but I assume Pod Manager isn't using this function yet.)

src/litDataset.ts
```diff
@@ -115,10 +115,72 @@ export async function fetchResourceInfo(
...
+    resourceAcl: resourceAcl,
+  };
+
+  return acl;
+}
+
+/**
+ * Experimental: fetch a LitDataset and its associated Access Control List.
+ *
+ * This is an experimental function that fetches both a Resource, the linked ACL Resource (if
```
Vinnl, line 155: Given that `fetchResourceInfo` does a `HEAD` request:

```suggestion
 * This is an experimental function that fetches both a Resource's metadata, the linked ACL Resource (if
```

src/litDataset.ts
```diff
@@ -115,10 +115,72 @@ export async function fetchResourceInfo(
...
+ * available), and the ACL that applies to it if the linked ACL Resource is not available. This can
```
Vinnl, line 156: Maybe good to make explicit that the fallback ACL can also be null (i.e. that the developer has to deal with that):

```suggestion
 * available), and the ACL that applies to it if the linked ACL Resource is not available (if accessible). This can
```

src/litDataset.ts
```diff
@@ -115,10 +115,72 @@ export async function fetchResourceInfo(
...
+ * result in many HTTP requests being executed, in lieu of the Solid spec mandating servers to
+ * provide this info in a single request. Therefore, and because this function is still
+ * experimental, prefer [[fetchLitDataset]] instead.
+ *
+ * If the Resource does not advertise the ACL Resource (because the authenticated user does not have
+ * access to it), the `acl` property in the returned value will be null. `acl.resourceAcl` will be
+ * undefined if the Resource's linked ACL Resource could not be fetched (because it does not exist),
+ * and `acl.fallbackAcl` will be null if the applicable Container's ACL is not accessible to the
+ * authenticated user.
```
Vinnl, line 165: Corrections: the `acl` object will no longer be null, and `acl.resourceAcl` will not be `undefined` but `null`. (And I guess it doesn't really matter to the developer whether the ACL was advertised or not, just whether they have access to it.)

```suggestion
 * If the Resource's linked ACL Resource could not be fetched (because it does not exist, or because
 * the authenticated user does not have access to it), `acl.resourceAcl` will be `null`. If the
 * applicable Container's ACL is not accessible to the authenticated user, `acl.fallbackAcl` will be
 * `null`.
```

src/litDataset.ts
```diff
@@ -115,10 +115,72 @@ export async function fetchResourceInfo(
...
+ *
+ * @param url URL of the LitDataset to fetch.
+ * @param options Optional parameter `options.fetch`: An alternative `fetch` function to make the HTTP request, compatible with the browser-native [fetch API](https://developer.mozilla.org/docs/Web/API/WindowOrWorkerGlobalScope/fetch#parameters).
+ * @returns A LitDataset and the ACLs that apply to it, if available to the authenticated user.
```
Vinnl, line 169:
```suggestion
 * @returns A Resource's metadata and the ACLs that apply to the Resource, if available to the authenticated user.
```

src/acl/agent.ts
```diff
@@ -37,59 +37,61 @@ import {
...
 export function unstable_getAgentAccessModesAll(
-  dataset: LitDataset & WithResourceInfo & unstable_Acl
+  resourceInfo: unstable_WithAcl
 ): unstable_AgentAccess | null {
-  if (unstable_hasResourceAcl(dataset)) {
-    const resourceAcl = unstable_getResourceAcl(dataset);
-    return unstable_getAgentResourceAccessModesAll(resourceAcl);
+  if (unstable_hasResourceAcl(resourceInfo)) {
+    return unstable_getAgentResourceAccessModesAll(
+      resourceInfo.acl.resourceAcl
```
Vinnl, line 90: Hmm, was there a specific reason to remove use of `getResourceAcl` here and `getFallbackAcl` on line 94? Right now they're simple wrappers to access a property, but if they ever do change to be part of the same Dataset, using those functions gives us a single place to update.

src/litDataset.ts
NSeydoux: I'm still quite hesitant. I'm really not a fan of `const resourceInfo = (await fetchResourceInfo).resourceInfo`, which the current proposal warrants. However, since all the functions of our interface expect a `WithAcl` rather than a `Acl`, `unstable_fetchResourceInfoWithAcl` returns a `WithResourceInfo & unstable_WithAcl`, and then for consistency it seems better than `unstable_fetchResourceInfo` returns a `WithResourceInfo`.

NSeydoux: But good point about `fetchAcl`: if it ever is no longer an internal function, it should be kept consistent with the API.

NSeydoux: Fixed in 89cb668f07cd0f0f8beb2d1fbbce91f3904f9279

src/acl/agent.ts
NSeydoux: No specific reason, I assume it was a overzealous cleanup when changing the return type. Fixed in 74af0a1a3c930ddb98887f848aa8648689581a0e

src/litDataset.ts
Vinnl: Yeah, I am more hesitant about `fetchResourceInfo` return a `WithResourceInfo` as well for some reason, though I'm finding it hard to justify why with actual arguments rather than a feeling of unease. (Especially since we're wrapping it in `{ resourceInfo }` in the only place we're using it anyway.)

I am wondering though: is `fetchResourceInfo` even relevant to consumers? Should we not mark it internal just like `fetchAcl`? Then we could also keep it as `WithResourceInfo['resourceInfo']` for now, until we can actually justify our feelings :P

NSeydoux: Well right now `fetchResourceInfo` doesn't do much, but are we sure we only want the API to expose the function `fetchResourceInfoAndAcl`, and not allowing lower-level `HEAD` requests that does not involve firing the additional `ACL` request ?

NSeydoux: Actually I can't think of an use-case that:
- Requires not to fetch the resource **AND**
- Requires to fetch the ACL link without fetching the ACL

Potentially some corner use-cases adding custom `Link` headers, but overall it seems fair to keep the API slim and focus on the use cases that have been expressed by actual users, namely "get the ACL without getting the resource". I'll make `fetchResourceInfo` internal for the time being.

Vinnl: Exactly. Plus, exposing something later is easier than un-exposing it :)

NSeydoux: Fixed in 00c31bdae5f8959f34e0c89a355377d6506e30df

PR #172, opened 2020-06-18
List npm deployments in the GitHub interface

Vinnl:
This adds a link to the npm page of the review package in the PR UI (see below - "This branch was successfully deployed", and the "View deployment" buttons).

Yes, this was kinda a diversion, but it's kinda useful too - especially if we're also going to do the same for the website.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -6,7 +6,7 @@ env:
   CI: true
 jobs:
   prepare-deployment:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
     outputs:
       tag-slug: ${{ steps.determine-npm-tag.outputs.tag-slug }}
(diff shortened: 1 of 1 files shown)


.github/workflows/cd.yml
```diff
@@ -5,31 +5,101 @@ on: push
 env:
   CI: true
 jobs:
+  prepare-deployment:
+    runs-on: ubuntu-18.04
```
pmcb55, line 9: I assume everything we do from now on should be on Ubuntu 20.04...?

Vinnl: It's still in preview, but yeah, it'll likely work fine and most contributors and servers will be on that version, so we can update it as we go. I'll change it here.

image

Vinnl: Fixed in 3cfeb03728a8c5a6b251c97c2fbc4df26f7f5644.

PR #176, opened 2020-06-19
Only check licences in CI

Vinnl:
The licences change infrequently, so I think it's best not to interrupt the
development process with a slow check that produces the same results
most of the time when all we need to do is make sure licences we
don't want do not make it into our dependencies before we release.


Vinnl:
Yep, that would be great, but I can't really find any way to do that other than the radio silence on [this StackOverflow question](https://stackoverflow.com/questions/50034679/hook-specifically-postinstall-after-any-npm-install-command) either...

NSeydoux:
https://xkcd.com/979/

PR #190, opened 2020-06-23
Attach the full LocalNode to a ThingLocal

Vinnl:
We previously attached just the `LocalNode`'s name, but that didn't
work for `getThingOne()` and `getThingAll()`: they would create a new
`LocalNode` and then try to match Quads from a Dataset on that
Subject, but because that new `LocalNode` was, in effect, a new
`BlankNode`, the relevant Quads did not match.

So the change here is that a `ThingLocal` now has a `localSubject` property attached to it, rather than just its `name` (where a `ThingPersisted` has its `iri`): an actual `BlankNode` instance that we can match on, with a `name` attached.

- [x] I've added a unit test to test for potential regressions of this bug: https://github.com/inrupt/lit-pod/pull/190/files#diff-1edc6ef23599ac668f6c25e183a097b2R993-R1007

### DIFF
--- a/src/thing.test.ts
+++ b/src/thing.test.ts
@@ -30,6 +30,7 @@ import {
   removeThing,
   createThing,
   asUrl,
+  toNode,
 } from "./thing";
 import {
   IriString,
@@ -39,6 +40,7 @@ import {
   LitDataset,
   WithResourceInfo,
   WithChangeLog,
+  LocalNode,
 } from "./interfaces";
 
 function getMockQuad(
@@ -69,15 +71,15 @@ describe("createThing", () => {
     const thing1: ThingLocal = createThing();
     const thing2: ThingLocal = createThing();
 
-    expect(typeof thing1.name).toBe("string");
-    expect(thing1.name.length).toBeGreaterThan(0);
-    expect(thing1.name).not.toEqual(thing2.name);
+    expect(typeof thing1.localSubject.name).toBe("string");
+    expect(thing1.localSubject.name.length).toBeGreaterThan(0);
+    expect(thing1.localSubject.name).not.toEqual(thing2.localSubject.name);
   });
 
   it("uses the given name, if any", () => {
     const thing: ThingLocal = createThing({ name: "some-name" });
 
-    expect(thing.name).toBe("some-name");
+    expect(thing.localSubject.name).toBe("some-name");
   });
 
   it("uses the given IRI, if any", () => {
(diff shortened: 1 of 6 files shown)


src/thing.test.ts
```diff
@@ -958,18 +968,40 @@ describe("asIri", () => {
...
+      DataFactory.blankNode("Arbitrary blank node"),
+      { name: "localSubject" }
+    );
+    const thing: ThingLocal = Object.assign(dataset(), {
+      localSubject: localSubject,
+    });
+    const node1 = toNode(thing);
+    const node2 = toNode(thing);
+    expect(node1.equals(node2)).toBe(true);
+    expect(node1).toEqual(node2);
```
NSeydoux, line 1005: Is there a difference between these two lines ?

src/thing.ts
```diff
@@ -247,7 +247,10 @@ export function createThing(options: CreateThingOptions = {}): Thing {
     return thing;
   }
   const name = (options as CreateThingLocalOptions).name ?? generateName();
-  const thing: ThingLocal = Object.assign(dataset(), { name: name });
+  const localSubject: LocalNode = getLocalNode(name);
+  const thing: ThingLocal = Object.assign(dataset(), {
+    localSubject: localSubject,
```
NSeydoux, line 252: For consistency, I've been using a destructuring notation (I think that's the name) to do similar things, e.g. here I would have written 
```typescript
const thing: ThingLocal = Object.assign(dataset(), { localSubject });
```
I imagine this falls under `https://eslint.org/docs/rules/prefer-destructuring`, and if so should we enforce this rule in our coding guidelines, one way or the other ?

src/thing.test.ts
Vinnl: Yep, the former uses the RDF/JS [`equals`](https://rdf.js.org/data-model-spec/#dom-term-equals) method, which checks that the Nodes are equivalent (i.e. have the same value and Node type), whereas the latter checks that they're actually the same instance.

(The latter does actually not need to be true... I'll remove it.)

src/thing.ts
Vinnl: Yeah I saw that; I've been _not_ doing destructing so far ;) I refrained from pointing that out because if we believe it's important to be consistent in that, it should be a lint rule. And I think that Jack's ESLint config actually contains a rule for that (IIRC it does it your way, though I'm not sure), but we haven't adopted that yet due to a couple of issues (logged in the config's repo, but that's still set to private). So this (or the other instances) will be fixed in due time.

src/thing.test.ts
Vinnl: Fixed in 6c4411e942711039716f8c6442908778ba8e9496.

PR #200, opened 2020-06-25
Implement saveAclFor()

Vinnl:
# New feature description

This adds the ability to save an ACL Resource back to the Pod, to associate it with the given Resource.

# Checklist

- [x] Relevant documentation, if any, has been written/updated. (API docs are there, tutorial update will follow later.)

### DIFF
--- a/src/litDataset.test.ts
+++ b/src/litDataset.test.ts
@@ -42,6 +42,7 @@ import {
   internal_fetchResourceInfo,
   createLitDataset,
   unstable_fetchResourceInfoWithAcl,
+  unstable_saveAclFor,
 } from "./litDataset";
 import {
   WithChangeLog,
@@ -49,6 +50,7 @@ import {
   IriString,
   LitDataset,
   LocalNode,
+  unstable_AclDataset,
 } from "./interfaces";
 
 function mockResponse(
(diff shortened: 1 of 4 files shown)


src/litDataset.test.ts
```diff
@@ -1657,3 +1659,150 @@ describe("saveLitDatasetInContainer", () => {
...
+    });
+
+    const savedAcl = await unstable_saveAclFor(withResourceInfo, aclResource, {
+      fetch: mockFetch,
+    });
+
+    expect(savedAcl.accessTo).toBe("https://some.pod/resource");
+  });
+
+  it("sends a PATCH if the ACL contains a ChangeLog and was originally fetched from the same location", async () => {
```
NSeydoux, line 1757: I think this is a non-issue, but I'm asking just to be sure: if the ACL does not exist in the first place (its IRI as advertised by the resource Link header returns a 404), is its `fetchedFrom` property set to this IRI anyways ? And if so, a `PATCH` can create a resource, right ?

Vinnl: I'm not sure if `PATCH` can create a Resource, but if it didn't exist, then this ACL will either have to have been created using `createLitDataset()` (in which case it will not have a `fetchedFrom`), or will have to have been based on a different Resource's ACL (in which case its `fetchedFrom` should be pointing to a different location), so it shouldn't be doing a `PATCH`.

src/litDataset.ts
```diff
@@ -408,6 +409,38 @@ export async function saveLitDatasetInContainer(
...
 
+/**
+ * Save the ACL for a Resource.
+ *
+ * @param resource The Resource to which the given ACL applies.
+ * @param resourceAcl An [[unstable_AclDataset]] whose ACL Rules will apply to `resource`.
+ * @param options Optional parameter `options.fetch`: An alternative `fetch` function to make the HTTP request, compatible with the browser-native [fetch API](https://developer.mozilla.org/docs/Web/API/WindowOrWorkerGlobalScope/fetch#parameters).
+ */
+export async function unstable_saveAclFor(
+  resource: WithResourceInfo & {
```
ajacksified, line 420: This type is a little hard to follow - with unions and excludes, at what point does it just make sense to create a new type (even if this is the implementation underneath)? Could be my unfamiliarity with Typescript, but it _feels_ like a smell if your type definition is 7 lines long.

Vinnl: Yep, that's fair - in fact, I think I did want to create a separate type alias for that after I finished it, but forgot. I'll do it in a separate PR, since, you know... It involves new names :P

Vinnl: https://github.com/inrupt/lit-pod/pull/206

PR #159, opened 2020-06-17
added introduction page

pmcb55:
# New feature description
Just added a suggested Introduction page (with links in the navbar, footer and a button beside 'Getting Started').

# Checklist

- [X] All acceptance criteria are met.
- [X] Relevant documentation, if any, has been written/updated.
- [ ] The changelog has been updated, if applicable.
- [ ] New functions/types have been exported in `index.ts`, if applicable.
- [X] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).

PR #210, opened 2020-06-29
Add an introductory guide

Vinnl:
This adds an introductory guide explaining the background and motivations for lit-pod. It's currently somewhat bare bones, partly because we have a couple of aspirational goals that will be listed here in due time but that we aren't supporting yet (because documentation should only refer to existing features), and partly because we'll want to be intentional about this and go over this together with Product - but at least there's a clear place to put this stuff. The guide is linked to from the header, the footer, and is the secondary call-to-action on the homepage.

This PR supersedes #159, which mostly involved our aspirational documentation. I've left TODOs indicating places where content based on that PR can later be added.

### DIFF
--- a/website/docs/guide/core-concepts.md
+++ b/website/docs/guide/core-concepts.md
@@ -1,0 +1,33 @@
+---
+id: core-concepts
+title: Core Concepts
+sidebar_label: Core Concepts
+---
+
+## Fetching data
+
+In general, working with data using lit-pod involves two steps: making a request to some web address (URL) to fetch an object containing all data at that address (a [`LitDataset`](../glossary#litdataset)), and then passing that object to functions to extract and manipulate sets of data from it that are relevant to you (which we call [`Thing`s](../glossary#thing)).
+
+<!--
+
+TODO:
+If the spec gets updated to support atomic fetching and updates of multiple Named Graphs at the same Resource,
+this section should be updated to clarify that; the slicing-and-dicing idea might make more sense with that context.
+
+-->
+
+For example, to read my name, you would first fetch a LitDataset from `https://vincentt.inrupt.net/profile/card`, the URL at which my personal information is stored. Then, from that LitDataset, you could extract the Thing that represents my profile. From that, you can then read my name.
+
+For a more extensive overview of fetching and manipulating data, see the tutorial [Working with Data](../tutorials/working-with-data).
+
+## Immutability
+
+The functions exposed in lit-pod are designed to take data as input, apply some transformation, and return the transformed data as output **without changing the input data**. That makes it easier for you to write unit tests, and enables frameworks that want to be notified of changes to data (like most modern front-end frameworks) to apply performance optimizations like memoisation, checking for updates by references rather than doing deep comparisons. However, it is good to be aware of this fact; when you're seeing stale data, that is likely the result of a variable pointing to an input value, rather than to the returned result.
+
+<!--
+
+TODO:
+Once we add a Fluent API, we should add a section here about gradually exposing advanced interfaces,
+add an "Advanced Usage" section to the documentation, and refer to that.
+
+-->
--- a/website/docs/guide/installation.md
+++ b/website/docs/guide/installation.md
@@ -1,0 +1,17 @@
+---
+id: installation
+title: Installation
+sidebar_label: Installation
+---
+
+To install the latest stable version of lit-pod, run:
+
+```bash
+npm install @solid/lit-pod
+```
+
+lit-pod will work in Node using CommonJS modules, and in the browser with a bundler like [Webpack](https://webpack.js.org), [Rollup](https://rollupjs.org) or [Parcel](https://parceljs.org).
+
+## To access private data
+
+By default, lit-will only enables to access public data on Solid Pods. To allow for access to restricted data, you will need to include a library that supports user authentication. If you have [solid-auth-client](https://www.npmjs.com/package/solid-auth-client) installed and have used it to authenticate the user, lit-pod will automatically pick up the authenticated session and include the user's credentials with each request.
--- a/website/docs/guide/motivation.md
+++ b/website/docs/guide/motivation.md
@@ -1,0 +1,37 @@
+---
+id: motivation
+title: Motivation
+sidebar_label: Motivation
+---
+
+[Solid](https://solidproject.org) is a wonderful project, but the technology it's built on —[RDF](https://en.wikipedia.org/wiki/Resource_Description_Framework)— is unfamiliar to many developers and has a relatively high barrier to entry.
+
+The main goal of lit-pod is to make it as easy as possible for engineers to get started writing apps both for the browser and for Node that store data on Solid Pods by providing an abstraction layer on top of both Solid and RDF principles, without hampering the ability to leverage the more advanced capabilities of RDF if so desired.
+
+## Goals
+
+To achieve this, lit-pod has the following major focus areas.
+
+### Documentation
+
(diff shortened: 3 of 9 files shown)


website/docs/guide/core-concepts.md
```diff
@@ -0,0 +1,33 @@
...
+
+<!--
+
+TODO:
+If the spec gets updated to support atomic fetching and updates of multiple Named Graphs at the same Resource,
+this section should be updated to clarify that; the slicing-and-dicing idea might make more sense with that context.
+
+-->
+
+For example, to read my name, you would first fetch a LitDataset from `https://vincentt.inrupt.net/profile/card`. Then, from that LitDataset, you could extract the Thing that represents my profile. From that, you can then read my name.
```
NSeydoux, line 19:
```suggestion
For example, to read my name, you would first fetch a LitDataset from `https://vincentt.inrupt.net/profile/card`, the URL of my profile document. Then, from that LitDataset, you could extract the Thing that represents my profile. From that, you can then read my name.
```

website/docs/guide/core-concepts.md
```diff
@@ -0,0 +1,33 @@
...
+
+For a more extensive overview of fetching and manipulating data, see the tutorial [Working with Data](../tutorials/working-with-data).
+
+## Immutability
+
+The functions exposed in lit-pod are designed to take data as input, apply some transformation, and return the transformed data as output without manipulating the input data. That makes it easier for you to write unit tests, and enables frameworks that want to be notified of changes to data (like most modern front-end frameworks) to apply performance optimizations like memoisation, checking for updates by references rather than doing deep comparisons. However, it is good to be aware of this fact; when you're seeing stale data, that is likely the result of a variable pointing to an input value, rather than to the returned result.
```
NSeydoux, line 25:
```suggestion
The functions exposed in lit-pod are designed to take data as input, apply some transformation, and return the transformed data as output **without changing the input data**. That makes it easier for you to write unit tests, and enables frameworks that want to be notified of changes to data (like most modern front-end frameworks) to apply performance optimizations like memoisation, checking for updates by references rather than doing deep comparisons. However, it is good to be aware of this fact; when you're seeing stale data, that is likely the result of a variable pointing to an input value, rather than to the returned result.
```

website/docs/guide/installation.md
```diff
@@ -0,0 +1,17 @@
...
+
+```bash
+npm install @solid/lit-pod
+```
+
+lit-pod will work in Node using CommonJS modules, and in the browser with a bundler like [Webpack](https://webpack.js.org), [Rollup](https://rollupjs.org) or [Parcel](https://parceljs.org).
+
+## To access private data
+
+By default, lit-pod will only enable you to access public data on Solid Pods. To enable access to restricted data, you will need to include a library that enables you to allow the user to authenticate. If you have [solid-auth-client](https://www.npmjs.com/package/solid-auth-client) installed and have used it to authenticate the user, lit-pod will automatically pick up the authenticated session and include the user's credentials with each request.
```
NSeydoux, line 17:
```suggestion
By default, lit-pod will only enables you to access public data on Solid Pods. To allow access to restricted data, you will need to include a library that supports user authentication. If you have [solid-auth-client](https://www.npmjs.com/package/solid-auth-client) installed and have used it to authenticate the user, lit-pod will automatically pick up the authenticated session and include the user's credentials with each request.
```

website/docs/guide/motivation.md
```diff
@@ -0,0 +1,37 @@
+---
+id: motivation
+title: Motivation
+sidebar_label: Motivation
+---
+
+[Solid](https://solidproject.org) is a wonderful project, but the technology it's built on —[RDF](https://en.wikipedia.org/wiki/Resource_Description_Framework)— is unfamiliar to many developers and has a relatively high barrier to entry.
```
NSeydoux, line 7: > has a relatively high barrier to entry.

No it has not, that's anti-RDF propaganda !

website/docs/guide/motivation.md
```diff
@@ -0,0 +1,37 @@
...
+
+The main goal of lit-pod is to make it as easy as possible for front-end engineers to get started writing web apps that store data on Solid Pods, without hampering the ability to leverage the more advanced capabilities of RDF if so desired.
```
NSeydoux, line 9:
```suggestion
The main goal of lit-pod is to make it as easy as possible for engineers to get started writing apps both for the browser and for Node that store data on Solid Pods by providing an abstraction layer on top of both Solid and RDF principles, without hampering the ability to leverage the more advanced capabilities of RDF if so desired.
```
I don't know if this qualifies as documenting for features we don't have, because as long as we don't have a library to log in from Node, the focus is much more on frontend, but technically nothing prevents from using lit-pod in both environments.

website/docs/guide/motivation.md
```diff
@@ -0,0 +1,37 @@
...
+
+Whereas lit-pod makes it easy to get started with Solid, it should not lock you in when you are ready to move on to more advanced use cases. That's why lit-pod is compatible with the [RDF/JS specification](https://rdf.js.org), allowing you to seamlessly switch between lit-pod and libraries that support more advanced use cases for RDF.
+
+## Non-goals
+
+To ensure we achieve those goals, it is important to have a clear focus. Therefore, the following areas are out of scope for the foreseeable future.
+
+### Reasoning
+
+While RDF is explicitly designed to support reasoning over data, it is not entirely clear yet how that would work in practice, especially in the context of Solid. Therefore, lit-pod requires you to explicitly indicate what data you are reading/writing.
```
NSeydoux, line 33:
```suggestion
While RDF is explicitly designed to support reasoning over data, it is not entirely clear yet how that would work in practice, especially in the context of Solid. Therefore, lit-pod requires you to explicitly indicate what data you are reading/writing, and only data explicitly stated in the target Pod will be considered.  
```

website/docs/guide/motivation.md
Vinnl: I'll allow it :P I mean, it already works in Node for public data, and if a Node-compatible fetcher is ever developed, you can use it with lit-pod today, so from our point of view, we definitely support it.

website/docs/guide/core-concepts.md
Vinnl: Hmm, I was kinda avoiding that - does clarifying that that's the URL of my _profile document_ clarify anything here for the reader? (i.e. my hunch is that the reader does not know what a profile document is, so it would only make it _less_ clear? Unless the intent is the other way around, i.e. to use this as an explanation of what a profile document is, but given that that's not the subject of this guide, and that it's not that important of a concept anyway, I'm not sure if that works as intended?)

website/docs/guide/installation.md
Vinnl: Haha, I sure do love the word "enable", it seems.

website/docs/guide/motivation.md
Vinnl: We have always been at war with RDF.

NSeydoux: This war has been going on for far too long, and caused far too much suffering

website/docs/guide/core-concepts.md
NSeydoux: Hmm, I see your point. My intention was to describe why we would be looking for your name at this particular IRI, but not necessarily to describe what a profile document is. Would it add less new concepts if we reused the phrasing from a previous paragraph, e.g. "the URL where my personal information is stored." ?

Vinnl: Yes, that makes sense. I'll make the update, thanks.

Vinnl: bcf080d6101e8f152820dd452bae057e92ab49ac

PR #211, opened 2020-06-30
Fix navigation links to the Getting Started page

Vinnl:
Because its set as the `homePageId`, its link is the base path of
the documentation. (Hence, we also only mark it as active when
*only* that base path is the current path.)


Vinnl:
Cool, so now you can verify that it actually works :)

Click "Get Started" here and get an error: https://inrupt.github.io/lit-pod/

Or here where it's successful: https://lit-pod-8ssxw0cvd.vercel.app/

PR #212, opened 2020-06-30
Allow reading public access to a Resource

Vinnl:
# New feature description

This PR adds `getPublicAccessModes()`, `getPublicResourceAccessModes()` and `getPublicDefaultAccessModes()`. They're mostly similar to the respective agent-specific functions, but applying to `acl:agentClass` `foaf:Agent`, of course.

# Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,5 +12,5 @@ First release! What's possible with this first release:
 - Store data back to Solid Pods.
 - Read data from datasets.
 - Manipulate data in datasets.
-- Inspect a user's permissions w.r.t. a given Resource. (Experimental.)
-- Retrieve, delete and/or write any file (including non-RDF) from/to a Pod.
+- Inspect a user's and public permissions w.r.t. a given Resource. (Experimental.)
+- Retrieve, delete and/or write any file (including non-RDF) from/to a Pod. (Experimental.)
--- a/src/acl/agent.ts
+++ b/src/acl/agent.ts
@@ -110,7 +110,7 @@ export function unstable_getAgentAccessModesAll(
  *
  * Keep in mind that this function will not tell you:
  * - what access the given Agent has through other ACL rules, e.g. public or group-specific permissions.
- * - what access the given Agent has to child Resources, in case the associated Resource is a Container.
+ * - what access the given Agent has to child Resources, in case the associated Resource is a Container (see [[unstable_getAgentDefaultAccessModesOne]] for that).
  *
  * Also, please note that this function is still experimental: its API can change in non-major releases.
  *
@@ -212,7 +212,7 @@ export function unstable_setAgentResourceAccessModes(
  *
  * Keep in mind that this function will not tell you:
  * - what access the given Agent has through other ACL rules, e.g. public or group-specific permissions.
- * - what access the given Agent has to Container Resource itself (see [[unstable_getAgentResourceAccessModesOne]] for that).
+ * - what access the given Agent has to the Container Resource itself (see [[unstable_getAgentResourceAccessModesOne]] for that).
  *
  * Also, please note that this function is still experimental: its API can change in non-major releases.
  *
--- a/src/acl/agentClass.test.ts
+++ b/src/acl/agentClass.test.ts
@@ -22,7 +22,10 @@
 import { describe, it, expect } from "@jest/globals";
 import { DataFactory } from "n3";
 import { dataset } from "@rdfjs/dataset";
-import { unstable_getPublicResourceAccessModes } from "./agentClass";
+import {
+  unstable_getPublicResourceAccessModes,
+  unstable_getPublicDefaultAccessModes,
+} from "./agentClass";
 import {
   LitDataset,
   WithResourceInfo,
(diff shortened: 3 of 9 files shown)


src/acl/agent.ts
```diff
@@ -110,7 +110,7 @@ export function unstable_getAgentAccessModesAll(
  *
  * Keep in mind that this function will not tell you:
  * - what access the given Agent has through other ACL rules, e.g. public or group-specific permissions.
- * - what access the given Agent has to child Resources, in case the associated Resource is a Container.
+ * - what access the given Agent has to child Resources, in case the associated Resource is a Container (see [[unstable_getAgentDefaultAccessModesOne]] for that).
```
pmcb55, line 113: Just noticing the plural in this name - i.e. I'd have named it `unstable_getAgentDefaultAccessModeOne`.

CHANGELOG.md
```diff
@@ -12,5 +12,5 @@ First release! What's possible with this first release:
 - Store data back to Solid Pods.
 - Read data from datasets.
 - Manipulate data in datasets.
-- Inspect a user's permissions w.r.t. a given Resource. (Experimental.)
-- Retrieve, delete and/or write any file (including non-RDF) from/to a Pod.
+- Inspect a user's and public permissions w.r.t. a given Resource. (Experimental.)
```
pmcb55, line 15: Remove the word 'and'...?

src/acl/agentClass.ts
```diff
@@ -0,0 +1,132 @@
...
+ * Find out what Access Modes have been granted to everyone for a given Resource.
+ *
+ * Keep in mind that this function will not tell you what access specific Agents have through other ACL rules, e.g. agent- or group-specific permissions.
+ *
+ * Also, please note that this function is still experimental: its API can change in non-major releases.
+ *
+ * @param resourceInfo Information about the Resource to which the given Agent may have been granted access.
+ * @returns Which Access Modes have been granted to everyone for the given LitDataset, or `null` if it could not be determined (e.g. because the current user does not have Control Access to a given Resource or its Container).
+ */
+export function unstable_getPublicAccessModes(
```
pmcb55, line 52: I guess it's a more general question of allowing plurals in method names (or class names)...? @ajacksified - not sure if you have an opinion on this general question...?

src/acl/agent.ts
Vinnl: That's kind of weird, since that sounds like you're just getting a single access mode (we don't have functions that return a single mode - you always get all modes together), because it's not clear what `One` refers to then. So to be extremely consistent, that should be `getAgentDefaultAccessModeAllOne`, which I don't think would help anyone, or `getAgentOneDefaultAccessModeAll()` or some other permutation, but that doesn't really feel like it's clarifying a lot either...

(And as you noted elsewhere, this also applies to `getAgentResourceAccessModes`, `getAgentAccessModes`, `getPublicAccessModes`, `getPublicResourceAccessModes` and `getPublicDefaultAccessModes`.)

CHANGELOG.md
Vinnl: That would be incorrect: you can inspect public permissions and user-specific permissions separately. There's no such thing as "public user permissions". Not sure if there's a clearer way to word that?

src/acl/agentClass.ts
Vinnl: I'll resolve this one - let's keep that discussion at https://github.com/inrupt/lit-pod/pull/212#discussion_r447661698 to keep it at a single place. (Or possible in the coding guidelines; given that it's already in the codebase, we shouldn't really block this PR over this.)

src/acl/agentClass.test.ts
```diff
@@ -0,0 +1,546 @@
...
+  agentClass:
+    | "http://xmlns.com/foaf/0.1/Agent"
+    | "http://www.w3.org/ns/auth/acl#AuthenticatedAgent"
+): unstable_AclDataset {
+  const subjectIri =
+    resource + "#" + encodeURIComponent(agentClass) + Math.random();
+  aclDataset.add(
+    DataFactory.quad(
+      DataFactory.namedNode(subjectIri),
+      DataFactory.namedNode("http://www.w3.org/1999/02/22-rdf-syntax-ns#type"),
```
NSeydoux, line 53: This could use the `rdf.type` constant defined in `constants.ts`, but it's really minor. This also applies to:
- http://www.w3.org/ns/auth/acl#Authorization
- http://www.w3.org/ns/auth/acl#accessTo
- http://www.w3.org/ns/auth/acl#mode
- foaf:Agent

Vinnl: Yep, but I've refrained from using conveniences from other parts of the codebase in the tests. It might be a bit overzealous in this case, but in case those constants were wrong, this means that the unit tests would catch that. We can change that, but then I suppose it's best to be consistent and do that over the entire code base (because all tests use exact URLs atm), i.e. in a separate PR :)

Vinnl:
@NSeydoux I wouldn't make it _that_ generic, since the spec explicitly enumerates which Agent Classes are supported, i.e. [Public](https://github.com/solid/web-access-control-spec/#public-access-all-agents) (which we're supporting now) and [Authenticated Agents](https://github.com/solid/web-access-control-spec/#authenticated-agents-anyone-logged-on). In other words, developers cannot assume that an arbitrary Agent Class will actually be checked by the server, so we shouldn't give them that idea through our API, I think.

That said, I did implement it this way to make it easy to update when the spec adds support for additional Agent Classes. Which automatically leads to the question: should we also add support for checking AuthenticatedAgent? It _is_ in the current spec, but [I've seen discussion about it](https://gitter.im/solid/specification?at=5eecbc62fa0c9221fc4a3fc5) and how it is [misleading](https://gitter.im/solid/specification?at=5eecfe99c4bdd83475f380b6), and is thus even more likely to change.

src/acl/agentClass.test.ts
NSeydoux: Oh yeah right my bad, you told me already that you intentionnally kept the tests self-sufficient, that actually makes a lot of sense.

NSeydoux:
I think we don't need to check for AuthenticatedAgent: as you say, it is under discussion, and as long as its status is unclear, we don't need to expose it in our API. And for the genericity, that's a good point, I did not realize that the standard explicitly listed the supported classes, but in this case we should not encourage non-compliance by enabling to use `agentClass` and `agentGroup` in similar fashions (e.g. providing an arbitrary IRI), when they are quite different.

Vinnl:
Well, Agent _Groups_ _can_ be an arbitrary URL, as long as it points to a `vcard:AddressBook`, IIRC :)

PR #213, opened 2020-06-30
Add isContainer()

### LINKED ISSUE #203: unstable_fetchResourceInfoWithAcl does not return type information, or webIds for ACL
The new API for fetching resource information without a full request (`unstable_fetchResourceInfoWithAcl`) does not return the information needed for our use-case. Currently it does not return the type information of the given resource and it appears to have a simplified ACL.

The desired information received from this call the following:

- The type of the resource
- All permissions granted to the resource and the webIds associated with the permissions (so we can display profile info for each permitted party)
- Fallback permissions

If I were to put this data into a JSON object, it would look something like this:

```json
{
  "type": "http://www.w3.org/ns/ldp#BasicContainer",
  "acl": {
    "https://dayton.dev.inrupt.net/profile/card#me": {
      "read": true,
      "write": true,
      "append": true,
      "control": true,
    },
    "https://ajacksified-dev.dev.inrupt.net/profile/card#me": {
      "read": true,
      "write": true,
      "append": true,
      "control": false,
    },
    "fallbackAcl": {
      "read": true,
      "write": false,
      "append": false,
      "control": false,
    }
  }

}
```

Vinnl:
# New feature description

This adds a function that allows developers to check whether a given Resource is a Container.

# Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ See https://inrupt.github.io/lit-pod/docs/
 
 # Browser support
 
-lit-pod uses relatively modern JavaScript features that will work in all commonly-used browsers, except Internet Explorer. If you need support for Internet Explorer, it is recommended to pass it through a tool like [Babel](https://babeljs.io), and to add polyfills for e.g. `Set`, `Promise`, `Headers` and `Array.prototype.includes`.
+lit-pod uses relatively modern JavaScript features that will work in all commonly-used browsers, except Internet Explorer. If you need support for Internet Explorer, it is recommended to pass it through a tool like [Babel](https://babeljs.io), and to add polyfills for e.g. `Set`, `Promise`, `Headers`, `Array.prototype.includes` and `String.prototype.endsWith`.
 
 # Changelog
 
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -27,6 +27,7 @@ import {
   createLitDataset,
   fetchLitDataset,
   unstable_fetchResourceInfoWithAcl,
+  isContainer,
   saveLitDatasetAt,
   saveLitDatasetInContainer,
   unstable_saveAclFor,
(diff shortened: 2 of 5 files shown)


ajacksified:
This itself doesn't fix #203 - we still need the contentType header exposed to close the issue.

Vinnl:
@ajacksified Ah, I thought you already had that. And you need it primarily to determine whether you're dealing with a LitDataset or not, right?

ajacksified:
Partly, but also so that we can attempt to do things such as showing images inline (which we can only do if we know a given file is an image.)

Vinnl:
@ajacksified Have you already verified that Content-Type is sufficient for that? I just uploaded a `.png` to a test Pod, but the only Content-Type I'm getting back is `application/octet-stream` (i.e. the Pod doesn't know).

ajacksified:
Yes, and I've confirmed - for example, a `.txt` file with an Accept header of `*/*` still returns `text/plain`.

Vinnl:
Starting to doubt that now. Let's hold off on merging this until we can confirm that with someone/the spec.

Vinnl:
Ah, [there we go](https://solid.github.io/specification/#uri-slash-semantics):

> The slash character in the URI path indicates hierarchical relationship segments, and enables relative referencing. The semantics of the slash character is shared by servers and clients. **Paths ending with a slash denote a container resource.**

(Emphasis mine.)

Vinnl:
That said, maybe to be extra safe, might be best to check for `ldp:Container` in the Link header anyway, that does _feel_ less likely to break....

NSeydoux:
Well I know full conformance to the LDP spec is under discussion, so since the `/` semantics is part of the Solid spec, I guess it's our best option... I would agree that it feels "hackier", but if it's stated explicitly in the spec, then it should be okay. Plus, it's an easy check, so I expect a lot of clients/apps to depend on it anyway rather than on the Link header. 

Long story short, I think this can be merged as is.

Vinnl:
Ah, right. Well, if you say so, let's go by that :) If it does change, we should be able to change to checking the header without changing the API anyway.

PR #214, opened 2020-06-30
Add isLitDataset() function

Vinnl:
# New feature description

This will allow applications that explore a Pod without knowing
what type of data they might encounter to determine what fetch
method to use when they have `WithResourceInfo`.

# Checklist

### DIFF
--- a/src/litDataset.ts
+++ b/src/litDataset.ts
@@ -182,10 +182,18 @@ export async function unstable_fetchResourceInfoWithAcl(
 export function parseResourceInfo(
   response: Response
 ): WithResourceInfo["resourceInfo"] {
+  const contentTypeParts =
+    response.headers.get("Content-Type")?.split(";") ?? [];
+  const isLitDataset =
+    contentTypeParts.length > 0 &&
+    ["text/turtle", "application/ld+json"].includes(contentTypeParts[0]);
+
   const resourceInfo: WithResourceInfo["resourceInfo"] = {
     fetchedFrom: response.url,
+    isLitDataset: isLitDataset,
     contentType: response.headers.get("Content-Type") ?? undefined,
   };
+
   const linkHeader = response.headers.get("Link");
   if (linkHeader) {
     const parsedLinks = LinkHeader.parse(linkHeader);
@@ -212,6 +220,14 @@ export function parseResourceInfo(
  */
 export function isContainer(resource: WithResourceInfo): boolean {
   return resource.resourceInfo.fetchedFrom.endsWith("/");
+}
+
+/**
+ * @param resource Resource for which to check whether it contains a LitDataset.
+ * @return Whether `resource` contains a LitDataset.
+ */
+export function isLitDataset(resource: WithResourceInfo): boolean {
+  return resource.resourceInfo.isLitDataset;
 }
 
 /**
@@ -324,7 +340,7 @@ export async function saveLitDatasetAt(
     litDataset
   )
     ? { ...litDataset.resourceInfo, fetchedFrom: url }
-    : { fetchedFrom: url };
+    : { fetchedFrom: url, isLitDataset: true };
   const storedDataset: LitDataset &
     WithChangeLog &
     WithResourceInfo = Object.assign(litDataset, {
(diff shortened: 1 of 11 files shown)


src/litDataset.ts
```diff
@@ -182,9 +182,17 @@ export async function unstable_fetchResourceInfoWithAcl(
 export function parseResourceInfo(
   response: Response
 ): WithResourceInfo["resourceInfo"] {
+  const contentTypeParts =
+    response.headers.get("Content-Type")?.split(";") ?? [];
+  const isLitDataset =
+    contentTypeParts.length > 0 &&
+    ["text/turtle", "application/json+ld"].includes(contentTypeParts[0]);
```
NSeydoux, line 189: Aren't `application/rdf+xml` and RDFa (I don't recall the mime type and I'm on my phone) spec-compliant serializations ?

acoburn: application/json+ld => application/ld+json

acoburn: @NSeydoux RDFa typically uses just `text/html`, so it's hard to identify it as an RDF serialization per se. Plus, there are other complications. `application/rdf+xml` is legit, but I'd recommend against it -- it doesn't support datasets and ... it's XML. Really, though, if you don't have to support RDF+XML, I'd recommend against it.

NSeydoux: Oh yeah, I would never encourage anyone to use RDF/XML, it's an awful serialization, but if we fetch a resource from a Pod and it happens to be RDF/XML, we should be able to recognize it as an RDF Resource I guess (maybe display a popup that goes "RDF/XML ? REALLY ?"), and if it is allowed by the spec I don't think there's a way around it. 

By the way, I just checked, N3 doesn't support RDF/XML. The current spec being ultra-precise (`Linked Data resources (RDF in the form of JSON-LD, Turtle, HTML+RDFa, etc)`, I suggest we leave RDF/XML alone in the cold and not support it. If detecting RDFa is also not possible based on `Content-Type`, this conversation is resolved.

acoburn:
```suggestion
    ["text/turtle", "application/ld+json"].includes(contentTypeParts[0]);
```

Vinnl: @NSeydoux The Solid spec mandates Turtle and JSON-LD, so if no `Accept` header is sent, the server can only manually serialise it to either of that.

If the developer wants to explicitly store RDFa or RDF/XML, they can do so using the file-related functions we expose - in fact, they should, because parsing that as a LitDataset might lose information.

@acoburn Wow, I'd recently looked up the MIME type of JSON-LD and apparently still got it wrong :|

NSeydoux: > The Solid spec mandates Turtle and JSON-LD explicitly, so if no Accept header is sent, the server can only manually serialise it to either of that, right?

I'd expect so yes.

> If the developer wants to explicitly store RDFa or RDF/XML, they can do so using the file-related functions we expose - in fact, they should, because parsing that as a LitDataset might lose information, and they can't rely on the server understanding that either .

The file-based approach would work indeed, and do you have the pointer to the part of the spec that explicitly says that only Turtle and JSON-LD are supported ? I think we can act like that's the case anyway (as mentioned above), but I didn't know things were that clear (which is great).

Vinnl: Hmm, I think Michiel said that to me once, but I can't really find it. What I _can_ find is [this](https://github.com/solid/solid-spec/blob/master/api-rest.md#reading-resources):

> **IMPORTANT:** a default `Content-Type: text/turtle` will be used for requests for RDF resources or views (containers) that do not have an `Accept` header.

So maybe I should just remove the JSON-LD MIME type and just check that it's `text/turtle`?

acoburn: @Vinnl I wouldn't rely on any statement from the old solid-spec repo. The default of `text/turtle` is true if the backend is a conforming LDP server, but that isn't a strict requirement of the new Solid spec, so I don't think you can rely on that. Personally, I would leave JSON-LD.

Vinnl: Thanks @acoburn. But then the follow-up question is: is _just_ JSON-LD enough, or is it also possible for Solid servers to suddenly default to a different representation?

acoburn: I really don't think you'll find a _default_ representation that's not Turtle or JSON-LD. At least not by something that claims to be Solid compliant.

Vinnl: Perfect! Thanks :)

Issue #202, opened 2020-06-26
Remove unstable_ prefix for unstable functions


Vinnl:
Concerns have been raised that `unstable_` looks unprofessional. It was intended to make it easier to migrate when the function signature changes later; instead, we'll preserve the deprecated functions with a `_deprecated` suffix when we introduce replacements, so a stopgap migration could always be a simple search-and-replace.

The experimental status of an API should still be highlighted in the API documentation as follows:

> The use of the $_ prefix indicates that this function is pre version 1.0 and there is a possibility the function signature will change.

Additionally, we'll mark the `internal_` functions using just `_`, i.e. the relatively common convention of Hungarian notation for private functions.

PR #227, opened 2020-07-03
Read Group permissions

NSeydoux:
# New feature description

This adds the ability to read Group permissions, be it for a given Resource or for the child Resources of a given Container.
Permissions may be read for a given Group, or for all Groups that have explicitly been granted access rights in the target Resource's ACL document.

# Checklist

### DIFF
--- a/src/acl/agent.ts
+++ b/src/acl/agent.ts
@@ -20,7 +20,6 @@
  */
 
 import {
-  WebId,
   WithResourceInfo,
   WithChangeLog,
   unstable_WithAcl,
@@ -28,6 +27,7 @@ import {
   unstable_Access,
   unstable_AclRule,
   IriString,
+  WebId,
 } from "../interfaces";
 import { getIriOne, getIriAll } from "../thing/get";
 import { acl, rdf } from "../constants";
@@ -310,15 +310,20 @@ export function unstable_setAgentDefaultAccess(
   return cleanedAcl;
 }
 
+/** @internal */
+export function getAclRulesForIri(
+  aclRules: unstable_AclRule[],
+  iri: IriString,
+  targetType: IriString
+): unstable_AclRule[] {
+  return aclRules.filter((rule) => getIriAll(rule, targetType).includes(iri));
+}
+
 function getAgentAclRulesForAgent(
   aclRules: unstable_AclRule[],
   agent: WebId
 ): unstable_AclRule[] {
-  return aclRules.filter((rule) => appliesToAgent(rule, agent));
-}
-
-function appliesToAgent(aclRule: unstable_AclRule, agent: WebId): boolean {
-  return getIriAll(aclRule, acl.agent).includes(agent);
+  return getAclRulesForIri(aclRules, agent, acl.agent);
 }
 
 function getAgentAclRules(aclRules: unstable_AclRule[]): unstable_AclRule[] {
@@ -439,20 +444,28 @@ function intialiseAclRule(access: unstable_Access): unstable_AclRule {
 }
 
 function getAccessByAgent(aclRules: unstable_AclRule[]): unstable_AgentAccess {
-  const agentAccess: unstable_AgentAccess = {};
+  return getAccessByIri(aclRules, acl.agent);
+}
+
+/** @internal */
+export function getAccessByIri(
+  aclRules: unstable_AclRule[],
+  targetType: IriString
+): unstable_AgentAccess {
+  const targetIriAccess: Record<IriString, unstable_Access> = {};
 
   aclRules.forEach((rule) => {
-    const ruleAgents = getIriAll(rule, acl.agent);
+    const ruleTargetIri = getIriAll(rule, targetType);
     const access = internal_getAccess(rule);
 
     // A rule might apply to multiple agents. If multiple rules apply to the same agent, the Access
     // Modes granted by those rules should be combined:
-    ruleAgents.forEach((agent) => {
-      agentAccess[agent] =
-        typeof agentAccess[agent] === "undefined"
+    ruleTargetIri.forEach((targetIri) => {
+      targetIriAccess[targetIri] =
+        typeof targetIriAccess[targetIri] === "undefined"
           ? access
-          : internal_combineAccessModes([agentAccess[agent], access]);
+          : internal_combineAccessModes([targetIriAccess[targetIri], access]);
     });
   });
-  return agentAccess;
+  return targetIriAccess;
 }
--- a/src/acl/agent.ts
+++ b/src/acl/agent.ts
@@ -43,6 +43,8 @@ import {
   unstable_getFallbackAcl,
   internal_accessModeIriStrings,
   internal_removeEmptyAclRules,
+  internal_getAclRulesForIri,
+  internal_getAccessByIri,
 } from "../acl";
 import { createThing, getThingAll, setThing } from "../thing";
 import { removeIri } from "../thing/remove";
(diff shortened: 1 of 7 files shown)


src/acl/agent.ts
```diff
@@ -439,20 +444,28 @@ function intialiseAclRule(access: unstable_Access): unstable_AclRule {
...
-  const agentAccess: unstable_AgentAccess = {};
+  return getAccessByIri(aclRules, acl.agent);
+}
+
+/** @internal */
+export function getAccessByIri(
+  aclRules: unstable_AclRule[],
+  targetType: IriString
+): unstable_AgentAccess {
+  const targetIriAccess: Record<IriString, unstable_Access> = {};
```
Vinnl, line 455: This makes sense, but maybe move it (and `getAclRulesForIri`) into `../acl.ts`, since it's not specific to Agents any more. Also, be sure to prefix it with `internal_` (we'll rename it in #202).

Vinnl: Also, maybe instead of `IriString`, make this `typeof acl.agent | typeof acl.agentGroup`, not so much as a restriction, but as documentation of how this function should be used? Or alternatively (or additionally), just adding a bit of TSDoc should help as well.

src/acl/group.test.ts
```diff
@@ -0,0 +1,1158 @@
...
+  unstable_getGroupResourceAccessOne,
+  unstable_getGroupResourceAccessAll,
+  unstable_getGroupDefaultAccessAll,
+  unstable_getGroupAccessOne,
+  unstable_getGroupAccessAll,
+} from "./group";
+
+function addAclRuleQuads(
+  aclDataset: LitDataset & WithResourceInfo,
+  group: WebId,
```
Vinnl, line 44: I'm not sure about this, but I don't think we call a Group's IRI a WebID? Or do we?

```suggestion
  group: IriString,
```

src/acl/group.test.ts
```diff
@@ -0,0 +1,1158 @@
...
+      { read: false, append: false, write: false, control: true },
+      "resource"
+    );
+    const litDatasetWithAcl = addAclDatasetToLitDataset(
+      litDataset,
+      resourceAcl,
+      "resource"
+    );
+
+    const access = unstable_getGroupAccessOne(
```
Vinnl, line 161: Ah this is a good idea, to name it `Group` rather than `AgentGroup`. Note to self: rename `agentClass.ts` to `class.ts`.

src/acl/group.test.ts
```diff
@@ -0,0 +1,1158 @@
...
+      read: false,
+      append: false,
+      write: false,
+      control: true,
+    });
+  });
+});
+
+describe("getAgentAccessAll", () => {
+  it("returns the Resource's own applicable ACL rules, grouped by Group WebID", () => {
```
Vinnl, line 332: Same question about whether Groups have a WebID.

src/acl/group.ts
```diff
@@ -0,0 +1,212 @@
...
+  getAclRulesForIri,
+  unstable_AgentAccess,
+  getAccessByIri,
+} from "./agent";
+import { acl } from "../constants";
+
+/**
+ * Find out what Access Modes have been granted to a given Group of agents specifically for a given Resource.
+ *
+ * Keep in mind that this function will not tell you what access the given Group has through other ACL rules, e.g. public permissions.
```
Vinnl, line 52:
```suggestion
 * Keep in mind that this function will not tell you what access members of the given Group have through other ACL rules, e.g. public permissions.
```

src/acl/group.ts
```diff
@@ -0,0 +1,212 @@
...
+      group
+    );
+  }
+  return null;
+}
+
+/**
+ * Find out what Access Modes have been granted to specific Groups of agents for a given Resource.
+ *
+ * Keep in mind that this function will not tell you what access each Group has through other ACL rules, e.g. public permissions.
```
Vinnl, line 82:
```suggestion
 * Keep in mind that this function will not tell you what access members of each Group have through other ACL rules, e.g. public permissions.
```

src/acl/group.ts
```diff
@@ -0,0 +1,212 @@
...
+  }
+  return null;
+}
+
+/**
+ * Given an ACL LitDataset, find out which access modes it provides to a Group for its associated Resource.
+ *
+ * Keep in mind that this function will not tell you:
+ * - what access the given Group has through other ACL rules, e.g. public permissions.
+ * - what access the given Group has to child Resources, in case the associated Resource is a Container (see [[unstable_getGroupDefaultAccessModesOne]] for that).
```
Vinnl, line 108:
```suggestion
 * - what access members of the given Group have through other ACL rules, e.g. public permissions.
 * - what access members of the given Group have to child Resources, in case the associated Resource is a Container (see [[unstable_getGroupDefaultAccessModesOne]] for that).
```

src/acl/group.ts
```diff
@@ -0,0 +1,212 @@
...
+  const groupAccessModes = groupResourceRules.map(internal_getAccess);
+  return internal_combineAccessModes(groupAccessModes);
+}
+
+/**
+ * Given an ACL LitDataset, find out which access modes it provides to specific Groups for the associated Resource.
+ *
+ * Keep in mind that this function will not tell you:
+ * - what access arbitrary Groups might have been given through other ACL rules, e.g. public permissions.
+ * - what access arbitrary Groups have to child Resources, in case the associated Resource is a Container.
```
Vinnl, line 135:
```suggestion
 * - what access arbitrary members of these Groups might have been given through other ACL rules, e.g. public permissions.
 * - what access arbitrary members of these Groups have to child Resources, in case the associated Resource is a Container.
```

src/acl/group.ts
```diff
@@ -0,0 +1,212 @@
...
+  );
+  return getAccessByGroup(resourceRules);
+}
+
+/**
+ * Given an ACL LitDataset, find out which access modes it provides to a given Group for the associated Container Resource's child Resources.
+ *
+ * Keep in mind that this function will not tell you:
+ * - what access the given Group has through other ACL rules, e.g. public permissions.
+ * - what access the given Group has to the Container Resource itself (see [[unstable_getGroupResourceAccessOne]] for that).
```
Vinnl, line 158:
```suggestion
 * - what access members of the given Group have through other ACL rules, e.g. public permissions.
 * - what access members of the given Group have to the Container Resource itself (see [[unstable_getGroupResourceAccessOne]] for that).
```

src/acl/group.ts
```diff
@@ -0,0 +1,212 @@
...
+  const groupAccessModes = groupDefaultRules.map(internal_getAccess);
+  return internal_combineAccessModes(groupAccessModes);
+}
+
+/**
+ * Given an ACL LitDataset, find out which access modes it provides to specific Groups for the associated Container Resource's child Resources.
+ *
+ * Keep in mind that this function will not tell you:
+ * - what access arbitrary Groups have through other ACL rules, e.g. public permissions.
+ * - what access arbitrary Groups have to the Container Resource itself (see [[unstable_getGroupResourceAccessAll]] for that).
```
Vinnl, line 185:
```suggestion
 * - what access arbitrary members of these Groups have through other ACL rules, e.g. public permissions.
 * - what access arbitrary members of these Groups have to the Container Resource itself (see [[unstable_getGroupResourceAccessAll]] for that).
```

src/acl/agent.ts
NSeydoux: Good point, I hesitated to do this in the first place and wanted to do it in a future PR, but if you see the need for it as well that's settled. Fixed in c439095268f5f2842fe17c17cecb2365d8fbf487.

src/acl/group.test.ts
NSeydoux: We don't, I tried to be careful about this but missed this one, good catch :)

NSeydoux: And also I realized that mid-dev, which is why there are some blatant errors such as `https://some.pod/group#webId` ^^

src/acl/group.test.ts
NSeydoux: Fixed in 2f3ede9b6e024a60563c7462b2db024d4f447345

src/acl/group.test.ts
NSeydoux: Refactor express, at your service: #228

Issue #114, opened 2020-06-03
Use @inrupt/lit-generated-vocab-solid-common


Vinnl:
In `constants.ts` we have a number of IRI's of specific namespaces. The package `@inrupt/lit-generated-vocab-solid-common` has now been published to npm and includes those same constants, so let's use those instead and verify how much that impacts bundle sizes.

ThisIsMissEm:
I think this maybe would me more complicated, as we'd want to move the Artifact Generator to be an officially supported piece of code — currently the SDK team doesn't regularly maintain it, and as @inrupt/lit-generated-vocab-solid-common is built with Artifact Generator, that'd be a blocker

ThisIsMissEm:
We're going to close this as "won't fix" for now, but we have this issue on our backlog to look at later. Artifact Generator is officially supported, but we have a lot of work in our backlog relating to it.

PR #229, opened 2020-07-06
replaced constants with generated vocabs, force IRIs instead of strings

pmcb55:
# New feature description
Use generated vocabs instead of constants, force the use of IRI's instead of strings.

# Checklist
<Just a draft PR for discussion>


Vinnl:
Have the issues described in #114 been resolved yet?

src/interfaces.ts
```diff
@@ -30,7 +30,7 @@ export type Iri = Url;
 /**
  * Alias to indicate where we expect to be given a URL.
  */
-export type UrlString = string;
+export type UrlString = Iri;
```
NSeydoux, line 33: I think one of the core discussion elements around this change was that it would require developers to use an additional wrapper around their string IRIs when using our methods, which is an overhead.

package.json
```diff
@@ -29,6 +29,7 @@
     "url": "https://github.com/inrupt/lit-pod.git"
   },
   "devDependencies": {
+    "@inrupt/vocab-common-rdfext": "^0.4.9",
```
NSeydoux, line 32: I'm not a big fan of using `rdfext` in the module name, since the `NamedNode` class is defined in the RDFJS interface that both rdflib and rdfext support. I'd suggest `"@inrupt/vocab-common-lite`

Vinnl:

PR #234, opened 2020-07-06
Move Resource-specific functionality to own module

### LINKED ISSUE #161: Isolate `Resource` functions from `LitDataset` functions
**Search terms you've used**

Resource, LitDataset

**Is your refactoring request related to a problem? Please describe.**

As more features are implemented that are relevant for both `LitDataset` and plain files, depending on the motivation to add these features they may be added arbitrarily to `src/litDataset.ts` or `src/nonRdfData.ts`. Centralizing these functions in one place would help having code and tests more focused.

`src/resource.ts` (and `src/resource.test.ts`) is created, containing functions such as `parseResourceInfo`, `internal_fetchResourceAcl`, etc, and the associated tests.

Keeping all the `Resource`-related code in the current `src/litdataset.ts` file

**Additional context**

This issue also aims at experimenting the `refactoring` issue template.

Vinnl:
There was a bunch of functionality that was not actually specific
to LitDatasets, but also applied to non-RDF Resources (e.g.
related to fetching its ACL). These are now moved into a separate
`resource` module. Note how they are [now grouped in their own module in the docs](https://lit-pod-git-refactor-161-resource-functions.inrupt.vercel.app/docs/api/modules/_resource_).

The next step would probably be to move this, `litDataset` and `nonRdfData` into a separate `resource/` folder (and do something similar for `acl.ts` and `thing.ts` and their respective folders).

This PR fixes #161.


src/resource.ts
```diff
@@ -0,0 +1,312 @@
...
+
+/**
+ * Retrieve the information about a resource (e.g. access permissions) without
+ * fetching the resource itself.
+ *
+ * @param url URL to fetch Resource metadata from.
+ * @param options Optional parameter `options.fetch`: An alternative `fetch` function to make the HTTP request, compatible with the browser-native [fetch API](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/fetch#Parameters).
+ * @returns Promise resolving to the metadata describing the given Resource, or rejecting if fetching it failed.
+ */
+export async function internal_fetchResourceInfo(
```
NSeydoux, line 49: I just realised that we don't export this function publicly. Is it something that we want to expose along `unstable_fetchResourceInfo` ?

Vinnl: I suppose we do, but I think we decided against it for now because it wasn't actually used and we weren't sure about the return type (being `{ resourceInfo: { /* the actual resourceInfo */ } }`).

NSeydoux: Oh right, I remember this discussion now. You're right, as long as nobody needs it, we should keep it internal (non-breaking change and all that)

NSeydoux:
There are conflicts due to the merging of #235, I'm solving these locally.

Vinnl:
Yeah I'd anticipated those; I'm happy to fix them if you want, since I'm more familiar with the changes in this PR. Let me know if you prefer that :)

NSeydoux:
I added 67995825f244c2ab187dcf47f977d26a10c9484d to move `getFetchedFrom` in the `resource.ts` file instead of `litDataset.ts`

Vinnl:
Thanks! I squashed it into the other commit and updated with `master`, will merge when CI succeeds.

PR #235, opened 2020-07-06
Aligns fetchFile with fetchDataset

### LINKED ISSUE #219: Align non-RDF functions return types with RDF ones
**Search terms you've used**

Non-RDF, Response

**Is your refactoring request related to a problem? Please describe.**

There is a mismatch between functions such as `unstable_fetchFile`, which returns a `Promise<Response>`, much like a regular fetch, and `fetchLitDataset`, which returns a `Promise<LitDataset & WithResourceInfo>`. There is a gap in abstraction between the two, and returning a plain `Response` puts more burden on the developer to extract relevant information from this response. 

Since #215 adds `Content-Type` to `WithResourceInfo`, which was initially missing and was a motivation to return a full Response for non-RDF files, the return type of should be `unstable_fetchFile` should be `Promise<Blob & WithResourceInfo>`.

Keep the current design, where the developers has to manually extract informations from the `Response`. 

**Additional context**

Libraries depending on `unstable_fetchFile` will need to be updated accordingly.

NSeydoux:
Fixes #219, #152. 

Instead of returning a Response that the developer has to parse, this returns the same types as fetchLitDataset. This does not change deleting and writing non-RDF data (will do in a future PR).

# Checklist

### DIFF
--- a/website/docs/tutorials/working-with-files.md
+++ b/website/docs/tutorials/working-with-files.md
@@ -25,18 +25,21 @@ Reading a file is just a matter of fetching the content available at a certain U
 the fetched file as a blob. It is then up to you to decode it appropriately.
 
 ```typescript
-import { unstable_fetchFile } from "lit-solid";
+import {
+  unstable_fetchFile,
+  isLitDataset,
+  getContentType,
+  getFetchedFrom,
+} from "lit-solid";
 
 const file = await unstable_fetchFile(
   "https://example.com/some/interesting/file"
 );
 // file is a Blob (see https://developer.mozilla.org/en-US/docs/Web/API/Blob)
 console.log(
-  `Fetched a ${file.resourceInfo.contentType} file from ${file.resourceInfo.fetchedFrom}.`
+  `Fetched a ${getContentType(file)} file from ${getFetchedFrom(file)}.`
 );
-console.log(
-  `The file is ${file.resourceInfo.isLitDataset ? "" : "not "}a dataset.`
-);
+console.log(`The file is ${isLitDataset(file) ? "" : "not "}a dataset.`);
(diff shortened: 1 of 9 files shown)


website/docs/tutorials/working-with-files.md
```diff
@@ -27,12 +27,16 @@ the fetched file as a blob. It is then up to you to decode it appropriately.
...
-const response = await unstable_fetchFile(
+const file = await unstable_fetchFile(
   "https://example.com/some/interesting/file"
 );
-if (response.ok) {
-  const myFile = await response.blob();
-}
+// file is a Blob (see https://developer.mozilla.org/en-US/docs/Web/API/Blob)
+console.log(
+  `Fetched a ${file.resourceInfo.contentType} file from ${file.resourceInfo.fetchedFrom}.`
```
Vinnl, line 35: If we want to be able to also include this data in the dataset (I don't see how, since it might not be a dataset, but anyway... DO IT IN RDF!), we should stimulate the use the functions we expose here:

```suggestion
  `Fetched a ${getContentType(file)} file from ${file.resourceInfo.fetchedFrom}.`
```

Likewise for `isLitDataset`, and I suppose we might want to add a function to expose a Resource's original URL as well?

Vinnl:
This also fixes #152.

website/docs/tutorials/working-with-files.md
NSeydoux: Good point. A `getFetchedFrom` is added in 0cfd4f4b1791538047dbf776f021825f07906fca, and the doc is updated in  40ddcc7addb30de2306bf2b88e311f3b5f56feba.

PR #239, opened 2020-07-08
Skeleton for LDN notifications

NSeydoux:
This adds the capability to send an LDN notification to a target resource. 

NOTE: I'm experimenting an alternative flow to ease review: each atomic feature (e.g. inbox discovery or notification building) are going to be developped in their individual branches (e.g. `feat/notification/discovery`) and gradually merged into this branch (`feat/notification/main`). Only when all the features are developped, this PR will be switched off the draft state for a final (quick) review, given that each individual feature has been reviewed previously. 

# Checklist

### DIFF
--- a/src/notifications/ldn.ts
+++ b/src/notifications/ldn.ts
@@ -19,18 +19,34 @@
  * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
-import { LitDataset, Iri, Thing, IriString } from "../interfaces";
+import {
+  LitDataset,
+  Iri,
+  Thing,
+  IriString,
+  WithResourceInfo,
+} from "../interfaces";
 import { dataset, DataFactory } from "../rdfjs";
 import { fetch } from "../fetcher";
 import {
   internal_fetchResourceInfo,
   hasInboxInfo,
   getInboxInfo,
+  internal_toString
 } from "../resource";
-import { fetchLitDataset } from "../litDataset";
+import { 
+  fetchLitDataset, 
+  saveLitDatasetInContainer 
+} from "../litDataset";
 import { getThingOne } from "../thing";
 import { getIriOne } from "../thing/get";
 import { ldp } from "../constants";
+import { triplesToTurtle } from "../formats/turtle";
+
+/** @internal */
+export const internal_defaultFetchOptions = {
+  fetch: fetch,
+};
 
 /**
  * Perform partial inbox discovery (https://www.w3.org/TR/ldn/#discovery) by only checking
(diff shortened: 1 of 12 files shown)


src/notifications/ldn.ts
```diff
@@ -0,0 +1,84 @@
...
+ * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+import { LitDataset, Iri, Thing, IriString } from "../interfaces";
+import { dataset, DataFactory } from "../rdfjs";
+import { fetch } from "../fetcher";
+
+/**
+ * Perform partial inbox discovery (https://www.w3.org/TR/ldn/#discovery) by only checking
+ * resource content. If the inbox is only advsertised in the resource metadata, it won't be
```
Vinnl, line 28: **Edit:** Ah, there's `fetchInbox` below. Is there a reason someone would want to use this function rather than that one?

---

We _could_ support discovery from Link headers as well? (Might be a good moment to decide attaching the whole parsed link header to `resourceInfo`, possibly tagged as `internal_`, as an escape hatch.)

Also:

```suggestion
 * resource content. If the inbox is only advertised in the resource metadata, it won't be
```

src/notifications/ldn.ts
```diff
@@ -0,0 +1,84 @@
...
+ * discovered.
+ *
+ * @param resource The IRI of the resource for which we are searching for the inbox
+ * @param dataset The dataset where the inbox may be found (typically fetched at the resource IRI)
+ */
+export function unstable_discoverInbox(
```
Vinnl, line 34: Would this do an HTTP request? (I'd assume so, given that you're passing an IRI? If not, I'd maybe expect a `LitDataset` as parameter.) If so:

```suggestion
export async function unstable_discoverInbox(
```

(Also, for the documentation, maybe use the `Url | UrlString` types.)

And given that this is an external API, `UrlString` as return type :)

src/notifications/ldn.ts
```diff
@@ -0,0 +1,84 @@
...
+export function unstable_buildNotification(
+  sender: Iri | IriString,
+  target: Iri | IriString,
+  type: Iri | IriString,
+  options?: {
+    body: LitDataset;
+  }
+): LitDataset {
+  // NOTE: Unimplemented
+  return dataset();
```
Vinnl, line 61: Protip: to prevent TypeScript from complaining, I usually do:

```suggestion
  throw new Error("Not implemented yet");
```

src/notifications/ldn.ts
```diff
@@ -0,0 +1,84 @@
...
+    fetch: typeof fetch;
+  }
+) {
+  // NOTE: Unimplemented
+  void 0;
+}
+
+export async function unstable_sendNotification(
+  notification: LitDataset,
+  receiver: Iri | IriString,
```
Vinnl, line 77: I'm not too familiar with LDN, but does this expect a WebID? (If so, maybe use that alias for it.)

src/notifications/ldn.ts
```diff
@@ -0,0 +1,84 @@
...
+ * resource metedata (i.e. Link headers) and resource content.
+ *
+ * @param resource The IRI of the resource for which we are searching for the inbox
+ * @param dataset The dataset where the inbox may be found (typically fetched at the resource IRI)
+ */
+export async function unstable_fetchInbox(resource: Iri | IriString) {
+  return DataFactory.namedNode("unimplemented");
+}
+
+export function unstable_buildNotification(
```
Vinnl, line 52: For my understanding: is this just a convenience function to set some common-but-not-obligatory properties on a notification? Or are these parameters actually obligatory?

src/notifications/ldn.ts
NSeydoux: I intended this function not to make an HTTP request, in order to make it possible to look for inboxes in a dataset that has already been fetched. Maybe the naming isn't good then, what about `_unstable_getInbox(...)` ? It's `unstable_fetchInbox` that does the actual fetching.

src/notifications/ldn.ts
NSeydoux: No, in LDN any LDP resource may have an inbox.

src/notifications/ldn.ts
NSeydoux: It's a convenience function: since no shape is mandatory in LDN (as far as I know), I don't think any attribute of a notification is mandatory. This intends at capturing the core of the notification (sender, receiver, activity type), and just takes a dataset where any additional information may be stored. I'd like additional development in this area to be driven by shapes, by just passing a shape, a dataset, and get back objects, but that's for further down the road :)

src/notifications/ldn.ts
NSeydoux: Good point, that's how I initially approached it but I changed to have tests passing on intermediary PRs leading to build this feature, I don't know if it's a good practice.

src/notifications/ldn.ts
NSeydoux: And the return type is fixed in 611967efedfd15dbd52f59048e93882440246695

src/notifications/ldn.ts
NSeydoux: Basically there exists two functions to cover two use case (and allow optimizations): either you want to send a notification to a target resource for which you know the IRI, then you `fetchInbox`, or you have a resource fetched already and want to look up its inbox (but not perform one, potentially two additional HTTP requests), and then you `discoverInbox`. At least that's the initial intent :)

PR #241, opened 2020-07-10
Adds inbox discovery

NSeydoux:
# New feature description

An inbox may be discovered according to the LDP spec (https://www.w3.org/TR/ldn/#discovery)

# Checklist


NSeydoux:
The unused namespaces are removed in df78822ced1c4d7daeb5fdeca1b25909756c8996. And indeed, these tests are aligned with the current consensus on the codebase, so that if we agree to change the way we do it we won't have to deal with inconsistencies.

PR #245, opened 2020-07-13
Feat/notification/discover and send

NSeydoux:
This adds a convenience method that enables, in one API call, to perform both inbox discovery and to send a notification. This complete the feature set, and once merged #239 may be merged as well.

# Checklist

### DIFF
--- a/src/notifications/ldn.test.ts
+++ b/src/notifications/ldn.test.ts
@@ -473,29 +473,30 @@ describe("unstable_sendNotification", () => {
 
   it("should send the notification to the inbox of the given resource if found in its content", async () => {
     const turtle = `
-      @prefix : <#>.
       @prefix ldp: <https://www.w3.org/ns/ldp#>.
 
-      :aResource ldp:inbox :anInbox.
+      </aContainer/aResource> ldp:inbox </anotherContainer/anInbox>.
     `;
     const mockFetch = jest.fn(window.fetch).mockReturnValue(
       Promise.resolve(
         mockResponse(turtle, {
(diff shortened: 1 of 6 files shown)


src/notifications/ldn.test.ts
```diff
@@ -378,28 +405,161 @@ describe("unstable_buildNotification", () => {
...
+          url: "https://some.pod/",
+          headers: {
+            Location: "https://some.pod/anInbox/notification",
+          },
+        })
+      )
+    );
+    await unstable_sendNotification(
+      dataset(),
+      DataFactory.namedNode("https://some.pod#aResource"),
```
pmcb55, line 493: Not a bug, or a big deal or anything, but it seems strange to have resource IRIs with fragment identifiers - generally I'd reserve those for non-information resources, i.e. 'conceptual things'. Maybe this illustrates another advantage to using a shared vocab for test terms - i.e. to demonstrate best practices generally. Anyway, no need to change (unless it looks a little weird to you too).

pmcb55: Actually another issue with this example IRI is that it doesn't clearly demonstrate containment - i.e. what's it's parent container...?! I assume it's the domain (i.e. `https://some.pod/`, which must end in a slash), so potential for confusion (or awkward questions!) maybe...

NSeydoux: I don't mind the long fragment, but you're right that showing containment might make things clearer. Fixed in b54c9f9adaee938686f543843a06fb4faf4a10db

PR #267, opened 2020-07-22
Add createAcl()

Vinnl:
Note: this PR builds on #263, so probably best to review this after that is merged.

# New feature description

This PR adds a function `createAcl()`, which initialises a new empty ACL for a Resource (and is used by `createAclFromFallbackAcl()`, which then fills it with the ACL Rules from the fallback ACL).

(I've also moved `saveAclFor()` and `deleteAclFor()` into the ACL module. No code has changed there.)

# Checklist


src/acl/acl.ts
```diff
@@ -221,6 +221,26 @@ export function unstable_getFallbackAcl(
   return dataset.acl.fallbackAcl;
 }
 
+/**
+ * Initialise an empty Resource ACL for a given Resource.
+ *
+ * @param targetResource A Resource that does not have its own ACL yet (see [[unstable_hasResourceAcl]]).
+ * @returns A Resource ACL for the given Resource, with no ACL Rules defined yet.
+ */
+export function unstable_createAcl(
```
NSeydoux, line 230: If you were to call `unstable_createAcl`, and save it right away, you would basically lock the resource away right, because nobody would have any right on it ? Should this at least take a WebID as an input to give control to someone ? Another approach would be to copy the content of the ACL that currently applies, but we established how clunky this is...

Vinnl: Yes, perhaps I should add a warning to _not_ save it directly to the docs. That said, I don't think forcing the granting of Control access is a good idea; if you want to set it to public access, then that means you'd have to manually remove the WebID access after calling this.

We do already have a function to copy the content of the ACL that currently applies (createAclFromFallbackAcl, IIRC). However, that only works if the current user actually has access to the fallback ACL.

NSeydoux: Oh right, I remembered we discussed about it, but I didn't recall the name. Then maybe adding a warning in the doc is enough, but I'm a bit nervous about this locking users out of their data... If you want to give public control, then having explicit control for your own WebID is redundant, but not breaking, so that might be a "safest" alternative (or at least less error-prone).

But I don't want this to become an involved discussion for such a niche use case, so documenting it might be enough.

Vinnl: There might not be an "own WebID" though. For example, what if a Container has a default access rule that gives public Control access over its children. Then you don't even need to login to create a new ACL.

Generally, I think we'll show people how to `createAclFromFallbackAcl()` and potentially use this function as a fallback for that. I'll make sure to emphasise in the docs that they'll have to give someone Control access afterwards, and that they should consider just not changing access if no fallback ACL is found.

And then it will still go wrong, because ACLs are such a mess :/

NSeydoux: Yeah, that seems reasonable, given the state of the spec there's not much else we can do I'm afraid ^^

PR #289, opened 2020-07-30
Export types for internal use

Vinnl:
Even though these types might not be useful for downstreams to
use, they should still be included in the type definitions -
otherwise, all of our functions that return those types will be
interpreted as returning `any`.

- [ ] I've added a unit test to test for potential regressions of this bug. N/A


NSeydoux:
Is this related to the issue @megoth encountered ?

Vinnl:
@NSeydoux It was actually in response to https://github.com/inrupt/solid-ui-react/pull/17#discussion_r462969453, but Arne indeed encountered this as well. (It's not the bug you're fixing in #290 though.)

PR #290, opened 2020-07-30
Fix access granting bug when ACL is both default and resource

NSeydoux:
If an ACL is both a default and a resource rule, an additional check must be run to verify that setting resource access for the agent doesn't escalate his default access.

### DIFF
--- a/src/acl/agent.test.ts
+++ b/src/acl/agent.test.ts
@@ -939,7 +939,7 @@ describe("setAgentResourceAccess", () => {
       DataFactory.triple(
         namedNode("https://arbitrary.pod/resource/?ext=acl#owner"),
         namedNode("http://www.w3.org/ns/auth/acl#agent"),
-        namedNode("https://ldp.demo-ess.inrupt.com/megoth/profile/card#me")
+        namedNode("https://arbitrary.pod/profileDoc#webId")
       )
     );
     sourceDataset.add(
@@ -1842,6 +1842,90 @@ describe("setAgentDefaultAccess", () => {
     expect(updatedQuads[5].object.value).toBe(
       "https://some.pod/profileDoc#webId"
     );
+  });
+
+  it("adds the appropriate Quads for the given Access Modes if the rule is both a resource and default rule", async () => {
+    const sourceDataset = Object.assign(dataset(), {
+      internal_accessTo: "https://arbitrary.pod/container/",
+      internal_resourceInfo: {
+        fetchedFrom: "https://arbitrary.pod/container/?ext=acl#owner",
+        isLitDataset: true,
+      },
+    });
+    sourceDataset.add(
+      DataFactory.triple(
+        namedNode("https://arbitrary.pod/container/?ext=acl#owner"),
+        namedNode("http://www.w3.org/1999/02/22-rdf-syntax-ns#type"),
+        namedNode("http://www.w3.org/ns/auth/acl#Authorization")
+      )
+    );
+    sourceDataset.add(
+      DataFactory.triple(
+        namedNode("https://arbitrary.pod/container/?ext=acl#owner"),
+        namedNode("http://www.w3.org/ns/auth/acl#resource"),
+        namedNode("https://arbitrary.pod/container/")
+      )
+    );
+    sourceDataset.add(
+      DataFactory.triple(
+        namedNode("https://arbitrary.pod/container/?ext=acl#owner"),
+        namedNode("http://www.w3.org/ns/auth/acl#agent"),
+        namedNode("https://arbitrary.pod/profileDoc#webId")
+      )
+    );
+    sourceDataset.add(
+      DataFactory.triple(
+        namedNode("https://arbitrary.pod/container/?ext=acl#owner"),
+        namedNode("http://www.w3.org/ns/auth/acl#default"),
+        namedNode("https://arbitrary.pod/container/")
+      )
+    );
+    sourceDataset.add(
+      DataFactory.triple(
+        namedNode("https://arbitrary.pod/container/?ext=acl#owner"),
+        namedNode("http://www.w3.org/ns/auth/acl#mode"),
+        namedNode("http://www.w3.org/ns/auth/acl#Read")
+      )
+    );
+    sourceDataset.add(
+      DataFactory.triple(
+        namedNode("https://arbitrary.pod/container/?ext=acl#owner"),
+        namedNode("http://www.w3.org/ns/auth/acl#mode"),
+        namedNode("http://www.w3.org/ns/auth/acl#Control")
+      )
+    );
+    sourceDataset.add(
+      DataFactory.triple(
+        namedNode("https://arbitrary.pod/container/?ext=acl#owner"),
+        namedNode("http://www.w3.org/ns/auth/acl#mode"),
+        namedNode("http://www.w3.org/ns/auth/acl#Write")
+      )
+    );
+
+    const updatedDataset = unstable_setAgentDefaultAccess(
+      sourceDataset,
+      "https://some.pod/profileDoc#webId",
+      {
+        read: true,
+        append: true,
+        write: false,
+        control: false,
+      }
+    );
+
+    // Explicitly check that the agent given resource access doesn't get additional privilege
+    getThingAll(updatedDataset).forEach((thing) => {
+      const agents = getIriAll(
+        thing,
+        "http://www.w3.org/ns/auth/acl#agent"
+      ).filter((agent) => agent === "https://some.pod/profileDoc#webId");
+      if (agents.length > 0) {
+        // The agent given resource access should not have resource access
+        expect(
+          getIriAll(thing, "http://www.w3.org/ns/auth/acl#accessTo")
+        ).toHaveLength(0);
+      }
+    });
   });
 
   it("does not alter the input LitDataset", () => {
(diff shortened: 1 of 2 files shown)


src/acl/agent.test.ts
```diff
@@ -911,6 +913,101 @@ describe("setAgentResourceAccess", () => {
...
 
+  it("adds the appropriate Quads for the given Access Modes if the rule is both a resource and default rule", async () => {
+    const sourceDataset = Object.assign(dataset(), {
+      internal_accessTo: "https://arbitrary.pod/resource",
+      internal_resourceInfo: {
+        fetchedFrom: "https://arbitrary.pod/resource/?ext=acl#owner",
+        isLitDataset: true,
+      },
+    });
+    sourceDataset.add(
```
pmcb55, line 924: Very clunky (and extremely verbose, and therefore hard-to-read) test data setup here. With hundreds of tests in this codebase already, we really should address this (later :) !)...

src/acl/agent.test.ts
```diff
@@ -911,6 +913,101 @@ describe("setAgentResourceAccess", () => {
...
+      DataFactory.triple(
+        namedNode("https://arbitrary.pod/resource/?ext=acl#owner"),
+        namedNode("http://www.w3.org/1999/02/22-rdf-syntax-ns#type"),
+        namedNode("http://www.w3.org/ns/auth/acl#Authorization")
+      )
+    );
+    sourceDataset.add(
+      DataFactory.triple(
+        namedNode("https://arbitrary.pod/resource/?ext=acl#owner"),
+        namedNode("http://www.w3.org/2000/01/rdf-schema#label"),
```
pmcb55, line 934: Do we really need an `rdfs:label` triple in this test data...?

NSeydoux: It's not necessary, but it was part of the data that prompted the bug, so I wanted the test to be as accurate as possible. It has no effect in this case, so it could be removed, but I was wondering if having real-world data in our tests wasn't an additional safety rail against regression.

pmcb55: I'd remove it. 
I don't understand what you mean about 'accurate as possible', or 'real-world data' here - i.e. if the original bug report also contained an 'rdfs:comment' triple and 15 'foaf:knows' triples, would you also include them in this test?! I understand you'd include them initially when trying to find out where the actual bug is, but once you understand and fix the bug, then the test data should be cleaned up to only contain whatever data is necessary to test the specific condition.

NSeydoux: Fixed in a9d6a522e9c5964b22392397c1e2492b2217fa28.

src/acl/agent.ts
```diff
@@ -386,6 +386,21 @@ function removeAgentFromResourceRule(
...
+    ).length > 0;
+  if (!agentHasDefaultAccess) {
+    // If the original rule if both a resource and a default rule, and did **not**
+    // give default access to the agent whose access are being set, only the resource
+    // access should be duplicated (not the default ones)
+    ruleForOtherTargets = removeIri(
+      ruleForOtherTargets,
+      acl.default,
+      resourceIri
+    );
```
Vinnl, line 401: Wouldn't this mean that, if the rule for other targets has an `acl:default` rule _and_ applies to other agents/agent classes/agent groups, that those other agents would no longer have default access any more either?

src/acl/agent.test.ts
Vinnl: There already are a couple of helper methods in this file for initialising mock data, perhaps we could re-use those?

src/acl/agent.ts
```diff
@@ -386,6 +386,21 @@ function removeAgentFromResourceRule(
     acl.accessTo,
     resourceIri
   );
+  const agentHasDefaultAccess =
+    getIriAll(rule, acl.agent).filter(
+      (originalAgent) => originalAgent === agent
+    ).length > 0;
```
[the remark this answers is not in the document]
NSeydoux: In the `updatedDataset`, some rules don't apply to the agent to which we are giving resource access, so this first step is just to filter out these and only check that the rules actually including the target agent don't include excessive access.

src/acl/agent.ts
NSeydoux: Ah darn you're right. Then I guess the simpler approach is to split the rules that are both resource and default into two separate rules, and then apply the original logic to the appropriate one, leaving the other unchanged.

NSeydoux: Wait no, because the existing rule (`ruleWithoutAgent`) still applies to agents whose access is unchanged. So either the new agent wasn't part of the original rule, in which case `const ruleWithoutAgent = removeIri(rule, acl.agent, agent);` doesn't impact the existing rule, or the agent whose access is changed was already in the rule, in which case `getIriAll(originalRule, acl.agent).includes(agent);` is true, and the existing access that isn't the subject of the change (e.g. default access for `setAgentResourceAccess`) is left untouched.

src/acl/agent.ts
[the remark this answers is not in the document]
Vinnl: OK, I think I understand what the issue is now:

1. We call `removeAgentFromResourceRule` to make sure that any existing rules for this agent-resource combination are removed, which splits up a rule into a rule for the agent (with different resources), and a rule for the resource (with different agents/groups/classes).
2. However, if the rule does not pertain to the given agent at all, creating a new rule for the agent (with different resources) means that all access given by that rule to other resources, will apply that access to the given agent.

Wouldn't the resolution here simply be to not split up the original rule if it does not apply to the given agent or to the given resource? Feels like it'd be easier to follow and safer that creating a new rule that possibly gives the agent too much access, and then remove that specific access again.

src/acl/agent.ts
Vinnl: Ah yes, you're right.

I do think I just thought of another bug though (pre-existing, not introduced by this PR): if a Rule applies to both an Agent and an Agent Group/Class, wouldn't _both_ `ruleWithoutAgent` and `ruleForOtherTargets` apply to those as well? Hence giving whatever access we're giving to the Agent to the Group or Class as well, in addition to the access they've already been given. Which would mean we'd have to call `removeAll(ruleForOtherTargets, acl.agentGroup)` and `removeAll(ruleForOtherTargets, acl.agentClass)` as well to fix it.

Let me know if you think that's correct, then I'll create another ticket/PR to fix it.

NSeydoux:
8126087fe12f4bb4d96e488c102c98b085264de9 makes the test data setup more compact by using helper functions, as suggested by @Vinnl

src/acl/agent.ts
NSeydoux: I added an `expect` to the test in 8126087fe12f4bb4d96e488c102c98b085264de9 to ensure this is true.

NSeydoux: And yeah that sounds correct. I'll merge the current PR, so that we can merge #295 as well and apply this fix on it directly, since it will likely be impacted

src/acl/agent.ts
```diff
@@ -386,6 +386,21 @@ function removeAgentFromResourceRule(
...
+  if (!agentHasDefaultAccess) {
+    // If the original rule if both a resource and a default rule, and did **not**
```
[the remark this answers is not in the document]
NSeydoux: Resolved in c0043138040ef6a43edb7e5a1d2b75269f94963f

src/acl/agent.ts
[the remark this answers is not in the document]
NSeydoux: You are right, it makes more sense to check if the agent is present in the rule before duplicating it rather than after to remove undue access. Fixed in 886ee47a9726eed22ec221d18e85a2ea4e8af64f.

src/acl/agent.test.ts
NSeydoux: Resolved in b7e66e9be4b42d61acf1fc7decd8ef7c2461f4b8

src/acl/agent.ts
```diff
@@ -386,6 +386,21 @@ function removeAgentFromResourceRule(
...
+    // give default access to the agent whose access are being set, only the resource
```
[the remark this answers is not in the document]
NSeydoux: Resolved in c0043138040ef6a43edb7e5a1d2b75269f94963f

src/acl/agent.ts
```diff
@@ -369,52 +372,36 @@ function duplicateAclRule(sourceRule: unstable_AclRule): unstable_AclRule {
...
+    );
+  } else {
+    // If the agent had no access, the new ACL rule is empty
+    ruleForOtherTargets = intialiseAclRule({
+      read: false,
+      append: false,
+      write: false,
+      control: false,
+    });
+  }
```
Vinnl, line 401: Ah, clever - I was thinking about returning `null` and then checking for that on the call-site, but given that empty rules are removed afterwards, this should fix that.

That said, I'd move this around a bit to make it an early exit condition:

```suggestion
  if(!getIriAll(rule, acl.agent).includes(agent)) {
    // If the agent had no access, We don't need to split this rule up:
    const emptyRule = intialiseAclRule({
      read: false,
      append: false,
      write: false,
      control: false,
    });
    return [rule, emptyRule];
  }
  // The existing rule will keep applying to Agents other than the given one: 
  const ruleWithoutAgent = removeIri(rule, acl.agent, agent);
  let ruleForOtherTargets: unstable_AclRule;
  // The agent already had some access in the rule, so duplicate it...
  ruleForOtherTargets = duplicateAclRule(rule);
  // ...but remove access to the original Resource:
  ruleForOtherTargets = removeIri(
    ruleForOtherTargets,
    ruleType === "resource" ? acl.accessTo : acl.default,
    resourceIri
  );
```

NSeydoux: Resolved in a9fd918d3f3062c014dd3729448ff3e52a8b691d

PR #295, opened 2020-08-02
added public ACL setters for resources and default

pmcb55:
# New feature description
Added new functions required by SDK-1073:
  unstable_setPublicDefaultAccess
  unstable_setPublicResourceAccess

Also exported two functions from src/acl/agent.ts (just for import into src/acl/class.ts):
  initialiseAclRule (and also fixed the typo in the original name)
  duplicateAclRule

# Checklist

### DIFF
--- a/src/acl/agent.ts
+++ b/src/acl/agent.ts
@@ -20,37 +20,36 @@
  */
 
 import {
-  WithResourceInfo,
-  WithChangeLog,
-  unstable_WithAcl,
-  unstable_AclDataset,
-  unstable_Access,
-  unstable_AclRule,
   IriString,
+  unstable_Access,
+  unstable_AclDataset,
+  unstable_AclRule,
+  unstable_WithAcl,
   WebId,
-  Iri,
+  WithChangeLog,
+  WithResourceInfo,
 } from "../interfaces";
 import { getIriOne, getIriAll } from "../thing/get";
-import { acl, rdf } from "../constants";
+import { acl } from "../constants";
 import {
-  internal_getAclRules,
-  internal_getResourceAclRulesForResource,
-  internal_getDefaultAclRulesForResource,
-  internal_getAccess,
+  duplicateAclRule,
+  initialiseAclRule,
   internal_combineAccessModes,
-  unstable_hasResourceAcl,
+  internal_getAccess,
+  internal_getAccessByIri,
+  internal_getAclRules,
+  internal_getAclRulesForIri,
+  internal_getDefaultAclRulesForResource,
+  internal_getResourceAclRulesForResource,
+  internal_removeEmptyAclRules,
+  unstable_getFallbackAcl,
   unstable_getResourceAcl,
   unstable_hasFallbackAcl,
-  unstable_getFallbackAcl,
-  internal_accessModeIriStrings,
-  internal_removeEmptyAclRules,
-  internal_getAclRulesForIri,
-  internal_getAccessByIri,
+  unstable_hasResourceAcl,
 } from "./acl";
-import { createThing, getThingAll, setThing } from "../thing/thing";
+import { getThingAll, setThing } from "../thing/thing";
 import { removeIri } from "../thing/remove";
 import { setIri } from "../thing/set";
-import { addIri } from "../thing/add";
 
 export type unstable_AgentAccess = Record<WebId, unstable_Access>;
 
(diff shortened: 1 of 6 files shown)


src/acl/agent.ts
```diff
@@ -334,7 +334,9 @@ function isAgentAclRule(aclRule: unstable_AclRule): boolean {
  *
  * @param sourceRule ACL rule to duplicate.
  */
-function duplicateAclRule(sourceRule: unstable_AclRule): unstable_AclRule {
+export function duplicateAclRule(
+  sourceRule: unstable_AclRule
+): unstable_AclRule {
```
Vinnl, line 339: It we're re-using these for groups and classes, we should probably move these to `./acl.ts`, since they're not agent-specific.

Also, given that exporting them makes them available to client libraries, we should prefix their names with `internal_`.

src/acl/agent.ts
```diff
@@ -418,7 +420,7 @@ function removeAgentFromDefaultRule(
   return [ruleWithoutAgent, ruleForOtherTargets];
 }
 
-function intialiseAclRule(access: unstable_Access): unstable_AclRule {
+export function initialiseAclRule(access: unstable_Access): unstable_AclRule {
```
Vinnl, line 423: It we're re-using these for groups and classes, we should probably move these to `./acl.ts`, since they're not agent-specific.

Also, given that exporting them makes them available to client libraries, we should prefix their names with `internal_`.

src/acl/class.test.ts
```diff
@@ -35,6 +37,8 @@ import {
   unstable_AclDataset,
   unstable_WithAcl,
 } from "../interfaces";
+import { Quad } from "rdf-js";
```
Vinnl, line 40: `DataFactory` is already imported, so to be consistent you could use `DataFactory.Quad` instead.

src/acl/class.test.ts
```diff
@@ -35,6 +37,8 @@ import {
...
+import { foaf } from "../constants";
```
Vinnl, line 41: Could you create a ticket about extracting constants from tests throughout the codebase? Now we'll have this test which uses one centrally-defined constant, and the rest defined inline, which I can imagine can confuse future contributors.

src/acl/class.ts
```diff
@@ -121,6 +125,154 @@ export function unstable_getPublicDefaultAccess(
...
 }
 
+/**
+ * Given an ACL LitDataset, modify the ACL Rules to set specific Access Modes for the public.
+ *
+ * If the given ACL LitDataset already includes ACL Rules that grant a certain set of Access Modes
+ * to the public, those will be overridden by the given Access Modes.
+ *
+ * Keep in mind that this function will not modify:
+ * - access arbitrary Agents might have been given through other ACL rules, e.g. public or group-specific permissions.
```
Vinnl, line 135:
```suggestion
 * - access arbitrary Agents might have been given through other ACL rules, e.g. agent- or group-specific permissions.
```

src/acl/agent.ts
pmcb55: Yep - agree they belong better in ./acl.ts.
I thought only functions exported from `index.ts` would be available to client libraries - but exporting anything from anywhere makes it available for client libraries...?

src/acl/agent.ts
pmcb55: Ditto to above.

src/acl/agent.ts
Vinnl: Yep. We can restrict that for users of Node 14 and upcoming versions of Webpack (see the ticket on ES Modules), but for every other downstream, everything we export will be available - which is why we're using the `internal_` prefix so often.

src/acl/class.test.ts
pmcb55: Created here: https://inrupt.atlassian.net/browse/SDK-1176 (although all this was already done in my previous PR!).

src/acl/agent.ts
pmcb55: Moved in `a3de75d`.

src/acl/class.test.ts
pmcb55: No, I think this is needed for the Type `Quad` (i.e. as in `const updatedQuads: Quad[]`), and not for the method `DataFactory.quad()` (i.e. the type `Quad` is not exported from `DataFactory.ts` (at least not the DataFactory exported from N3, which this test uses (which is another whole issue that my previous PR dealt with - i.e. using types from N3 instead of from RDF/JS))).

Vinnl: Ah whoops, you're right!

PR #302, opened 2020-08-03
Added a default Accept header when fetching a dataset

NSeydoux:
When fetching a dataset, we expect to get a turtle response, so this expectation must be materialized by an Accept header


NSeydoux:
Note that a better support for parsers is planned in the future. When starting out this PR, I initially worked on https://github.com/inrupt/solid-client-js/tree/feat/support-rdf-parser, but that is outside of the scope of this PR and advanced compared to what we need in the foreseeable future.

PR #306, opened 2020-08-04
Prevents copying over ACLs for rules with mixed audience

NSeydoux:
If an ACL has a mixed audience (e.g. Agent and Group), then editing access of one of the audiences no longer copies over access for the other. It prevents duplication, and potentially privilege escalation.

### DIFF
--- a/src/acl/class.test.ts
+++ b/src/acl/class.test.ts
@@ -53,15 +53,13 @@ function addAclRuleQuads(
     | "http://www.w3.org/ns/auth/acl#AuthenticatedAgent"
     | string,
   ruleIri?: IriString,
-  targetType?:
+  targetType:
     | "http://www.w3.org/ns/auth/acl#agentClass"
     | "http://www.w3.org/ns/auth/acl#agent"
-    | "http://www.w3.org/ns/auth/acl#agentGroup"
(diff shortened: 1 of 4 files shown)


src/acl/class.test.ts
```diff
@@ -49,11 +50,18 @@ function addAclRuleQuads(
...
     | "http://xmlns.com/foaf/0.1/Agent"
-    | "http://www.w3.org/ns/auth/acl#AuthenticatedAgent",
-  ruleIri?: IriString
+    | "http://www.w3.org/ns/auth/acl#AuthenticatedAgent"
+    | string,
+  ruleIri?: IriString,
+  targetType?:
+    | "http://www.w3.org/ns/auth/acl#agentClass"
+    | "http://www.w3.org/ns/auth/acl#agent"
+    | "http://www.w3.org/ns/auth/acl#agentGroup"
```
Vinnl, line 59: No biggie (especially since it's just a helper function for a test), but you can specify a default value in JS as follows:

```suggestion
  targetType:
    | "http://www.w3.org/ns/auth/acl#agentClass"
    | "http://www.w3.org/ns/auth/acl#agent"
    | "http://www.w3.org/ns/auth/acl#agentGroup"
    = "http://www.w3.org/ns/auth/acl#agentClass"
```

(or whatever formatting Prettier would choose)

PR #313, opened 2020-08-05
Feat/add get by locale

pmcb55:
# New feature description
Add the function 'getStringByLocale()' - makes the assumption you want xsd:string literal values back too (keyed on the empty string).

# Checklist

### DIFF
--- a/src/thing/get.test.ts
+++ b/src/thing/get.test.ts
@@ -1200,6 +1200,27 @@ describe("getStringByLocaleAll", () => {
     ]);
   });
 
+  it("allows empty language tags - literal is still an rdf:langString", () => {
+    const thing = dataset()
+      .add(makeQuad(SUBJECT, PREDICATE, "chien", "fr"))
+      .add(makeQuad(SUBJECT, PREDICATE, "dog", ""));
+
+    const thingWithLocaleStrings = Object.assign(thing, {
+      internal_url: SUBJECT,
+    });
+
+    // An empty language tag is valid, and even in Turtle should be serialized
+    // as an rdf:langString, e.g.:
+    //  <SUBJECT> <PREDICATE> "chien"@fr .
+    //  <SUBJECT> <PREDICATE> "dog"^^rdf:langString .
+    expect(
+      Array.from(getStringByLocaleAll(thingWithLocaleStrings, PREDICATE))
+    ).toEqual([
+      ["fr", ["chien"]],
+      ["", ["dog"]],
+    ]);
+  });
+
   it("ignores non-string literals...", () => {
     const thing = dataset()
       .add(makeQuad(SUBJECT, PREDICATE, "value 1", "fr"))
--- a/src/thing/get.test.ts
+++ b/src/thing/get.test.ts
@@ -1145,10 +1145,10 @@ describe("getStringWithLocale", () => {
 
 describe("getStringByLocale", () => {
   function makeQuad(
-    subject: string,
-    predicate: string,
-    value: string,
-    locale: string | NamedNode | undefined
+    subject: IriString,
+    predicate: IriString,
+    value: IriString,
+    locale?: IriString | Iri
   ): Quad {
     return DataFactory.quad(
       DataFactory.namedNode(subject),
--- a/src/thing/get.test.ts
+++ b/src/thing/get.test.ts
@@ -23,7 +23,7 @@ import { describe, it, expect } from "@jest/globals";
 import { dataset } from "@rdfjs/dataset";
 import { NamedNode, Quad, Literal } from "rdf-js";
 import { DataFactory } from "n3";
-import { IriString, Thing } from "../interfaces";
+import { IriString, Iri, Thing } from "../interfaces";
 import {
   getUrl,
   getBoolean,
@@ -48,6 +48,22 @@ import {
   getTermAll,
 } from "./get";
 import { xmlSchemaTypes } from "../datatypes";
+
+const SUBJECT = "https://arbitrary.vocab/subject";
+const PREDICATE = "https://some.vocab/predicate";
+
+function makeQuad(
+  subject: IriString,
+  predicate: IriString,
+  value: IriString,
+  locale?: IriString | Iri
+): Quad {
+  return DataFactory.quad(
+    DataFactory.namedNode(subject),
+    DataFactory.namedNode(predicate),
+    DataFactory.literal(value, locale)
+  );
+}
 
 function getMockQuadWithLiteralFor(
   predicate: IriString,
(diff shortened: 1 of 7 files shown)


src/thing/get.test.ts
```diff
@@ -1139,7 +1141,97 @@ describe("getStringWithLocale", () => {
...
+    expect(
+      Array.from(getStringByLocaleAll(thingWithLocaleStrings, predicate))
+    ).toEqual([
+      ["en", ["value 1", "value 2"]],
+      ["fr", ["value 3"]],
+      ["es", ["value 4"]],
+    ]);
+  });
+
+  it("includes non-language string values keyed with the empty language tag", () => {
```
NSeydoux, line 1182: I don't have a problem with this, but for consistency we should make sure that `getStringWithLocale` returns `xsd:string` when the parameter is an empty string.

src/thing/get.test.ts
```diff
@@ -1139,7 +1141,97 @@ describe("getStringWithLocale", () => {
   });
 });
 
-describe("getStringsWithLocaleAll", () => {
+describe("getStringByLocale", () => {
+  function makeQuad(
+    subject: string,
+    predicate: string,
+    value: string,
+    locale: string | NamedNode | undefined
```
Vinnl, line 1149: We have aliases for these :) 

```suggestion
    subject: IriString,
    predicate: IriString,
    value: IriString,
    locale?: IriString | Iri
```

Vinnl: Also, it looks like `locale` is actually `localeOrLiteralType`?

src/thing/get.ts
```diff
@@ -281,6 +281,43 @@ export function getStringWithLocaleAll(
   return matchingQuads.map((quad) => quad.object.value);
 }
 
+/**
+ * Retrieves all string literals for the specified property from the specified
+ * Thing.
```
Vinnl, line 286: This links to the Thing interface.

```suggestion
 * [[Thing]].
```

src/thing/get.ts
```diff
@@ -281,6 +281,43 @@ export function getStringWithLocaleAll(
...
+ *
+ * NOTE: Assumes you also want non-locale string literals too (i.e. literals
```
Vinnl, line 288: Just for consistency with other comments in the codebase.

```suggestion
 * Note: Assumes you also want non-locale string literals too (i.e. literals
```

src/thing/get.ts
```diff
@@ -281,6 +281,43 @@ export function getStringWithLocaleAll(
...
+ * with a key of the empty string.
+ *
+ * @param thing The [[Thing]] to read the localised string values from.
+ * @param property The given Property for which you want the localised string values.
+ * @returns A Map of objects, keyed on locale with the value an array of string values (for that locale).
+ */
+export function getStringByLocaleAll(
+  thing: Thing,
+  property: Url | UrlString
+): Map<string, string[]> {
```
Vinnl, line 299: Could you add to the README that people might need a polyfill for `Map` as well?

src/thing/get.test.ts
pmcb55: Actually yeah, I've changed this now to remove this special handling for empty locale values - it just keeps everything simpler and more explicit.
In fact, perhaps this would be a reason to expose a `getStringAll()` function that actually returned all string values, keys on locale, and in this case also include the `xsd:string` values keyed on either an empty locale, or maybe some proprietary constant value (like "NoLocaleProvided", or whatever).  I don't like the idea of the key here extending beyond the simple RFC notion of a language tag though, so would need to discuss - but lets see if anyone ever asks for such a function first :) !

src/thing/get.ts
pmcb55: `Map` feels like the correct structure here (as opposed to `Array`), but no super-strong opinion either way. It's just when I ask for something 'by' some key, I kinda expect to be able to then lookup by a key value easily - i.e. using `Map.get(key)` instead of having to iterate through an array.

src/thing/get.test.ts
```diff
@@ -1139,7 +1157,100 @@ describe("getStringWithLocale", () => {
...
+    expect(
+      Array.from(getStringByLocaleAll(thingWithLocaleStrings, PREDICATE))
+    ).toEqual([
+      ["en", ["value 1", "value 2"]],
+      ["fr", ["value 3"]],
+      ["es", ["value 4"]],
+    ]);
+  });
+
+  it("excludes non-language string values", () => {
```
NSeydoux, line 1194: What happens for empty-string locales ? Are they ignored or added with an empty string key ?

Vinnl:

pmcb55:
@Vinnl I've no problem closing this if the feature is no longer required by anyone - but wasn't this feature an explicit ask (from Ameya I think)? So is that feature request currently satisfied by the current code, or is the feature itself no longer being asked for...?!

Vinnl:
@pmcb55 No, it's just that there was no activity on this PR (e.g. answering the open questions), so I thought I'd close it until you've got time to push it over the finish line, at which time you can reopen it :)

pmcb55:
Ah Ok - so it seems this was just an oversight on my part then I think. I'll re-open this and try and get it merged...

src/thing/get.test.ts
pmcb55: Great question. At first I thought we should handle this case specially, but in fact it seems an empty language tag is still perfectly valid, i.e. the literal is still of type `rdf:langString`. This made me curious about how that could be serialized in Turtle (since `<> rdfs:label "Some value"@ .` would definitely look like a syntax error!), but in fact the correct serialization is simply: `<> rdfs:label "Some value"^^rdf:langString .`. I demonstrated this in a CodeSandbox, but deploying to Vercel gave me an error (https://csb-izxlf-21qfb3kdk.vercel.app/).
Anyway I just extended the tests to check for this: commit 8c58c414

pmcb55:
> Come to think of it: I actually discussed the signature [on a Tripledoc issue](https://gitlab.com/vincenttunru/tripledoc/-/issues/50#note_363732708). The signature we ended up with there was `Array<{ locale: string; value: string }`, i.e. making
> 
> ```
> [ { locale: "nl-nl", value: "dit is een goed idee" }, { locale: "fr-fr", value: "C'est une bonne idée" } ]
> ```
> 
> Then we could just set `locale` to `null`, feels more idiomatic to me (but that's mostly intuitive), and it's a bit more ergonomic to access.
> 
> Just a suggestion.

Yeah sure, I'll leave it as it is for now, but we can review later perhaps based on actual developer feedback

Vinnl:
Changing the return type later would be a breaking change later though, so there's quite a bit of inertia against changing this later. So best to be pretty convinced about the signature that we end up releasing first.

pmcb55:
I think the current signature makes sense right now - developers will always have different opinions about what's 'better', but the current signature isn't 'wrong', so I'm good with this.

PR #320, opened 2020-08-05
forgot to export the renamed getFileWithAcl function - thanks to Trav…

pmcb55:
…is Vachon


Vinnl:
@pmcb55 A bit of a nit, but it's good practice to keep the heading line of a commit message under 50 chars to make it easier to review the logs. I've resubmitted this in #326, using Travis' original commit as credit :)

Issue #251, opened 2020-07-16
Should we support deprecated predicate acl:defaultForNew?


megoth:
**Search terms you've used**

acl, defaultForNew

**Is your feature request related to a problem? Please describe.**

I'm working on an app that uses [acl/agent#unstable_getAgentAccessAll](https://inrupt.github.io/lit-pod/docs/api/modules/_acl_agent_#unstable_getagentaccessall), and noticed that my app failed to load the default access controls for a resource. I realized the Pod I'm working on used a deprecated version of `acl:default`, namely `acl:defaultForNew`.

**Describe the solution you'd like**

This is not a bug per ce, but wondered if there is plans to support reading `acl:defaultForNew`? Could help some confussion by developers who've used the Solid systems for a while, and might still have uses of `acl:defaultForNew` hanging around.

**Describe alternatives you've considered**

If we don't want to support `acl:defaultForNew`, we might want to note that in the documentation.

**Additional context**

Don't think any extra context is required.

NSeydoux:
Just for my understanding, is there any additional logic attached to supporting `acl:defaultForNew` compared to `acl:default` ? If not (i.e. both predicates are strictly equivalent), supporting the deprecated predicate seems rather harmless, so I don't see why we shouldn't. I can't think of a way to issue a warning to the developer that the predicate used is deprecated though, do you have anything in mind ? It's not really a big deal since the app developer isn't likely to have control over the Pod server, but it might encourage the ecosystem to use the latest version of the spec as much as possible.

megoth:
As I understand it `acl:default` and `acl:defaultForNew` are completely equivalent, it's just that the latter has been deprecated in favor of the former for the sake of brevity. I also cannot find any references to `acl:defaultForNew` in [the WAC spec](https://github.com/solid/web-access-control-spec), so I'm fairly certain of this.

Warning is perhaps to much, but a small note in the API documentation on all methods that involve default handling might be appropriate. I reckon this might be a distracting though, and not very helpful, so I understand that this isn't a good solution.

I just know that this is going to be very annoying in those 0.1% of cases where an app aren't able to handle the deprecated version 😛

acoburn:
[...]

PR #321, opened 2020-08-06
Support legacy defaultForNew predicate in ACLs

travis:
# Support legacy defaultForNew predicate in ACLs

Implements #251 

# Checklist

### DIFF
--- a/src/acl/acl.test.ts
+++ b/src/acl/acl.test.ts
@@ -476,7 +476,7 @@ describe("createAcl", () => {
 });
 
 describe("createAclFromFallbackAcl", () => {
-  it("creates a new ACL including existing default rules as Resource rules", () => {
+  function expectCreatesNewAclWithDefault(defaultPredicate: string) {
     const aclDataset: AclDataset = Object.assign(dataset(), {
       internal_accessTo: "https://arbitrary.pod/container/",
       internal_resourceInfo: {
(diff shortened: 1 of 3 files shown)


travis:
I don't think this is quite ready, but wanted to push it up for feedback. I suspect this needs more unit tests, but I'm not totally sure about the way I implemented the existing one - any suggestions for alternate approaches?

src/acl/acl.test.ts
```diff
@@ -476,7 +476,7 @@ describe("createAcl", () => {
 });
 
 describe("createAclFromFallbackAcl", () => {
-  it("creates a new ACL including existing default rules as Resource rules", () => {
+  function expectCreatesNewAclWithDefault(defaultPredicate: string) {
```
Vinnl, line 479: I'd prefer to keep tests self-contained, i.e. copy-paste the existing test but using `defaultForNew` instead of `default`, rather than extracting it into a function. (I think the acronym people like here is to "keep your tests DAMP".)

travis: yeaaa - that's what I was thinking too, thanks!

Vinnl:
If I understood it correctly @NSeydoux confirmed with you that he'd finish this up, so I'm closing this in favour of https://github.com/inrupt/solid-client-js/pull/377 (which includes this work). Thanks!

PR #332, opened 2020-08-06
Use myst-parser and Update Edit Link

kay-kim:
# 2 commits (Use myst-parser instead of recommonmark, install Sphinx after myst-parser to use latest Sphinx) and update edit link

## Commit 1: Update to use myst-parser

- [ myst-parser](https://github.com/executablebooks/MyST-Parser) allows for the rendering of markdown tables 
- With the use of myST, had to reorder packages in requirements as  Sphinx is on 3.1.x and myST is currently pinned to a 2.4.4 version. although, upgrade to 3.1 is [in the roadmap](https://github.com/executablebooks/MyST-Parser/issues/183) 
During pip install of requirements, will get `myst-parser 0.9.1 requires sphinx<3,>=2, but you'll have sphinx 3.1.2 which is incompatible.` but so far okay with the actual build html step.

## Commit 2: Update Edit Link

-  Needed to update Edit on Github link - was appending .txt when we changed to use .rst

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -134,13 +134,13 @@ CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
 OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-### recommonmark
+### myst-parser
 
-The documentation uses [recommonmark](https://github.com/readthedocs/recommonmark/), which is under the MIT License:
+The documentation uses [myst-parser](https://github.com/executablebooks/myst-parser), which is under the MIT License:
 
-The MIT License (MIT)
+MIT License
 
-Copyright (c) 2014 Steve Genoud
+Copyright (c) 2020 ExecutableBookProject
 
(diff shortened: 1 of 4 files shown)


Vinnl:
https://lit-pod-8xfml98yj.vercel.app/api.html still seems to work great and tables render properly.

One other question: it appears the headings do have an anchor attached (i.e. I can link to `getGroupResourceAccess` specifically: https://lit-pod-8xfml98yj.vercel.app/api/modules/_acl_group_.html#getgroupresourceaccess) - do you know what the syntax would be to link to that from other pages? I thought I saw something with a colon in between, but that didn't seem to work as far as I could see.

kay-kim:
👍 
Regarding the question about linking, to link from the rst docs to markdown docs, I think we'd have to use the :apidocs: link to point to https://docs.inrupt.com/client-libraries/solid-client-js/api/modules%s  for now.

It's because in restructured text, you can put reference links above headings a

`.. _linktargetA:

getPublicAccess
`

and use :ref:`linktargetA` to link.  The markdown files, at least what's being generated with Typedoc, don't put anything before these section headings.  We may be able to process the markdown files a bit using myST syntax extensions - but will need to investigate.

Thanks for the tip on commit messages.  I think it's because I'm used to having a specific docs ticket (at my previous, we have a Jira/github integration where if you prefix with the DOCS ticket DOCS-12345, and a commit is done on github, it linked that commit to the JIRA ticket).

PR #338, opened 2020-08-10
Align rule-splitters for Agents and the Public

Vinnl:
This is somewhat complex code that is mostly equivalent to both,
and will also be mostly equivalent for when we write support for
setting Group access and other potential future targets. Thus, to
prevent confusion, make sure that the functions actually _look_
mostly the same in addition to _behaving_ mostly the same.

# Checklist

### DIFF
--- a/src/acl/agent.ts
+++ b/src/acl/agent.ts
@@ -352,15 +352,15 @@ function removeAgentFromRule(
   }
   // The existing rule will keep applying to Agents other than the given one:
   const ruleWithoutAgent = removeIri(rule, acl.agent, agent);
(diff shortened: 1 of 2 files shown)


NSeydoux:
Do you think it would be worth considering factoring these two (soon three) functions into one, or would it add an additional complexity greater that the gain in reducing redundancy ?

Vinnl:
Yeah I was thinking about that too - in fact, not just these functions, but the entire files are mostly the same. But I'm pretty torn; it feels like it would be very hard to follow, and the function names would get even more unwieldy, probably. So the answer is: maybe, I don't know :shrug:

PR #340, opened 2020-08-10
Use current solid-client-authn-browser API in docs

Vinnl:
The code example in the docs wasn't aligned with how solid-client-authn is actually used now.

@kay-kim Unfortunately we can no longer just "pick up" solid-client-authn, which made it a bit harder for me to make it easy to follow. So if you have suggestions for this, they'd be very welcome :)

### DIFF
--- a/docs/source/examples/login.js
+++ b/docs/source/examples/login.js
@@ -25,24 +25,31 @@ import {
   Session,
   getClientAuthnWithDependencies
 } from '@inrupt/solid-client-authn-browser'
+import { getSolidDataset } from "@inrupt/solid-client":
 
 // Build a session
-const session = new solidClientAuthn.Session({
-    clientAuthn: solidClientAuthn.getClientAuthnWithDependencies({})},
+const session = new Session({
+    clientAuthn: getClientAuthnWithDependencies({})},
     "mySession"
 );
 
-// Redirect the user to their identity provider...
-await session.login({
-    // The URL of the user's OIDC issuer
-    oidcIssuer: 'https://identityProvider.com', 
-    // The url the system should redirect to after login
-    redirectUrl: 'https://mysite.com/redirect',
-});
+if (!session.sessionInfo.isLoggedIn) {
+  // Redirect the user to their identity provider:
+  // (This moves the user away from the current page.)
+  session.login({
+      // The URL of the user's OIDC issuer
+      oidcIssuer: 'https://inrupt.net',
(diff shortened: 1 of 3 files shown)


docs/source/examples/login.js
```diff
@@ -27,22 +27,28 @@ import {
...
-onLogin((sessionInfo) => {
-   // Do stuff
+// At https://mysite.com/redirect:
+if (!session.sessionInfo.isLoggedIn) {
+  await session.handleIncomingRedirect(new URL(window.location.href));
+}
 
-});
+// You can now make authenticated requests by passing session.fetch:
+fetchSolidDataset(session.sessionInfo.webId, { fetch: session.fetch });
```
kay-kim, line 52: Question: Should that be getSolidDataset?

docs/source/tutorial/read-write-data.rst
```diff
@@ -52,15 +52,18 @@ Resource.
 
 To authenticate, you can use the Inrupt `solid-client-authn
 <https://www.npmjs.com/package/@inrupt/solid-client-authn>`_ library.
-After authenticating with ``solid-client-authn``, |product| picks up
-the authenticated session and includes the user's credentials with each
-request.
+After authenticating, the ``Session`` object obtained from
+``solid-client-authn`` will give you a `fetch` function that you can pass to
+|product| to includes the user's credentials with a request.
```
kay-kim, line 57: will give you a -> provides a ``fetch`` function. 
to includes -> to include
Maybe:

> After authentication, the ``Session`` objection obtained from ``solid-client-authn`` provides a ``fetch`` function. You can pass this function as an option to |product| to include the user's credentials with a request.

docs/source/examples/login.js
kay-kim: Oh - or is this a separate function - not part of any package.

Or, if want to show usage with solid-client, do we need to update the include in this example file?

Vinnl: Whoopsie! We've been renaming so often I'm losing track of it myself... And yes, the import should be updated as well. I'll fix that.

Vinnl: 03da5c0864c528578c74ca8893591cc02de30cfc

docs/source/tutorial/read-write-data.rst
Vinnl: Much better. 92f357b40447334014b70cdffe0e88a937a89400

PR #344, opened 2020-08-11
Add mock functions for internal data structures

Vinnl:
# New feature description

This adds functions that mocks our objects and our runtime metadata. This makes it easy for downstreams to write unit tests for code that expects our runtime metadata, without breaking when we change how we store that metadata (i.e. when we move it into a Dataset with our proprietary named graph).

/cc @norbert989

# Checklist

- [ ] Relevant documentation, if any, has been written/updated. - ~~I will create a separate PR with a tutorial about writing tests, so Kay can review that. API documentation is updated though.~~ **Edit:** Although, this might be easier to do once I've seen how actual applications use this, rather than just component libraries.)

### DIFF
--- a/src/acl/mock.ts
+++ b/src/acl/mock.ts
@@ -1,0 +1,123 @@
+/**
+ * Copyright 2020 Inrupt Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to use,
+ * copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the
+ * Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,
+ * INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A
+ * PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+ * HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+ * OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+ * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+import {
+  WithResourceInfo,
+  WithResourceAcl,
+  WithAcl,
+  WithAccessibleAcl,
+  WithFallbackAcl,
+  UrlString,
+} from "../interfaces";
+import { getSourceIri } from "../resource/resource";
+import { createAcl, internal_getContainerPath } from "./acl";
+import { mockContainerFrom } from "../resource/mock";
+
+/**
+ * Function for use in unit tests to mock a [[SolidDataset]] that has its own ACL attached.
+ *
+ * Warning: do not use this function in actual production code.
+ * This function initialises a new empty ACL and attaches it to a given [[SolidDataset].
+ * This is useful to mock a SolidDataset with an ACL in tests of code that call e.g.
+ * [[getResourceAcl]].
+ *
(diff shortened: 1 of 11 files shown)


src/acl/mock.ts
```diff
@@ -0,0 +1,123 @@
...
+ *
+ * @param resource The Resource that we should pretend has its own ACL.
+ * @returns The input Resource, with an empty Resource ACL attached.
+ * @since Not released yet.
+ */
+export function addMockResourceAclTo<T extends WithResourceInfo>(
+  resource: T
+): T & WithResourceAcl {
+  const aclUrl =
+    resource.internal_resourceInfo.aclUrl ?? getSourceIri(resource) + ".acl";
```
NSeydoux, line 50:
```suggestion
  // The ACL URL being the resource URL with a .acl prefix is just for readability, 
  // but it is not a convention that should be relied on. ACL discovery is described in
  // the Solid specification
  const aclUrl =
    resource.internal_resourceInfo.aclUrl ?? getSourceIri(resource) + ".acl";
```

I'm afraid that not having a warning might lead some people looking at the code to jump to conclusions. I'm unsure whether the warning should include a link to the new spec, as it is currently a draft, to the old spec, wich will soon be outdated, or remain as it is there.

src/acl/mock.ts
```diff
@@ -0,0 +1,123 @@
...
+  const containerUrl = internal_getContainerPath(getSourceIri(resource));
+  const aclUrl = containerUrl + ".acl";
+  const mockContainer = setMockAclUrl(mockContainerFrom(containerUrl), aclUrl);
+  const aclDataset = createAcl(mockContainer);
+
+  const resourceWithFallbackAcl: typeof resource &
+    WithFallbackAcl = Object.assign(resource, {
+    internal_acl: {
+      resourceAcl:
+        ((resource as unknown) as WithAcl).internal_acl?.resourceAcl ?? null,
```
NSeydoux, line 100: For my own education, why is it necessary to cast to unknown before casting to `WithAcl` ?

src/resource/mock.test.ts
```diff
@@ -0,0 +1,98 @@
...
+});
+
+describe("mockFileFrom", () => {
+  it("is linked to the given source URL", async () => {
+    const mockedFile = mockFileFrom("https://some.pod/file.png");
+
+    expect(getSourceIri(mockedFile)).toBe("https://some.pod/file.png");
+  });
+
+  it("is recognised as a SolidDataset", async () => {
```
NSeydoux, line 79:
```suggestion
  it("is not recognised as a SolidDataset", async () => {
```

src/resource/mock.ts
```diff
@@ -0,0 +1,127 @@
...
+  const solidDatasetWithResourceInfo: SolidDataset &
+    WithResourceInfo = Object.assign(solidDataset, {
+    internal_resourceInfo: {
+      sourceIri: sourceIri,
+      isRawData: false,
+      contentType: "text/turtle",
+    },
+  });
+
+  return solidDatasetWithResourceInfo;
```
NSeydoux, line 96: This could return `mockSolidDatasetFrom(url)`

src/acl/mock.ts
Vinnl: Yeah good point. I think it might be best to just use an obviously arbitrary URL, rather than basing it off the source URL.

src/acl/mock.ts
Vinnl: Because the type of `resource` is not assignable to `WithAcl`, and vice versa. The background knowledge here is that these are not casts, but type assertions, i.e. they do not actually change the type of the object, but tell TypeScript that you know better than it what the type is. Which is often fine if the types are somewhat compatible, but since it's more likely to not be what you meant if they're completely complementary, TypeScript does warn you when you try to do that. And since `unknown` is compatible with all types, the double assertion does work.

It's explained more clearly here: https://basarat.gitbook.io/typescript/type-system/type-assertion#double-assertion

src/acl/mock.ts
Vinnl: bcbfd70767b29c2ad74ace146b4d83048ddfab6d

src/resource/mock.test.ts
Vinnl: Whoopsie 41a0a451d8fc51b21e722fd5bcbd48703d9130ed

src/resource/mock.ts
Vinnl: Now why did I not do this the first time...? Maybe because we might want to check a header or something for this down the road, but that's something we can resolve at that time.

f65305ca4c70ebd473cb937bd492b4f7770f7d50

src/acl/mock.ts
NSeydoux: Thanks !

PR #345, opened 2020-08-11
Add documentation on setting public access

Vinnl:
# New feature description

This PR fills in this section: https://docs.inrupt.com/client-libraries/solid-client-js/tutorial/manage-access-control-list.html#set-public-access

image

It can be previewed here: https://lit-pod-9ibmlh30i.vercel.app/tutorial/manage-access-control-list.html#set-public-access

# Checklist

### DIFF
--- a/docs/source/tutorial/manage-access-control-list.rst
+++ b/docs/source/tutorial/manage-access-control-list.rst
@@ -157,9 +157,28 @@ The general process of changing access to a Resource is as follows:
 Set Public Access
 -----------------
 
-.. note::
+Given a Resource's ACL obtained as specified in
+:ref:`manage-acl-change-access-to-resource`, you can:
 
-   |product| currently does not support setting public access.
+- Use :doc:`setPublicResourceAccess
+  </api/modules/_acl_class_>` to grant Access Modes
+  to everybody (whether logged in or not) for the Resource itself:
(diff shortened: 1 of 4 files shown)


docs/source/tutorial/manage-access-control-list.rst
```diff
@@ -157,9 +157,28 @@ The general process of changing access to a Resource is as follows:
 Set Public Access
 -----------------
 
-.. note::
+Given a Resource's ACL obtained as specified in
+:ref:`manage-acl-change-access-to-resource`, you can:
+
+- Use :doc:`setPublicResourceAccess
+  </api/modules/_acl_class_>` to grant Access Modes
+  to everybody for the Resource itself:
```
kay-kim, line 165: Q: Do you think we should explicitly mention that everybody refers to whether people are logged in or not like "everybody (whether logged in or not)" Technically, everybody should cover it, but then ... it's hard to tell how people actually read everybody.

Vinnl: Yes, that's a good idea - the standard currently also supports targeting just people logged in. (But they're leaning towards removing it, which is why we don't support it - but if we ever add it, this clarification will be helpful, so I'll add it.)

PR #346, opened 2020-08-11
update docs examples to use common-vocab library for foaf terms, remove mention of mysite.com since real site

kay-kim:
# Commit description

- Change mysite.com/redirect to example.com/redirect since mysite.com is a real site :D 
- Change references to foaf terms to use our common-vocabulary lib 

# Checklist

### DIFF
--- a/docs/source/examples/login.js
+++ b/docs/source/examples/login.js
@@ -37,9 +37,9 @@ if (!session.sessionInfo.isLoggedIn) {
   // Redirect the user to their identity provider:
   // (This moves the user away from the current page.)
   session.login({
-      // The URL of the user's OIDC issuer. Specify your OIDC issuer URL.
+      // Specify the URL of the user's OIDC issuer.
       oidcIssuer: 'https://inrupt.net',
-      // The URL the system should redirect to after login. Specify your redirect URL.
+      // Specify the URL the system should redirect to after login.
       redirectUrl: 'https://example.com/redirect',
   });
 }
--- a/docs/source/examples/login.js
+++ b/docs/source/examples/login.js
@@ -37,14 +37,14 @@ if (!session.sessionInfo.isLoggedIn) {
   // Redirect the user to their identity provider:
   // (This moves the user away from the current page.)
   session.login({
-      // The URL of the user's OIDC issuer
+      // The URL of the user's OIDC issuer. Specify your OIDC issuer URL.
       oidcIssuer: 'https://inrupt.net',
-      // The url the system should redirect to after login
-      redirectUrl: 'https://mysite.com/redirect',
+      // The URL the system should redirect to after login. Specify your redirect URL.
+      redirectUrl: 'https://example.com/redirect',
   });
 }
 
-// At https://mysite.com/redirect:
+// At your redirect site (e.g., https://example.com/redirect):
 if (!session.sessionInfo.isLoggedIn) {
   await session.handleIncomingRedirect(new URL(window.location.href));
 }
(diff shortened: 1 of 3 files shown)


docs/source/examples/login.js
```diff
@@ -37,19 +37,19 @@ if (!session.sessionInfo.isLoggedIn) {
   // Redirect the user to their identity provider:
   // (This moves the user away from the current page.)
   session.login({
-      // The URL of the user's OIDC issuer
+      // The URL of the user's OIDC issuer. Specify your OIDC issuer URL.
```
Vinnl, line 40: This seems confusing - as I understand it, the OIDC issuer isn't determined by the developer, but by the user. (e.g. if their Pod is hosted on inrupt.net, they enter inrupt.net, if it's hosted on solid.community, they enter solid.community.) @NSeydoux is that correct?

(It also seems somewhat redundant with the sentence before it?)

docs/source/examples/read-data-get-data.js
```diff
@@ -27,32 +27,34 @@ import {
   getUrlAll,
 } from "@inrupt/solid-client";
 
+import { FOAF } from "@inrupt/vocab-common-rdf";
```
Vinnl, line 30: On the one hand I think it's good that we're showing people what @inrupt/vocab-common-rdf can do for them, but by relying solely on that we leave no breadcrumbs at all that can guide the user to what to do if the vocab they need is not part of that package. Something to ponder...

docs/source/examples/read-data-get-data.js
```diff
@@ -27,32 +27,34 @@ import {
...
-// - Specifically, get the acquaintances (http://xmlns.com/foaf/0.1/knows).
+// - Specifically, get the acquaintances (FOAF.knows).
 // - Acquaintances is of type URL.
 // - Use getStringNoLocale expecting a single value.
 // - Use getStringNoLocaleAll if expecting multiple values.
 
-const acquaintances = getUrlAll(thing, "http://xmlns.com/foaf/0.1/knows");
-// => an array of URLs, presumably pointing to the Things describing acquaintances.
+const acquaintances = getUrlAll(thing, FOAF.knows);
+// => an array of URLs, pointing to the Things representing ("http://xmlns.com/foaf/0.1/knows")
```
Vinnl, line 58: Shouldn't this then also be

```suggestion
// => an array of URLs, pointing to the Things representing FOAF.knows
```

?

kay-kim: Hey - so ... you know how you mentioned above what about if people don't use a vocabulary in our library?  Heh, this was my clumsy way to show that FOAF.knows corresponds to "http://xmlns.com/foaf/0.1/knows"

docs/source/examples/login.js
kay-kim: Ah - yup.  I'll pop in "Specify " in front of the first sentence.  I just wanted to distinguish between comments where we're explaining what the source is doing and comments where we're pointing out places that users need to update.

docs/source/examples/read-data-get-data.js
kay-kim: Yup - so, I was trying that by leaving the comments below with the "representing the http://xmlns.com/foaf/0.1/<x>" but that seems to have missed the mark.  Will definitely ponder over a better way.

docs/source/examples/read-data-get-data.js
Vinnl: Haha, well that went right over my head. How about:

```suggestion
// => an array of URLs, pointing to the Things representing FOAF.knows ("http://xmlns.com/foaf/0.1/knows")
```

kay-kim: No worries.  I like your suggestion.  Will do so for the other examples in this file.  Thanks!

PR #356, opened 2020-08-14
Add CodeSandbox to PRs

### LINKED ISSUE #8: Set up CodeSandbox CI
After our first public release to npm (need not be stable), we can set up CodeSandbox to automatically incorporate a branch into a running app, so that we can quickly verify whether everything still works as expected.

More info:

- https://codesandbox.io/docs/ci
- https://codesandbox.io/post/codesandbox-ci-embeds

Vinnl:
# New feature description

This sets up CodeSandbox to build the code that's in the PR, and to add it as the dependency in the included sandbox. This allows us to verify that a PR will still work in a browser-based environment.

~~Unfortunately the CodeSandbox currently fails due to https://github.com/rdfjs/N3.js/issues/217, but I'm just submitting this PR to make sure it at least triggers it. After that bug gets fixed, we can see whether the Sandbox itself also works properly, and if so, we can detect issues like that sooner in the future.~~ Well what do you know, this one actually does work. I'm guessing because it builds it itself, and thereby takes into account the lockfile? That means this is ready to go :)

Fixes #8.

# Checklist

- [ ] All acceptance criteria are met. N/A


.codesandbox/sandbox/src/index.ts
```diff
@@ -0,0 +1,23 @@
+import {
+  getSolidDataset,
+  getThing,
+  getStringNoLocale,
+} from "@inrupt/solid-client";
+import { foaf } from "rdf-namespaces";
+
+getSolidDataset("https://vincentt.inrupt.net/profile/card").then(
```
NSeydoux, line 8: Is it okay to use your own profile in this, or would you rather point to a dummy ?

Vinnl: My profile is public, so I don't see the harm, but I guess it's better practice to use a dummy. I'll set something up.

Vinnl: 2be5ab93332682428d1dc7bd6f6bb438ae19d39d

PR #364, opened 2020-08-18
createEmptyContainerAt() and createEmptyContainerInContainer()

Vinnl:
# New feature description

This adds the ability to initialise a new Container in the Pod, and returns a SolidDataset that can be modified to add additional properties to the Container.

It adds a specific workaround for NSS, which does not support creating Containers using PUT, as mandated by the spec: https://github.com/solid/node-solid-server/issues/1465

It also adds a small workaround for ESS, which currently expects a Link header.

Do note that ESS does currently have a bug that causes it to return a 412 Precondition Failed after creating the Container (even though it does properly create the Container) when using `createEmptyContainerAt`.

# Checklist

- [ ] All acceptance criteria are met. _Apparently I chose slightly different function names. The others were thought of relatively ad-hoc, so I don't think that's a big issue?_

### DIFF
--- a/src/resource/solidDataset.test.ts
+++ b/src/resource/solidDataset.test.ts
@@ -40,6 +40,7 @@ import {
   getSolidDatasetWithAcl,
   createSolidDataset,
   createEmptyContainerAt,
+  createEmptyContainerInContainer,
 } from "./solidDataset";
 import {
   WithChangeLog,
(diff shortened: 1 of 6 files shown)


src/resource/solidDataset.test.ts
```diff
@@ -915,6 +917,432 @@ describe("saveSolidDatasetAt", () => {
...
+
+    expect(solidDataset.internal_resourceInfo.aclUrl).toBeUndefined();
+  });
+
+  it("provides the relevant access permissions to the Resource, if available", async () => {
+    const mockFetch = jest.fn(window.fetch).mockReturnValue(
+      Promise.resolve(
+        new Response(undefined, {
+          headers: {
+            "wac-aLLOW": 'public="read",user="read write append control"',
```
NSeydoux, line 1067: Why the casing on `aLLOW` ? Is that to test case-insensitivity ?

src/resource/solidDataset.test.ts
```diff
@@ -915,6 +917,432 @@ describe("saveSolidDatasetAt", () => {
...
+        fetch: mockFetch,
+      }
+    );
+
+    await expect(fetchPromise).rejects.toThrow(
+      new Error("Creating the empty Container failed: 403 Forbidden.")
+    );
+  });
+
+  it("returns a meaningful error when the server returns a 404", async () => {
```
NSeydoux, line 1182: Can a `PUT` return 404 ?

src/resource/solidDataset.test.ts
Vinnl: image

But yeah, that was the intention, but it was part of the original `getSolidDataset` tests that I copied this from - it's more confusing here, so I'll normalise that.

src/resource/solidDataset.test.ts
Vinnl: Ehm no, so this test doesn't really make sense. I'll remove it (it really is redundant with the 403 one anyway).

PR #366, opened 2020-08-18
Add ESS e2e test

NSeydoux:
This PR adds e2e tests against a demo ESS instance. The user profile associated with the Pod is dedicated to this only usage. The data has been created using https://github.com/NSeydoux/podpecker (deployed at https://podpecker.zwifi.eu/).

# Checklist

### DIFF
--- a/src/e2e.test.ts
+++ b/src/e2e.test.ts
@@ -52,176 +52,175 @@ import {
   createContainerInContainer,
 } from "./index";
 
-describe("End-to-end tests", () => {
-  it("should be able to read and update data in a Pod", async () => {
-    const randomNick = "Random nick " + Math.random();
+describe.each([["https://lit-e2e-test.inrupt.net/public/"]])(
+  "End-to-end tests against %s",
+  (rootContainer) => {
+    it("should be able to read and update data in a Pod", async () => {
+      const randomNick = "Random nick " + Math.random();
 
-    const dataset = await getSolidDataset(
-      "https://lit-e2e-test.inrupt.net/public/lit-pod-test.ttl"
-    );
-    const existingThing = getThing(
-      dataset,
-      "https://lit-e2e-test.inrupt.net/public/lit-pod-test.ttl#thing1"
-    );
-
-    expect(getStringNoLocale(existingThing, foaf.name)).toBe(
-      "Thing for first end-to-end test"
-    );
-
-    let updatedThing = setDatetime(
-      existingThing,
-      schema.dateModified,
-      new Date()
-    );
-    updatedThing = setStringNoLocale(updatedThing, foaf.nick, randomNick);
-
-    const updatedDataset = setThing(dataset, updatedThing);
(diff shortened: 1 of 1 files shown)


src/e2e.test.ts
```diff
@@ -52,16 +52,17 @@ import {
   createContainerInContainer,
 } from "./index";
 
-describe("End-to-end tests", () => {
+describe.each([
```
Vinnl, line 55: Hey cool, I didn't know about this function. That's great :)

src/e2e.test.ts
```diff
@@ -93,32 +94,34 @@ describe("End-to-end tests", () => {
...
 
   it("can create and remove empty Containers", async () => {
-    const newContainer1 = await createContainerAt(
-      "https://lit-e2e-test.inrupt.net/public/container-test/some-container/"
-    );
-    const newContainer2 = await createContainerInContainer(
-      "https://lit-e2e-test.inrupt.net/public/container-test/",
-      { slugSuggestion: "some-other-container" }
-    );
+    if (!rootContainer.includes("demo-ess")) {
```
Vinnl, line 107: It's somewhat nitpicky, but I think it'd be good to add something to the output showing that this test was skipped.

One approach would be to include the entire `it()` in the `if` clause, and then in the `else` add:

```js
   it.todo("can create and remove empty Containers")
```

(Or alternatively, instead of `it` use `const nss = rootContainer.includes("demo-ess") ? it.skip : it`, and then do `nss("can create and remove empty Containers", async () => {` - so it is explicitly marked as skipped.)

This is assuming that Jest doesn't do weird magic that prevents you from actually passing around `it()`.

src/e2e.test.ts
```diff
@@ -93,32 +94,34 @@ describe("End-to-end tests", () => {
...
+      // FIXME: An ESS bug regading PUTting containes prevents this test from passing.
+      // Once the bug is fixed, this should be cleaned up.
+      const newContainer1 = await createContainerAt(
+        `${rootContainer}container-test/some-container/`
```
Vinnl, line 111: I'm assuming this is a remnant from attempting to run these against ESS anyway, but https://ldp.demo-ess.inrupt.com/105177326598249077653/test-data/container-test/ does contain `some-container`, it seems :)

src/e2e.test.ts
NSeydoux: Great idea, Jest doesn't complain about it: 876d84d8cddbe69167828c246f3e7b7d7e6eddb9

src/e2e.test.ts
NSeydoux: Good catch, I just removed it.

src/e2e.test.ts
Vinnl: Very cool.

image

Vinnl: (Very cool if you're a geek and interested in this sort of thing, I guess.)

NSeydoux: Who isn't ?

PR #367, opened 2020-08-18
Refactor Docs: Separate API docs

kay-kim:
# Refactor Docs: Separate API docs

This commit separates API docs (docs.inrupt.com/client-libraries/api/javascript/solid-client)  from the Library reference docs (docs.inrupt.com/client-libraries/solid-client-js) 

With this change:
- API docs are built separately from the reference docs. 
  - Only API docs requires myST.
  - API docs are not editable.

- API page on the reference docs just contains links to the API docs.
  - Removed API-only myST package from requirements.

- Updated the workflow for the 2 distinct make htmls.

Plan of action to ensure we don't strand users once this commit is merged:

1. Will publish the api docs first.
2. Publish the reference site along with redirects for the current API pages. 
This should ensure that thing don't break.

Afterwards, I'll work on creating a site that uses docs content from the three library repos to  create a unified javascript site.


Vinnl (commented):
Thanks @kay-kim! I'm currently moving everything in `docs/` to `docs/prose/` and the API docs from `apidocs/` to `docs/api`, so that the build setups for the docs are together, and to fix the preview deployments. Let me know if you foresee any problems with that.

I'll let you know when that's done, so you can have a final look at what I changed. Other than that, I think this is good to go.

Btw, maybe not the right place to ask, but do you know where we can define redirects if we rename our modules? (Which would mean the links in our API docs change as well.)

docs/prose/source/api.rst
```diff
@@ -0,0 +1,8 @@
+=============
+API Reference
+=============
+
+.. toctree::
+
+   solid-client API <https://docs.inrupt.com/client-libraries/api/javascript/solid-client>
```
Vinnl, line 7: Do you know if there is a way to use `:apisolidclient:` here as well, so that the link is also correct in the Preview deployment? (No biggie if not; just decreases the likelihood of confusion in the future.)

docs/prose/conf.py
```diff
@@ -56,8 +59,9 @@
     'myst_parser',
 ]
 
+apidocsbasepath = ('https://' + os.environ.get('VERCEL_URL') + '/api/api%s', '') if 'VERCEL_URL' in os.environ else ('https://docs.inrupt.com/client-libraries/api/js/solid-client%s','')
```
Vinnl, line 62: Related to the other question: if the API docs go into an API subdirectory, then I think this would need to be:

```suggestion
apidocsbasepath = ('https://' + os.environ.get('VERCEL_URL') + '/api/api%s', '') if 'VERCEL_URL' in os.environ else ('https://docs.inrupt.com/client-libraries/api/js/solid-client/api%s','')
```

Additionally, these links do not include the `.html` extension, whereas Sphinx _does_ seem to add them? To be consistent, we could use them as well, as in 76bc593a5de29837e425c083dd68db0c289466ae?

Vinnl (approved):
@kay-kim Alright, you can see my changes in 2f27f14fb172628960e1b6f976fd23e0284d33be (and fe8789f908728a1253ddc057f3a8d5e6014e9840 to add the `.html` extension to links to the API docs).

I still have a couple of open challenges that I added inline below.

docs/prose/source/api.rst
kay-kim: Unfortunately, I don't think we can in a toctree.

package.json
```diff
@@ -11,7 +11,8 @@
     "list-licenses": "license-checker --production --csv --out LICENSE_DEPENDENCIES_ALL",
     "check-licenses": "license-checker --production --failOn \"AGPL-1.0-only; AGPL-1.0-or-later; AGPL-3.0-only; AGPL-3.0-or-later; Beerware; CC-BY-NC-1.0; CC-BY-NC-2.0; CC-BY-NC-2.5; CC-BY-NC-3.0; CC-BY-NC-4.0; CC-BY-NC-ND-1.0; CC-BY-NC-ND-2.0; CC-BY-NC-ND-2.5; CC-BY-NC-ND-3.0; CC-BY-NC-ND-4.0; CC-BY-NC-SA-1.0; CC-BY-NC-SA-2.0; CC-BY-NC-SA-2.5; CC-BY-NC-SA-3.0; CC-BY-NC-SA-4.0; CPAL-1.0; EUPL-1.0; EUPL-1.1; EUPL-1.1;  GPL-1.0-only; GPL-1.0-or-later; GPL-2.0-only;  GPL-2.0-or-later; GPL-3.0; GPL-3.0-only; GPL-3.0-or-later; SISSL;  SISSL-1.2; WTFPL\"",
     "lint": "eslint --config .eslintrc.js --fix",
-    "build-api-docs": "npx typedoc --out docs/source/api"
+    "build-api-docs": "npx typedoc --out docs/api/source/api",
```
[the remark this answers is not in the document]
kay-kim: So, because source contains index.rst, typedoc seems to have issues outputting to a directory that contains documents not created by typedoc.  
Although, now that I think about it, in the Makefile, during the copy, we can move these to a top-level.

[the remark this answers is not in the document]
kay-kim: Given your absolutely correct comment below, that probably should be done.  That is in the 
Makefile:

	cp -R $(SOURCEDIR)/index.rst $(SOURCECOPYDIR)
	cp -R $(SOURCEDIR)/api/modules/ $(SOURCECOPYDIR)/modules/
	cp -R $(SOURCEDIR)/api/README.md $(SOURCECOPYDIR)/

...

ifdef IS_MAC
	find $(SOURCECOPYDIR)/modules/ -type f -name "*.md" -exec sed -i '' -e 's?\.md#?\.html#?g' {} \;
else
	find $(SOURCECOPYDIR)/modules/ -type f -name "*.md" -exec sed -i  -e 's?\.md#?\.html#?g' {} \;
endif

Then in  the index.rst, we just update the toc for the new directory structure:

.. toctree::
   :glob:
   :titlesonly:

   *
   /modules/*

docs/prose/conf.py
kay-kim: Ah - you are absolutely correct.  And it probably shouldn't go in the api directory - to avoid repetition of api in the site structure.

kay-kim:
This is great!  Thank you for making this correct.  

I should have created two separate workflows, but was thinking more in terms of the workflow as a check for avoiding build errors - totally wrong of me I know.   

Because during publishing, prose and api docs won't necessarily need to be always built together.

kay-kim:
As for redirects, what I had planned on doing was keeping a redirects file

/old/module/foo -> /new/module/bar

Then, in the Makefile, there will be a new target redirects, which would then create the appropriate redirects - which will be part of the build and so, that as part of the deploy, it'll deploy the site pages followed by the redirects.

I'll send an email to you an NicolasS to provide some insight.

PR #370, opened 2020-08-20
Add docs redirects.txt file

kay-kim:
Add redirects file in prep for site reorg


kay-kim:
Closing for now. I think this particular set of redirects belong with the repo that builds the new javascript site (I think).

PR #375, opened 2020-08-20
Docs tweak for includes

kay-kim:
To allow for include in the unified sites - 
- Commit 1 adds a snippet to install just the solid-client lib
- Commit 2 adds comments in the index.rst so that we can include parts of the file in other files.

### DIFF
--- a/docs/prose/source/includes/install-solid-client.rst
+++ b/docs/prose/source/includes/install-solid-client.rst
@@ -1,0 +1,5 @@
+You can use `npm <https://www.npmjs.com/>`__ to install |product|:
+
+.. code-block:: sh
+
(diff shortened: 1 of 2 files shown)


Vinnl (approved):
If this allows us to include just a section of `index.rst`, shouldn't we then just remove `includes/table-issues-help.rst`?

kay-kim:
as in, include  the content from table-issues-help.rst file directly into the index.rst and then include sections of the index.rst  in other files? 
In this case, we could - but in general, in the long run, I've found that having the smaller include files turn out way better.

Vinnl:
OK, let's keep it this way then :)

PR #377, opened 2020-08-21
Support legacy predicate `acl:defaultForNew`

NSeydoux:
This supersedes #321, adding a couple of tests to validate that the code behaves appropriately when dealing with `acl:defaultForNew`. In particular, when setting default access, these tests ensure that `acl:defaultForNew` predicates aren't preserved, which could lead to privilege escalation where even if one tries to set a new default acces, an Agent (or the public) could keep a pre-existing default access if it was specified with `acl:defaultForNew`. 

Going through the code, nowhere `acl.default` is present without an associated `acl.defaultForNew`, which suggests that all our use cases are covered.

# Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
   solid-client data structures in your unit tests.
 - `getFileWithAcl`: like `getSolidDatasetWithAcl`, this function lets you fetch a file along with
   its ACLs, if available.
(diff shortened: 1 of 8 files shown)


Vinnl (approved):
Looks good, great work @NSeydoux and @travis! I added one test to verify that we're not _writing_ new `defaultForNew` predicates (since they're deprecated and all major servers support `default`): 58d7d9cd431f861006678508edc04dea9b9e7b1c If Nicolas can review those, then with a couple of squashes this this looks good to go to me.

PR #386, opened 2020-08-25
Trigger CI on PR

NSeydoux:
In order to have the CI run for third-party contributions coming from a fork, they must run on PR, and not only on push, since technically a third-party contribution hasn't been pushed to a branch of the repo.

This should make sure the CI runs in cases like https://github.com/inrupt/solid-client-authn-js/pull/277.


Vinnl:

(In this case, I think there's still the question that it _should_ still work on forks (because they push as well), but that fork still needs to explicitly allow GitHub Actions to run (because that will have access to their secrets). Likewise, if they don't do that, then this change won't make them run either, since our repository also just won't run modified code of third parties (because that will have access to our secrets).)

PR #387, opened 2020-08-26
DOCS-121: review acl/agents docs

kay-kim:
Review acl/agents API docs.

Am chunking the review of the API docs into pieces -- so that I don't have one really huge branch -- as well as so that I can switch to other docs work in-between.

1. Commit 1: Initial updates to acl/acl API docs.
2. Commit 2: Updates per feedback.
3. Commit 3: Updates per feedback.
4. Commit 4: Update acl/agents API docs.
5. Commit 5: Add @returns to the agents setAgentAccess functions API.

### DIFF
--- a/src/acl/acl.ts
+++ b/src/acl/acl.ts
@@ -142,20 +142,17 @@ export function internal_getContainerPath(resourcePath: string): string {
 }
 
 /**
- * Verify whether an ACL was found for the given Resource.
+ * Verify whether the given Resource has a resource ACL attached.
  *
- * A Resource fetched with its ACL (e.g. using [[getSolidDatasetWithAcl]]) _might_ have a resource ACL attached, but
- * we cannot be sure: it might be that none exists for this specific Resource (in which case the
- * fallback ACL applies), or the currently authenticated user (if any) might not have Control access
- * to the fetched Resource.
+ * A Resource returned using [[getSolidDatasetWithAcl]], i.e., the returned `dataset,
+ * has its resource ACL attached if its resource ACL exists and the current user
+ * has Control access to the Resource. Otherwise, the Resource will not have a resource ACL attached.
  *
- * This function verifies that the Resource's ACL is accessible.
- *
- * Please note that the Web Access Control specification is not yet finalised, and hence, this
- * function is still experimental and can change in a non-major release.
+ * _Note_: The Web Access Control specification is not yet finalised. As such, this
+ * function is still experimental and subject to change.
  *
  * @param resource A Resource that might have an ACL attached.
- * @returns Whether `dataset` has an ACL attached.
+ * @returns Whether the Resource has a resource ACL attached.
  */
 export function hasResourceAcl<Resource extends WithAcl & WithResourceInfo>(
   resource: Resource
@@ -170,16 +167,16 @@ export function hasResourceAcl<Resource extends WithAcl & WithResourceInfo>(
 }
 
 /**
- * Access the ACL attached to a Resource.
+ * Return the resource ACL attached to a Resource.
  *
- * Given a Resource that has an ACL attached, this function will give you access to that ACL. To
- * verify whether the ACL is available, see [[hasResourceAcl]].
+ * Given a Resource that has its Resource ACL attached and is accessible by the user
+ * (see [[hasResourceAcl]]), this function returns the Resource ACL.
  *
- * Please note that the Web Access Control specification is not yet finalised, and hence, this
- * function is still experimental and can change in a non-major release.
+ * _Note_: The Web Access Control specification is not yet finalised. As such, this
+ * function is still experimental and subject to change.
  *
  * @param resource A Resource with potentially an ACL attached.
- * @returns The ACL, if available, and undefined if not.
+ * @returns The resource ACL if available and `null` if not.
  */
 export function getResourceAcl(
   resource: WithAcl & WithResourceInfo & WithResourceAcl
@@ -197,19 +194,18 @@ export function getResourceAcl(
 }
 
 /**
- * Verify whether a fallback ACL was found for the given Resource.
+ * Verify whether the given Resource has a fallback ACL attached.
  *
- * A Resource fetched with its ACL (e.g. using [[getSolidDatasetWithAcl]]) _might_ have a fallback ACL
- * attached, but we cannot be sure: the currently authenticated user (if any) might not have Control
- * access to one of the fetched Resource's Containers.
- *
+ * A Resource returned using [[getSolidDatasetWithAcl]], i.e., the returned `dataset,
+ * has a fallback ACL attached if the current user
+ * has Control access to the Resource's Container from which the Resource inherits its ACL.
  * This function verifies that the fallback ACL is accessible.
  *
- * Please note that the Web Access Control specification is not yet finalised, and hence, this
- * function is still experimental and can change in a non-major release.
+ * _Note_: The Web Access Control specification is not yet finalised. As such, this
+ * function is still experimental and subject to change.
  *
  * @param resource A [[SolidDataset]] that might have a fallback ACL attached.
- * @returns Whether `dataset` has a fallback ACL attached.
+ * @returns Whether the Resource has a fallback ACL attached.
  */
 export function hasFallbackAcl<Resource extends WithAcl>(
   resource: Resource
@@ -218,16 +214,17 @@ export function hasFallbackAcl<Resource extends WithAcl>(
 }
 
 /**
- * Access the fallback ACL attached to a Resource.
+ * Return the fallback ACL attached to a Resource.
  *
- * Given a Resource that has a fallback ACL attached, this function will give you access to that
- * ACL. To verify whether the fallback ACL is available, see [[hasFallbackAcl]].
+ * Given a Resource that has a fallback ACL attached and is accessible by the user
+ * (see [[hasFallbackAcl]]), this function returns the Resource ACL.
  *
- * Please note that the Web Access Control specification is not yet finalised, and hence, this
- * function is still experimental and can change in a non-major release.
+ *
+ * _Note_: The Web Access Control specification is not yet finalised. As such, this
+ * function is still experimental and subject to change.
  *
  * @param resource A Resource with potentially a fallback ACL attached.
- * @returns The fallback ACL, or null if it coult not be accessed.
+ * @returns The fallback ACL if available or `null` if not.
  */
 export function getFallbackAcl(resource: WithFallbackAcl): AclDataset;
 export function getFallbackAcl(dataset: WithAcl): AclDataset | null;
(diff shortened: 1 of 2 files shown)


src/acl/acl.ts
```diff
@@ -142,20 +142,17 @@ export function internal_getContainerPath(resourcePath: string): string {
...
- * This function verifies that the Resource's ACL is accessible.
- *
- * Please note that the Web Access Control specification is not yet finalised, and hence, this
- * function is still experimental and can change in a non-major release.
+ * _Note_: The Web Access Control specification is not yet finalised. As such, this
+ * function is still experimental and subject to change.
  *
  * @param resource A Resource that might have an ACL attached.
- * @returns Whether `dataset` has an ACL attached.
+ * @returns Whether the Resource has a resource ACL attached.
```
kay-kim, line 155: So, I kind of wanted to update to  @returns `true` if the Resource has a ...
because the function states 
`
return ( resourceAcl !== null && ... )  
`
but I got a little confused because we put the following
`
: resource is Resource & WithResourceAcl & WithAccessibleAcl
`, so that typedoc states it returns a Resource  followed by the @returns sentence.

src/acl/acl.ts
```diff
@@ -142,20 +142,17 @@ export function internal_getContainerPath(resourcePath: string): string {
...
+ * A Resource returned using [[getSolidDatasetWithAcl]], i.e., the returned `dataset,
+ * has its resource ACL attached if its resource ACL exists and the current user
+ * has Control access to the Resource. Otherwise, the Resource will not have a resource ACL attached.
  *
- * This function verifies that the Resource's ACL is accessible.
- *
- * Please note that the Web Access Control specification is not yet finalised, and hence, this
- * function is still experimental and can change in a non-major release.
+ * _Note_: The Web Access Control specification is not yet finalised. As such, this
+ * function is still experimental and subject to change.
```
kay-kim, line 152: So, I wasn't sure when you were stating that the function can change in a non-major release - whether that was exclusively only in non-major releases or in non-major releases as well as a major release.  So, I just reduced it to subject to change.

But, if it's explicitly non-major releases, let me know - because that is something to call attention to.

src/acl/acl.ts
```diff
@@ -170,16 +167,16 @@ export function hasResourceAcl<Resource extends WithAcl & WithResourceInfo>(
...
+ * (see [[hasResourceAcl]]), this function returns the Resource ACL.
  *
- * Please note that the Web Access Control specification is not yet finalised, and hence, this
- * function is still experimental and can change in a non-major release.
+ * _Note_: The Web Access Control specification is not yet finalised. As such, this
+ * function is still experimental and subject to change.
  *
  * @param resource A Resource with potentially an ACL attached.
- * @returns The ACL, if available, and undefined if not.
+ * @returns The resource ACL if available and `null` if not.
```
kay-kim, line 179: Same thing here:

We seem to return either null or the resourceAcl

but we have 
`
export function getResourceACL(): AclDataset
`
so that the typedoc states it returns AclDataset followed by the above sententce that this returns the resource ACL.

src/acl/acl.ts
```diff
@@ -142,20 +142,17 @@ export function internal_getContainerPath(resourcePath: string): string {
...
 /**
- * Verify whether an ACL was found for the given Resource.
+ * Verify whether the given Resource has a resource ACL attached.
  *
- * A Resource fetched with its ACL (e.g. using [[getSolidDatasetWithAcl]]) _might_ have a resource ACL attached, but
- * we cannot be sure: it might be that none exists for this specific Resource (in which case the
- * fallback ACL applies), or the currently authenticated user (if any) might not have Control access
- * to the fetched Resource.
+ * A Resource returned using [[getSolidDatasetWithAcl]], i.e., the returned `dataset,
+ * has its resource ACL attached if its resource ACL exists and the current user
```
kay-kim, line 148: Should I clarify that when getSolidDatasetWithAcl - must also be called with the Control access in case two different users call getSolidDatasetWithAcl and hasResourceAcl?

src/acl/acl.ts
```diff
@@ -197,19 +194,18 @@ export function getResourceAcl(
...
- * Verify whether a fallback ACL was found for the given Resource.
- *
- * A Resource fetched with its ACL (e.g. using [[getSolidDatasetWithAcl]]) _might_ have a fallback ACL
- * attached, but we cannot be sure: the currently authenticated user (if any) might not have Control
- * access to one of the fetched Resource's Containers.
+ * Verify whether the given Resource has a fallback ACL attached.
  *
+ * A Resource returned using [[getSolidDatasetWithAcl]], i.e., the returned `dataset,
+ * has a fallback ACL attached if the current user
+ * has Control access to the Resource's Container from which the Resource inherits its ACL.
```
kay-kim, line 201: Couple of questions since I don't think the above captures the situation completely.

1. If a Resource has its own resourceAcl and its container has also set the fallbackAcl, getSolidDatasetWithAcl would return both resourceAcl and fallbackAcl? 
If so, I should clarify "from which the Resource inherits its ACL if the Resource does not have its own resourceAcl" or something like that.

2. Also, if a resource is using its container's fallbackAcl, but the user calling getSolidDatasetWithAcl has Control access to the container's container's fallbackAcl - what is returned in the fallbackAcl?

src/acl/acl.ts
```diff
@@ -142,20 +142,17 @@ export function internal_getContainerPath(resourcePath: string): string {
 }
 
 /**
- * Verify whether an ACL was found for the given Resource.
+ * Verify whether the given Resource has a resource ACL attached.
```
kay-kim, line 145: Just as an FYI --  some people go by the rule that the description should be a declarative sentence "Verifies ..." rather than imperative "Verify ..."

I've never been particularly strict about it one way or the other as long as people are consistent - as I've never found people to be confused by the use of one over the other.  My rule is if everything else is clear, no one cares as long as we're consistent.

src/acl/acl.ts
```diff
@@ -142,20 +142,17 @@ export function internal_getContainerPath(resourcePath: string): string {
...
- * to the fetched Resource.
+ * A Resource returned using [[getSolidDatasetWithAcl]], i.e., the returned `dataset`,
+ * has its resource ACL attached if its resource ACL exists and the current user
+ * has Control access to the Resource. Otherwise, the Resource will not have a resource ACL attached.
  *
- * This function verifies that the Resource's ACL is accessible.
- *
- * Please note that the Web Access Control specification is not yet finalised, and hence, this
- * function is still experimental and can change in a non-major release.
+ * _Note_: The Web Access Control specification is not yet finalised. As such, this
```
kay-kim, line 151: If you don't like it as a note, can revert.  Just wasn't sure how to draw attention to it.

src/acl/acl.ts
NSeydoux: This type signature is a Typescript type guard, so it does return a boolean, and this boolean is true if the `resource` argument had properties for the types `Resource`, `WithResourceAcl`, `WithAccessibleAcl`. The `@return` could be `true if the Resource has a resource ACL attached`.

src/acl/acl.ts
NSeydoux: Technically, as long as the major version is 0, any release can be breaking, but that's standard semver. It was specifically called out here because we know that the ACL API is particularly likely to change.

src/acl/acl.ts
NSeydoux: `AclDataset` is the type of any ACL, so both a resource ACL and a fallback ACL are of type `AclDataset`. I'm not sure that's what you are asking though :)

src/acl/acl.ts
NSeydoux: `getSolidDatasetWithAcl` may be called without control access, but then the ACL won't be available to the caller. I think it's one of the tricky parts of the code, so I think you can clarify that as often as possible :)

src/acl/acl.ts
NSeydoux: 1. Yes, if both ACLs exist and are accessible for a given resource, both are returned by `getSolidDatasetWithAcl`.
2. The fallback ACL is the ACL of the first container `C` in the hierarchy that has a `default` ACL, so the `default` ACLs of  containers parent of `C` don't apply to children resources of `C`.

src/acl/acl.ts
Vinnl: Yeah, it could be in any release, whereas usually, breaking changes are only introduced in major releases. Thus, the comment is there to emphasise that usual expectations about breaking changes being limited to major releases do not hold for this function.

src/acl/acl.ts
Vinnl: Whoops, yes, the `undefined` should be `null` (that was changed at some point).

There are multiple instances of `export function getResourceAcl` - these are type overloads. The first says that if you pass in a resource of which we know that it has an ACL attached (i.e. `WithResourceAcl` applies), you will definitely get an `AclDataset` and not `null`. However, if we do not know for sure that it also has a Resource ACL, we might still return `null`.

So yes, a description that says it returns the Resource ACL if it is available, and `null` if not` is correct.

src/acl/acl.ts
Vinnl: No strong preference here, and agreed with the consistency - feel free to change it to whatever you feel is best (and let us know which form to use in the future).

src/acl/acl.ts
Vinnl: No objection to using `_Note_:` from my side.

src/acl/acl.ts
kay-kim: Yup -- 
but the typedoc outputs the following markdown

```
**Returns:** *resource is Resource & WithFallbackAcl*

Whether the Resource has a fallback ACL attached.

```

So, I'll update @returns to `true if ...`

but it'll output

```
**Returns:** *resource is Resource & WithFallbackAcl*

`true` if the Resource ...

```

I wasn't sure if there was a way that typedoc didn't output the first part.

src/acl/acl.ts
kay-kim: Cool. Didn't realize that ACL was type AclDataset.

src/acl/acl.ts
kay-kim: Got it -  getSolidDatasetWithAcl will return not have the fallbackAcl if called by someone w/o control access to container C but only to the parents of C.

src/acl/acl.ts
kay-kim: I'll update to using declarative as I go through the API docs since sooner or later, we'll get feedback from a user stating we should be using the declarative - and we should going forward use the declarative  -- just to be consistent not just within this library but across libraries.

src/acl/acl.ts
kay-kim: When I get to the getSolidDatasetWithAcl method, I'll go into deeper details there - and then maybe loop back in here, once we're good with how we want to describe getSolidDatasetWithAcl

src/acl/acl.ts
kay-kim: Added back the non-major release.

kay-kim:
Am chunking the review of the API docs into pieces - this way, I can switch to other docs work in-between.

Vinnl (approved):
Ah, I thought GitHub would email us when this was marked as ready for review, but it doesn't look like it did. Don't think there are any major issues, so I approved it.

PR #417, opened 2020-09-02
Docs: Acl/Groups API

kay-kim:
Updates to the ACL/Groups api docs.  To better highlight the notes, am now using the note admonition and moved the note up.  If acceptable, will proceed with that model for the other API docs.

### DIFF
--- a/src/acl/group.ts
+++ b/src/acl/group.ts
@@ -45,15 +45,16 @@ import {
 import { acl } from "../constants";
 
 /**
- * Find out what Access Modes have been granted to a given Group of agents specifically for a given Resource.
+ * _Note_: This function is still experimental and subject to change, even in a non-major release.
  *
- * Keep in mind that this function will not tell you what access members of the given Group have through other ACL rules, e.g. public permissions.
+ * Returns a Group's explicity-granted Access Modes for a given Resource.
  *
- * Also, please note that this function is still experimental: its API can change in non-major releases.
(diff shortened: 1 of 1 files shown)


src/acl/group.ts
```diff
@@ -45,15 +45,16 @@ import {
 import { acl } from "../constants";
 
 /**
- * Find out what Access Modes have been granted to a given Group of agents specifically for a given Resource.
+ * ```{note} This function is still experimental and subject to change, even in a non-major release.
```
Vinnl, line 48: Huh, that's cool that that works. :+1: from me.

src/acl/group.ts
```diff
@@ -93,17 +96,20 @@ export function getGroupAccessAll(
...
 /**
- * Given an ACL SolidDataset, find out which access modes it provides to a Group for its associated Resource.
+ * ```{note} This function is still experimental and subject to change, even in a non-major release.
+ * ```
  *
- * Keep in mind that this function will not tell you:
- * - what access members of the given Group have through other ACL rules, e.g. public permissions.
- * - what access members of the given Group have to child Resources, in case the associated Resource is a Container (see [[getGroupDefaultAccessModes]] for that).
+ * Returns the Access Modes explicitly granted to a Group for the Resource
+ * associated with an ACL (Access ControlList).
```
Vinnl, line 103:
```suggestion
 * associated with an ACL (Access Control List).
```

Vinnl (approved):

PR #423, opened 2020-09-04
Docs: Resource/NonRDFData API

kay-kim:
Updates to the resource/nonRDFData API.
 
### DIFF
--- a/src/resource/nonRdfData.ts
+++ b/src/resource/nonRdfData.ts
@@ -92,14 +92,14 @@ export async function getFile(
  *
  * If the user calling the function does not have access to the file's resource ACL,
  * [[hasAccessibleAcl]] on the returned blob returns false.
- * If the user has access to the file's resource ACL but the resource ACL,
+ * If the user has access to the file's resource ACL but the resource ACL does not exist,
  * [[getResourceAcl]] on the returned blob returns null.
  * If the fallback ACL is inaccessible by the user,
  * [[getFallbackAcl]] on the returned blob returns null.
  *
  * ```{tip}
  * To retrieve the ACLs, the function results in multiple HTTP requests rather than a single
- * request as mandated by the Solid spec. As such, prefer [[getFile]] instead.
+ * request as mandated by the Solid spec. As such, prefer [[getFile]] instead if you do not need the ACL.
  * ```
  *
  * @param url The URL of the fetched file
--- a/src/resource/nonRdfData.ts
+++ b/src/resource/nonRdfData.ts
@@ -50,12 +50,14 @@ function containsReserved(header: Headers): boolean {
 }
 
 /**
- * Gets a file at a given URL, and returns it as a blob of data.
+ * ```{note} This function is still experimental and subject to change, even in a non-major release.
+ * ```
  *
- * Please note that this function is still experimental: its API can change in non-major releases.
+ * Retrieves a file from a URL and returns the file as a blob.
  *
- * @param url The URL of the fetched file
+ * @param url The URL of the file to return
  * @param options Fetching options: a custom fetcher and/or headers.
+ * @returns The file as a blob.
  */
 export async function getFile(
   input: Url | UrlString,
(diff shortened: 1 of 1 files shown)


src/resource/nonRdfData.ts
```diff
@@ -82,23 +84,27 @@ export async function getFile(
...
+ * [[hasAccessibleAcl]] on the returned blob returns false.
+ * If the user has access to the file's resource ACL but the resource ACL,
+ * [[getResourceAcl]] on the returned blob returns null.
+ * If the fallback ACL is inaccessible by the user,
+ * [[getFallbackAcl]] on the returned blob returns null.
  *
- * If the Resource does not advertise the ACL Resource (because the authenticated user does not have
- * access to it), the `acl` property in the returned value will be null. `acl.resourceAcl` will be
- * undefined if the Resource's linked ACL Resource could not be fetched (because it does not exist),
- * and `acl.fallbackAcl` will be null if the applicable Container's ACL is not accessible to the
```
kay-kim, line 96: instead of direct references to acl and its properties, I ended up using hasAccessibleAcl/getResourceAcl/getFallbackACL instead.  Let me know, I can put them back or put in addition to.

src/resource/nonRdfData.ts
```diff
@@ -147,17 +153,15 @@ type SaveFileOptions = GetFileOptions & {
...
+ * not be the given `slug`.
  *
- * Please note that this function is still experimental: its API can change in non-major releases.
- *
- * @param folderUrl The URL of the folder where the new file is saved
- * @param file The file to be written
- * @param options Additional parameters for file creation (e.g. a slug)
+ * @param folderUrl The URL of the folder where the new file is saved.
+ * @param file The file to be written.
+ * @param options Additional parameters for file creation (e.g. a slug).
```
kay-kim, line 164: Could not get this to work.  I keep getting 401 Unauthenticated even though I passed in the fetch
`saveFileInContainer(
            'https://kay.inrupt.net/test',
            new Blob([textPostFile.text], { type: 'plain/text'}, { fetch: session.fetch })
  );`

Only tried the saveFileInContainer.

src/resource/nonRdfData.ts
```diff
@@ -147,17 +153,15 @@ type SaveFileOptions = GetFileOptions & {
...
 /**
- * Saves a file in a folder at a given URL. The server will return the final
- * filename (which may or may not be the given `slug`), it will return it in
- * the response's Location header.
+ * ```{note} This function is still experimental and subject to change, even in a non-major release.
+ * ```
  *
- * If something went wrong saving the file, the returned Promise will be rejected with an Error.
+ * Saves a file in a folder associated with the given URL. The final filename may or may
+ * not be the given `slug`.
```
kay-kim, line 160: Do you think it might help to state that we do a POST

src/resource/nonRdfData.ts
```diff
@@ -82,23 +84,27 @@ export async function getFile(
...
+ * and its fallback ACL from a URL and returns them as a blob.
  *
- * This is an experimental function that gets both a file, the linked ACL Resource (if
- * available), and the ACL that applies to it if the linked ACL Resource is not available. This can
- * result in many HTTP requests being executed, in lieu of the Solid spec mandating servers to
- * provide this info in a single request. Therefore, and because this function is still
- * experimental, prefer [[getFile]] instead.
+ * If the user calling the function does not have access to the file's resource ACL,
+ * [[hasAccessibleAcl]] on the returned blob returns false.
+ * If the user has access to the file's resource ACL but the resource ACL,
```
NSeydoux, line 95:
```suggestion
 * If the user has access to the file's resource ACL but the resource ACL does not exist,
```

src/resource/nonRdfData.ts
```diff
@@ -82,23 +84,27 @@ export async function getFile(
...
+ * [[getFallbackAcl]] on the returned blob returns null.
  *
- * If the Resource does not advertise the ACL Resource (because the authenticated user does not have
- * access to it), the `acl` property in the returned value will be null. `acl.resourceAcl` will be
- * undefined if the Resource's linked ACL Resource could not be fetched (because it does not exist),
- * and `acl.fallbackAcl` will be null if the applicable Container's ACL is not accessible to the
- * authenticated user.
+ * ```{tip}
+ * To retrieve the ACLs, the function results in multiple HTTP requests rather than a single
+ * request as mandated by the Solid spec. As such, prefer [[getFile]] instead.
```
NSeydoux, line 102:
```suggestion
 * request as mandated by the Solid spec. As such, prefer [[getFile]] instead if you don't need the ACL.
```

src/resource/nonRdfData.ts
NSeydoux: I wouldn't go into that level of details, since what we want is to add an abstraction layer on top of the standard.

src/resource/nonRdfData.ts
NSeydoux: I'll try this locally, but a possible issue is that the container URI must end with a `/`. If that's actually the problem, normalizing the URI in the function might be a good idea (note to myself).

NSeydoux (approved):

src/resource/nonRdfData.ts
Vinnl: The trailing slash does not seem to be required: https://codesandbox.io/s/inruptsolid-client-js-sandbox-forked-8kw7t?file=/src/index.ts

The only difference is that I am not using solid-client-authn's fetch, so presumably the problem lies there?

NSeydoux: Actually even authenticated, I can't reproduce the issue: slash or no slash, the file is created. See https://codesandbox.io/s/hungry-sky-ycos2?file=/src/App.js

kay-kim: Thanks @Vinnl  and @NSeydoux !  Found out why my stuff was goofy.  I've messed up the ending parenthesis for the Blob  - (ugh - it's like being back with Lisp) - it should have been new Blob( [], {} ), { fetch: ... }) instead of  what I had with new Blob( [], {}, { fetch: } ) ).

PR #434, opened 2020-09-07
Remove* conditional expects in tests

Vinnl:
The Jest ESLint plugin added the no-conditional-expect rule to its
recommended ruleset (see
https://github.com/jest-community/eslint-plugin-jest/blob/d16b84c23acda12d95244968313dc60396369b31/CHANGELOG.md#2400-2020-09-04).

This commit cheats a bit: in most cases the expect calls are no
longer in an `if` block, but they can still not be called because
of an `if` block in the code path before it returning.

That said, the early exit condition does feel slightly cleaner.
Since we only need to run the assertions on specific Things that
we have to filter on, I don't think there's a way of preventing the
assertions to be conditional. We could add an expect.assertions(),
but since we often have a couple of if conditions that match on the
same rule (e.g. isAgentGroupRule, isAgentClassRule, etc.), I felt
that it would just be more confusing to have fewer assertions than
conditions, so I felt the risk of the assertions somehow no longer
being called in the future was worth it.

This will allow CI to succed for #432.


Vinnl:
@NSeydoux The end-to-end tests do not expect there to be an ACL file at https://ldp.demo-ess.inrupt.com/105177326598249077653/test-data/?ext=acl (the NSS equivalent at https://lit-e2e-test.inrupt.net/public/lit-pod-acl-initialisation-test/resource.ttl.acl does not exist either). I do not appear to be able to remove it without being authenticated, can you?

NSeydoux (approved):

PR #435, opened 2020-09-07
Add more context when throwing errors

Vinnl:
# New feature description

When an HTTP request results in an error, the error now contains
more context about what data was being sent (if any), and where it
was sent to.

# Checklist

- [x] The changelog has been updated, if applicable. N/A

### DIFF
--- a/src/acl/acl.ts
+++ b/src/acl/acl.ts
@@ -622,7 +622,9 @@ export async function deleteAclFor<
 
   if (!response.ok) {
     throw new Error(
-      `Deleting the ACL failed: ${response.status} ${response.statusText}.`
+      `Deleting the ACL of the Resource at \`${getSourceUrl(
+        resource
+      )}\` failed: ${response.status} ${response.statusText}.`
     );
   }
 
--- a/src/resource/solidDataset.test.ts
+++ b/src/resource/solidDataset.test.ts
@@ -304,12 +304,14 @@ describe("getSolidDataset", () => {
         Promise.resolve(new Response("Not allowed", { status: 403 }))
       );
 
-    const fetchPromise = getSolidDataset("https://arbitrary.pod/resource", {
+    const fetchPromise = getSolidDataset("https://some.pod/resource", {
       fetch: mockFetch,
     });
 
     await expect(fetchPromise).rejects.toThrow(
-      new Error("Fetching the Resource failed: 403 Forbidden.")
+      new Error(
+        "Fetching the Resource at `https://some.pod/resource` failed: 403 Forbidden."
+      )
     );
   });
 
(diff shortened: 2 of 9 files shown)


src/resource/solidDataset.test.ts
```diff
@@ -921,6 +925,92 @@ describe("saveSolidDatasetAt", () => {
...
+        "https://some.pod/resource",
+        mockDataset,
+        {
+          fetch: mockFetch,
+        }
+      );
+
+      await expect(fetchPromise).rejects.toThrow(
+        "Storing the Resource at `https://some.pod/resource` failed: 404 Not Found.\n\n" +
+          "The changes that were sent to the Pod are listed below.\n\n"
```
NSeydoux, line 1011: The expected values for the changelog are not tested for. Should they be ?

NSeydoux (approved):

src/acl/acl.ts
```diff
@@ -622,7 +622,9 @@ export async function deleteAclFor<
 
   if (!response.ok) {
     throw new Error(
-      `Deleting the ACL failed: ${response.status} ${response.statusText}.`
+      `Deleting the ACL of the Resource at \`${getSourceUrl(
```
pmcb55, line 625: Just to note - the convention we use throughout ESS is to delimit variables in messages with '[' and ']' - so I think it makes sense to align with that (i.e. rather than using backticks instead).

pmcb55 (approved):
LGTM

src/resource/solidDataset.test.ts
Vinnl: Yep, that's the case for every test that no longer explicitly tests on `Error` but just on the error message. My reasoning here was that for these tests, the actual contents of the dataset isn't too relevant, as long as it says the status code, URL to be saved on, and that it will display the changes. (The actual changes dataset is empty.)

Whether the dataset's Markdown representation is correct is already covered by another test, and this way, we don't need to update this specific test if we change the human-readable representation of a dataset. That said, if we somehow remove the call to `solidDatasetAsMarkdown()` in the code, this test will not catch that, so... Pros and cons, I guess. My deliberation landed on this approach, but happy to change to the other if you thing those pros/cons outweigh this one's.

src/acl/acl.ts
Vinnl: I don't have a _strong_ preference, but given that the consumers of ESS error messages and solid-client error messages are going to be quite different people, and that the latter have a common culture of marking raw content with backticks, I do have a slight preference for aligning with our ecosystem if that's OK with you.

pmcb55: Yeah, I don't have a strong preference either, the most important thing is simply consistency. Ultimately I'd actually like to see all these messages expressed in RDF (allowing us provide unique identifiers for each of them, allowing for translations, allowing meta-data like 'seeAlso', etc.), and not hard-coded like this at all, and so in fact I don't see particular audiences as being too important. I'll discuss with Jack and see what his take on it is (since whatever we decide should be in the coding guidelines to try and ensure we get that consistency).

src/resource/solidDataset.test.ts
NSeydoux: No, that seems perfectly reasonable. We can always test for the content of the message in an end-to-end test if the need arises, but I agree that having separate tests for the markdownification of a dataset on one hand, and for the contextual info on the other feels like the right call.

PR #442, opened 2020-09-12
introduce an explicit type for internal_defaultFetchOptions

travis:
Because `stripInternal` is set to true and internal_defaultFetchOptions is set to internal, the generated type definitions referred to an undefined symbol, which broke an [ncc](https://www.npmjs.com/package/@zeit/ncc) build I was working on. Introducing an explicit type that can be exported with the type definitions fixed this.

Not sure which procedure below to follow for this, but happy to update if y'all need more to get this merged!

Tested these changes with this branch of the library I'm working on:

https://github.com/itme/swrlit/tree/ncc

This PR fixes #<issue ID>.

# New feature description

# Checklist

This PR bumps the version to <version number>.

# Checklist

- [ ] I used `npm version <major|minor|patch>` to update `package.json`, inspecting the changelog to determine if the release was major, minor or patch.
- [ ] The CHANGELOG has been updated to show version and release date - https://keepachangelog.com/en/1.0.0/.
- [ ] `@since X.Y.Z` annotations have been added to new APIs.
- [ ] The **only** commits in this PR are:
  - the CHANGELOG update.
  - the version update.
  - `@since` annotations.
- [ ] I will make sure **not** to squash these commits, but **rebase** instead.
- [ ] Once this PR is merged, I will push the tag created by `npm version ...` (e.g. `git push origin vX.Y.Z`).

### DIFF
--- a/src/acl/acl.ts
+++ b/src/acl/acl.ts
@@ -50,15 +50,14 @@ import {
   getSourceUrl,
   internal_defaultFetchOptions,
   getResourceInfo,
+  FetchOptions,
 } from "../resource/resource";
 import { addIri } from "..";
 
 /** @internal */
 export async function internal_fetchResourceAcl(
   dataset: WithResourceInfo,
-  options: Partial<
-    typeof internal_defaultFetchOptions
-  > = internal_defaultFetchOptions
+  options: Partial<FetchOptions> = internal_defaultFetchOptions
 ): Promise<AclDataset | null> {
   if (!hasAccessibleAcl(dataset)) {
     return null;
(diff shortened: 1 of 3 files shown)


travis:
For what it's worth I ended up moving away from ncc and back to rollup, which seems to make this error go away. I don't have a strong need for this patch to be merged, but the undefined symbols in the generated declarations do seem odd - could absolutely just be my lack of understanding of deep JS-foo so take that with a grant of salt!

Vinnl:
Thanks @travis! I'm not sure if we're ready to add `DefaultFetchOptions` to our external API (since it's not unlikely that the functions using it will get their own unique Options interfaces later), but this is most definitely a bug. As an alternative solution, I created #445, which simply no longer excludes `internal_defaultFetchOptions` from the type definitions, while still not exposing it in the API docs. I tested it with the `ncc` branch of swrlit, and where `npm run build` fails with current stable solid-client, it succeeds after

    npm install @inrupt/solid-client@fixno-stripInternals

Let me know if anything with that approach is unsatisfactory. I'll close this PR in favour of that one. Thanks again for spotting it!

travis:
Ah, makes sense, and that sounds right to me!

PR #445, opened 2020-09-14
Export value used for external type

Vinnl:
Even though we don't necessarily support using
internal_defaultFetchOptions directly as part of our external API,
its inferred type is used as part of some of our type definitions
(i.e. those that accept a compatible type as options, such as our
fetching functions). Thus, even though it shouldn't be listed in
our API docs, it should be available to TypeScript, if developers
use it.

- [ ] I've added a unit test to test for potential regressions of this bug. N/A (Though eventually an e2e test on the published package would help.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,12 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
   thrown when the Container already exists. Now, as with other Solid servers, an error will be
   thrown if the Container you are trying to create already exists.
 
+### Bugs fixed
+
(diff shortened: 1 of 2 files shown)


pmcb55 (commented):
I don't see an actual code change in here (just the CHANGELOG.md update, and a TSDoc annotation change (from @internal to @ignore). Is that annotation sufficient to fix the missing type declaration and prevent potential compilation errors...?!

Vinnl:
@pmcb55 That is correct, the `@internal` annotation removes it from our type definitions entirely, instead of just from our API docs (which `@ignore` does).

pmcb55 (approved):
LGTM

PR #461, opened 2020-09-21
More helpful error on incorrect saveAclFor() calls

Vinnl:
Developers not using TypeScript will not be warned when calling
saveAclFor() with a Resource that does not expose an ACL URL to
the current user, so they'll have to try it out and see it fail in
the right circumstances. This should make the error message that
then occurs provide more assistance as to how to resolve it.

# Checklist

### DIFF
--- a/src/acl/acl.ts
+++ b/src/acl/acl.ts
@@ -582,6 +582,13 @@ export async function saveAclFor(
     typeof internal_defaultFetchOptions
   > = internal_defaultFetchOptions
 ): Promise<AclDataset & WithResourceInfo> {
+  if (!hasAccessibleAcl(resource)) {
+    throw new Error(
+      `Could not determine the location of the ACL for the Resource at \`${getSourceUrl(
+        resource
+      )}\`; possibly the current user does not have Control access to that Resource. Try calling \`hasAccessibleAcl()\` before calling \`saveAclFor()\`.`
+    );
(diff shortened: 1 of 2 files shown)


src/acl/acl.ts
```diff
@@ -582,6 +582,13 @@ export async function saveAclFor(
     typeof internal_defaultFetchOptions
   > = internal_defaultFetchOptions
 ): Promise<AclDataset & WithResourceInfo> {
+  if (!hasAccessibleAcl(resource)) {
+    throw new Error(
+      `Could not determine the location of the ACL for the Resource at \`${getSourceUrl(
```
pmcb55, line 587: BTW, I discussed the question of which delimiter to use for variables in messages, and we both agreed on using '[' and ']. This complete message though is a good example of where that works well, 'cos the subsequent uses in this message for delimiting function names (e.g. `hasAccessibleAcl()` and `saveAclFor()`) can continue to use backticks (as they aren't variable placeholders), thereby making it clearer which parts of that message actually are 'variables'. I'll add a backlog ticket to search-and-replace all uses of backticks.

pmcb55 (approved):
LGTM

src/acl/acl.ts
Vinnl: OK, sounds good.

PR #464, opened 2020-09-22
Also initialise default rules from a fallback ACL

Vinnl:
When a Resource does not have its own ACL, the default rules from
its closest parent apply to it — and to its children. However, when
initialising a new Resource ACL, we copied over the default rules
from its closest parent to apply to the Resource itself, but *not*
to its children. This PR fixes that: copied-over rules now
apply to the Resource's children as well.


pmcb55 (approved):
LGTM

PR #465, opened 2020-09-22
Return null if getThing() cannot find a Thing

Vinnl:
When `getThing()` cannot find any data, it used to return a new, empty `Thing`. This did not make a lot of sense: if you know a Thing's URL, you'd expect it to contain data - otherwise you'd use `createThing`. Hence, it now returns `null` if no data is found.


pmcb55 (approved):
LGTM

PR #468, opened 2020-09-23
Add deleteSolidDataset and deleteContainer

Vinnl:
# New feature description

This adds `deleteSolidDataset` and `deleteContainer`, analogous to `deleteFile`. Additionally, `deleteFile` now also takes the actual file to delete, rather than just its URL.

# Checklist


src/resource/nonRdfData.test.ts
```diff
@@ -345,7 +346,7 @@ describe("Non-RDF data deletion", () => {
       >;
     };
 
-    fetcher.fetch.mockReturnValue(
+    fetcher.fetch.mockReturnValueOnce(
```
NSeydoux, line 349: Was there a specific motivation to change this ? And should the other instances of mock fetcher be aligned with this ?

src/resource/solidDataset.ts
```diff
@@ -538,6 +565,39 @@ export async function createContainerInContainer(
...
+export async function deleteContainer(
+  container: Url | UrlString | WithResourceInfo,
+  options: Partial<
+    typeof internal_defaultFetchOptions
+  > = internal_defaultFetchOptions
+): Promise<void> {
+  const url = hasResourceInfo(container)
+    ? internal_toIriString(getSourceUrl(container))
+    : internal_toIriString(container);
+  if (!url.endsWith("/")) {
```
NSeydoux, line 582: Shouldn't this use `isContainer` ? Possibly, we could create a new `isContainerUrl` function that would be called by `isContainer`, in order to take a `Url | UrlString` as a parameter instead of a `WithResourceInfo`.

NSeydoux (approved):

src/resource/solidDataset.ts
Vinnl: Yes: e0a107f8c66c7e1c4b5c2cc3f1afb6368e8bdca0

src/resource/nonRdfData.test.ts
Vinnl: Yes, since there's a single mock instance of the `../fetcher` module, mocking the return value permanently also affects the other tests. There weren't any before, IIRC, but now that there are, that became problematic. I'll log a ticket to go over the entire code base to see where else this could change, to avoid it becoming an issue in future tests.

NSeydoux (approved):

PR #478, opened 2020-09-28
Add getPodOwner() and isPodOwner()

Vinnl:
# New feature description

This adds `getPodOwner` and `isPodOwner`, which allow a developer to check who owns the Pod that contains a given Resource, if supported by the Pod server and exposed to the current user.

Note that I flipped the arguments of `isPodOwner()` with what was initially specified, because it’s mostly talking about the given WebID, and thus it’s expected to be the most “stable” argument (i.e. you’d sooner call it multiple times with the given WebID and a different SolidDataset than vice versa).

Also, I (probably) found a bug in the Link header parsing library we use, which I added workarounds for with comments linking to the issue: https://github.com/jhermsmeier/node-http-link-header/issues/22

# Checklist

### DIFF
--- a/src/interfaces.ts
+++ b/src/interfaces.ts
@@ -116,10 +116,18 @@ export type WithResourceInfo = {
      * The URL reported by the server as possibly containing an ACL file. Note that this file might
      * not necessarily exist, in which case the ACL of the nearest Container with an ACL applies.
      *
+     * `linkedResources`, which this property is redundant with, was added later.
+     * Thus, this one will be removed at some point.
+     *
      * @ignore We anticipate the Solid spec to change how the ACL gets accessed, which would result
      *         in this API changing as well.
      */
     aclUrl?: UrlString;
+    /**
+     * An object of the links in the `Link` header, keyed by their `rel`.
+     * @hidden
+     */
+    linkedResources: Record<string, string[]>;
     /**
      * Access permissions for the current user and the general public for this resource.
      *
--- a/src/resource/nonRdfData.ts
+++ b/src/resource/nonRdfData.ts
@@ -34,6 +34,8 @@ import {
   internal_parseResourceInfo,
   internal_fetchAcl,
   getSourceIri,
+  getResourceInfo,
+  isRawData,
 } from "./resource";
 
 type GetFileOptions = {
@@ -165,12 +167,13 @@ type SaveFileOptions = GetFileOptions & {
  * @param folderUrl The URL of the folder where the new file is saved.
  * @param file The file to be written.
  * @param options Additional parameters for file creation (e.g. a slug).
+ * @returns A Promise that resolves to the saved file, if available, or `null` if the current user does not have Read access to the newly-saved file. It rejects if saving fails.
  */
 export async function saveFileInContainer(
   folderUrl: Url | UrlString,
   file: Blob,
   options: Partial<SaveFileOptions> = defaultGetFileOptions
-): Promise<Blob & WithResourceInfo> {
+): Promise<(Blob & WithResourceInfo) | null> {
   const folderUrlString = internal_toIriString(folderUrl);
   const response = await writeFile(folderUrlString, file, "POST", options);
 
@@ -191,12 +194,26 @@ export async function saveFileInContainer(
 
   const blobClone = new Blob([file]);
 
-  return Object.assign(blobClone, {
-    internal_resourceInfo: {
-      sourceIri: fileIri,
-      isRawData: true,
-    },
-  });
+  let resourceInfo: WithResourceInfo;
+  try {
+    resourceInfo = await getResourceInfo(fileIri, options);
+  } catch (e) {
+    return null;
+  }
+
+  if (getSourceIri(resourceInfo) !== fileIri) {
+    throw new Error(
+      `Data integrity error: the server reports a URL of \`${getSourceIri(
+        resourceInfo
+      )}\` for the file saved to \`${fileIri}\`.`
+    );
+  }
+  if (!isRawData(resourceInfo)) {
+    throw new Error(
+      `Data integrity error: the server reports that the file saved to \`${fileIri}\` is not a file.`
+    );
+  }
+  return Object.assign(blobClone, resourceInfo);
 }
 
 /**
@@ -225,11 +242,10 @@ export async function overwriteFile(
 
   const blobClone = new Blob([file]);
 
+  const resourceInfo = internal_parseResourceInfo(response);
+
   return Object.assign(blobClone, {
-    internal_resourceInfo: {
-      sourceIri: fileUrlString,
-      isRawData: true,
-    },
+    internal_resourceInfo: resourceInfo,
   });
 }
 
--- a/src/resource/nonRdfData.ts
+++ b/src/resource/nonRdfData.ts
@@ -241,12 +241,11 @@ export async function overwriteFile(
   }
 
   const blobClone = new Blob([file]);
-
   const resourceInfo = internal_parseResourceInfo(response);
+  resourceInfo.sourceIri = fileUrlString;
+  resourceInfo.isRawData = true;
 
-  return Object.assign(blobClone, {
-    internal_resourceInfo: resourceInfo,
-  });
+  return Object.assign(blobClone, { internal_resourceInfo: resourceInfo });
 }
 
 /**
--- a/src/resource/resource.test.ts
+++ b/src/resource/resource.test.ts
@@ -62,6 +62,9 @@ describe("fetchAcl", () => {
         sourceIri: "https://some.pod/resource",
         isRawData: false,
         aclUrl: "https://some.pod/resource.acl",
+        linkedResources: {
+          acl: ["https://some.pod/resource.acl"],
+        },
       },
     };
 
@@ -81,6 +84,7 @@ describe("fetchAcl", () => {
       internal_resourceInfo: {
         sourceIri: "https://some.pod/resource",
         isRawData: false,
+        linkedResources: {},
       },
     };
 
@@ -114,6 +118,9 @@ describe("fetchAcl", () => {
         sourceIri: "https://some.pod/resource",
         isRawData: false,
         aclUrl: "https://some.pod/resource.acl",
+        linkedResources: {
+          acl: ["https://some.pod/resource.acl"],
+        },
       },
     };
 
@@ -158,6 +165,9 @@ describe("fetchAcl", () => {
         sourceIri: "https://some.pod/resource",
         isRawData: false,
         aclUrl: "https://some.pod/resource.acl",
+        linkedResources: {
+          acl: ["https://some.pod/resource.acl"],
+        },
       },
     };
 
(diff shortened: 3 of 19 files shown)


src/resource/resource.test.ts
```diff
@@ -502,6 +510,52 @@ describe("getResourceInfo", () => {
...
+    );
+
+    const solidDatasetInfo = await getResourceInfo(
+      "https://some.pod/container/resource",
+      { fetch: mockFetch }
+    );
+
+    expect(
+      solidDatasetInfo.internal_resourceInfo.linkedResources
+    ).toBeUndefined();
```
NSeydoux, line 539: Would it make sense that if there are no links defined, the `linkedResources` parameter is an empty map, as opposed to `undefined` ? What would be the most convenient to test for ?

src/resource/resource.ts
```diff
@@ -165,6 +166,13 @@ export function internal_parseResourceInfo(
         resourceInfo.sourceIri
       ).href;
     }
+    // Parse all link headers and expose them in a standard way
+    // (this can replace the parsing of the ACL link above):
+    const linkedResources = parsedLinks.refs.reduce((rels, ref) => {
+      rels[ref.rel] = new URL(ref.uri, resourceInfo.sourceIri).href;
+      return rels;
+    }, {} as Record<string, string>);
```
NSeydoux, line 174: Here, I could see `linkedResources` be `{}` if `linkHeader` is undefined (related to my comment above).

NSeydoux (approved):

src/interfaces.ts
```diff
@@ -116,10 +116,18 @@ export type WithResourceInfo = {
...
+     *
      * @ignore We anticipate the Solid spec to change how the ACL gets accessed, which would result
      *         in this API changing as well.
      */
     aclUrl?: UrlString;
+    /**
+     * An object of the links in the `Link` header, keyed by their `rel`.
+     * @hidden
+     */
+    linkedResources?: Record<string, string>;
```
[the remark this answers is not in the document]
Vinnl: I thought every `rel` was only allowed to be used once, but it seems I misread:

> The rel parameter MUST be present but MUST NOT appear more than once in a given link-value; occurrences after the first MUST be ignored by parsers.

https://httpwg.org/specs/rfc8288.html#header-type

Looks like there should just be a single `rel` _per link_, not _per Link header_. I'll revisit this.

src/resource/resource.test.ts
Vinnl: Yes, good that you bring this up, because that's something we should discuss. I initially went this route, but there was one area where this is problematic: Resources newly created in a Container.

What we currently do is, after creating a Resource in a Container, we manually initialise a new dataset, include the data we just saved in it, and set the ResourceInfo on it. However, this is not possible for `linkedResources` - we can only infer those by actually making a (HEAD) request to the newly-created Resources and inspecting its headers. But then we run into the problem that the current user might not have Read access to the newly-created Resource, e.g. if it was created in a Container like the inbox, to which they only have Append access.

Now that I write this, I think maybe the best solution would be to make that request anyway, and return `null` from the functions that create something in a Container if the current user does not have Read access to it. After all, there's not much you can do with the returned value anyway, and if you want to inspect what data you _sent_ to the Container, then you can just inspect the dataset you passed as input.

What do you think?

NSeydoux: Just to be sure that I understand correctly, your proposal in this case would be to: 
1. Create the resource
2. Make the `HEAD` request to get the data for the `resourceInfo`
3. If the HEAD request fails (i.e. the user doesn't have read access), `null` is returned. Otherwise, the parsing of the `resourceInfo` happens as usual.
4. The `linkedResources` property is always defined for a non-`null` `resourceInfo`, being `{}` if no links are defined, and a `Record<string, string[]>` otherwise (or the type that you see appropriate).

Is that correct ?

Vinnl: Yes, that is correct. (And technically, `{}` is a `Record<string, string[]>` as well, just an empty one :stuck_out_tongue: ) Steps 1-3 would be in `saveInContainer()` and similar functions. `saveSolidDatasetAt()` and the likes (that do not create things in Containers) already get the required data from the Response.

Vinnl: - Returning `null` for Resources created in a Container without Read access: 8e80579af1e2ee67af4fa4d8b6f39983e08a863b
- Don't allow `undefined` for `linkedResources`: 9fa30d11e48e1eb9f30686dba3996bf9e9413ea4

src/interfaces.ts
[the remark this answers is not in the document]
Vinnl: Multiple links per `rel`: 91d99040d10922974b50036f1acc0e7925e0ff0e

src/resource/resource.test.ts
Vinnl: Ah, forgot running the build and missed a couple of things:

- Returning `null` for `createFileInContainer` as well: bb511bd89e30306cd42f0f5614289038ab92c196 and f9fd294824ac51acd4e828ee6b974b1d7f156782
- Some places that stilled had `undefined` for `linkedResources`: 2186b8ae1f57b6da575bbaf95c61d13bbda4ff5f and fcc8faf466a579c3b4148fa10ba1d769cc32da62

src/resource/nonRdfData.ts
```diff
@@ -188,12 +192,14 @@ export async function saveFileInContainer(
 
   const blobClone = new Blob([file]);
 
-  return Object.assign(blobClone, {
-    internal_resourceInfo: {
-      sourceIri: fileIri,
-      isRawData: true,
-    },
-  });
+  try {
```
pmcb55, line 195: This new code block appears a tad dangerous. It seems the change is to perform a HEAD on the actual resource saved at the server, to ensure we retrieve all the 'authoritative' meta-data, so is the assumption that the server may be adding further meta-data (i.e. beyond what we already know here in the client, which is the actual 'location' value and the fact the data should be 'rawData')...?
If so, then the HEAD request seems fine, but rather than immediately overwriting the 'location' and 'isRawData' values from that retrieved meta-data, shouldn't we really be checking that what the server says is the meta-data really does match what we expect it to, e.g.
```
    const resourceInfo = await getResourceInfo(fileIri, options);
    if (resourceInfo.internal_resourceInfo.sourceIri !== fileIri ||
    resourceInfo.internal_resourceInfo.isRawData !== true) {
      throw new Error (`Data integrity error - server meta-data for resource [${fileIri}] out-of-sync with our expectations...`);
  }
```
Also, swallowing any exceptions here and returning 'null' seems quirky (at the very least, if we leave it as-is then I think the function's TSDoc needs to explain what a 'null' return from this function actually means). But I think I can see the logic (i.e. the server has already reported that it has saved the resource correctly without any problem), so I'm not 100% sure that throwing an exception if the HEAD request fails is correct either.
It feels like perhaps this should be two functions (to allow the develop choose the exact behaviour they want) - i.e. one function that attempts to save the resource and returns the `fileIri` value if successful, or throws if not. And a second function that saves the resource and then attempts to explicitly retrieve it's meta-data, returning the full dataset if Ok, or throwing if either the save operation or the meta-data retrieval fails...? That 2nd function would need to be named something like `saveFileInContainerAndGetMetadata()`.

Vinnl: >  so is the assumption that the server may be adding further meta-data (i.e. beyond what we already know here in the client, which is the actual 'location' value and the fact the data should be 'rawData')...?

Yes - for example, which Resources are linked to the newly-created Resource.

> If so, then the HEAD request seems fine, but rather than immediately overwriting the 'location' and 'isRawData' values from that retrieved meta-data, shouldn't we really be checking that what the server says is the meta-data really does match what we expect it to, e.g.

Hmm, I think that that was some confusion on my part. I thought that the Response actually did not include that data, but I mixed it up with the Response of the save operation, not of the HEAD request. I've added the data integrity check: ee230e6c24c8cc364661f3116e27934e835209ae

> (at the very least, if we leave it as-is then I think the function's TSDoc needs to explain what a 'null' return from this function actually means)

Good point, forgot that: 0409896dc85314337dc6c0eb0570d9c34bdfb0c4

> so I'm not 100% sure that throwing an exception if the HEAD request fails is correct either.

I think it's not: the HEAD request failing is an expected condition, i.e. when the user does not have Read access to it.

> It feels like perhaps this should be two functions (to allow the develop choose the exact behaviour they want) - i.e. one function that attempts to save the resource and returns the fileIri value if successful, or throws if not. And a second function that saves the resource and then attempts to explicitly retrieve it's meta-data, returning the full dataset if Ok, or throwing if either the save operation or the meta-data retrieval fails...? That 2nd function would need to be named something like saveFileInContainerAndGetMetadata().

I don't actually think we need a second function: if we just return the saved location from this function, then the developer can simply call `getFile()` afterwards. Additionally, that means that if the user does not have Read access, they will still have the URL to the newly-saved File, which might be useful.

The main downside I see there is that this would make it inconsistent with the other `save*` functions, which do return the saved data. What do you think? (/cc @NSeydoux)

Vinnl: So, additional complication: a `HEAD` on a Container does not tell us whether the new Container is an RDF Resource on NSS: https://github.com/solid/node-solid-server/issues/1481

Thus, I've disabled the check for `createContainerInContainer()` for now: 96fe0bbcf6acfec528a494f022042e2bee2e0b1a

NSeydoux: I'm not against returning the location, but I'd be more inclined to keep the `save*` functions consistent. If I'm understanding correctly, here are the options: 
- The non-rdf data is saved in a place where we have read access, in which case we return the blob and resourceinfo
- The non-rdf data is saved in a place where we don't have read access, in which case we return the location.

I would suggest we  align with the other `save*` functions: either they all return `null`, or all return the location (it's not super-useful right now to have the IRI of an unreadable resource, but we can imagine requesting access to it or some other use case).

Vinnl: Hmm, you bring up an interesting point: the other `save*` functions never return `null`, because the Response always gives us the relevant metadata (even without Read access). Unfortunately, it (or at least, NSS) doesn't return the `Wac-Allow` header, so we cannot tell whether the saved Resource is actually readable. We could of course add another HEAD request to the other functions as well, but that feels like wasteful given it's such a fringe use case.

I wouldn't suggest the options you show: I'd either _always_ return the location (but just for the `InContainer` functions), or return `null` if it's saved somewhere we don't have Read access.

Alternatively, we can just return the dataset without the metadata, instead of `null`.

pmcb55: Yeah, I certainly wouldn't suggest the options Nic shows either - completely different return **_types_** would be very bad!
So it seems to me the simplest thing for now is that all save functions just return the save location, and do not attempt to also retrieve the resource (GET), or even it's meta-data (HEAD). That way everything is consistent and simple. If you do want the 'actual' resource data as saved on the server, or you do want the meta-data (or both), then you need to explicitly ask for that using the location you get back from the `save*` operation. That might fail with a 401 (or whatever) if you don't have read access, but that's correct and easy-to-understand behaviour.
If we want more convenience functions (i.e. that do attempt to do 'multiple things in one call' (i.e. save a resource, and then retrieve it again from the server, with or without meta-data) then we can add new functions for those things separately (as they are a higher level of abstraction, with far more complex failure possibilities).
So yeah, it would mean **_not_** returning the dataset at all in our current `save*` functions. But actually that does feel more correct to me anyway - i.e. currently we're kinda giving the illusion that these functions might be returning a 'different' dataset than was passed into them. (Would be good to check first though with Jack to see if that would have significant implications on his side (I doubt it would, but...))
I'm kinda concerned that the client side could easily get out-of-sync with the server side if we don't immediately fetch a resource+it's meta-data directly after asking our SDK to do a `save*` operation, but then again, **_any_** client's view of Pod data is **_always_** potentially out-of-sync with the server side anyway (since I could have 15 automated bots constantly making changes to that exact same resource anyways!).
(But how we'd communicate that particular notion to users in our documentation is a totally different question...)

Vinnl: The one major concern I have with not returning the saved dataset from the other `save*` functions, is that after saving it, people will probably continue working with the input dataset, which now has a mismatched ChangeLog (and thus _is_ different from the dataset that we return), rather than doing another GET to get the actual saved dataset. Additionally, the saved location is really not that useful for the other `save*` functions, since it's the exact same value as the input parameter.

Personally, I'm leaning towards either keeping things as they are in this PR right now (i.e. potentially returning `null` for the `InContainer` functions, where not having Read access is a relatively common scenario), _or_ doing a `GET` at the end of all our `save` functions (and thus potentially returning `null` everywhere), because saving is a relatively uncommon operation (at least using plain HTTP requests), so the overhead of an extra request is not that big, and then the developer would be sure they'd have the latest data after a save.

NSeydoux: Having the `save*inContainer` functions return the location sounds reasonable. For consistency, it would also make sense that the `save*At` return the location, even if in this case that isn't useful information, since the location is specified in the input. If we don't want to return the dataset with the resolved IRIs, I think returning `void` also makes sense.

> If we want more convenience functions (i.e. that do attempt to do 'multiple things in one call' (i.e. save a resource, and then retrieve it again from the server, with or without meta-data) then we can add new functions for those things separately (as they are a higher level of abstraction, with far more complex failure possibilities).

One of the convenience functions that would be useful I think is to make it possible to do what is currently implemented on save, some kind of `saveThenGet` function. In the case of datasets, where the local IRIs are resolved when saving, I think it would still be really convenient.

pmcb55: Yeah, I want to discuss the `saveThenGet()` thing with Vinnie tomorrow (as that was my original 'split this into two functions' suggestion). I'm thinking now a separate `refreshDataset()` or `syncLocalDatasetFromServer()` function might be nicer (semantically) - i.e. it would be explicitly up to the developer to request a re-sync instead of trying to conflate 'save and then fetch' into a single function - and this function could be called repeatedly (e.g. polled) independently of any 'save' function.
(And later we might attempt to have a `keepLocalDatasetInSyncWithServer()` which would use notifications to try and keep a local dataset instance **_constantly_** in-sync with the actual data on the Pod (getting into AutoMerge/Yjs territory there probably, which would be cool!)).

Vinnl: Refresh and syncing do not apply: we work with immutable datasets, and do not assume control over the source of truth of the data.  This allows the client to manage caching and data refreshing and make sure that that is in sync with the view and the runtime environment.

Anyway, would you guys mind moving this discussion out of this PR? Since it affects API that is not specific to detecting Pod Ownership, and I've got other things to work on that depends on this.

Vinnl: OK, so recap from our call just now:

- We're going to modify all `save` functions to return the data they _sent_ to the Pod (which can be something different from what's currently _at_ the Pod), and clearly note in the documentation that developers who want the up-to-date data will have to do another GET.
- The React SDK will be updated to do the GET.
- I'll file a separate PR for that work.

NSeydoux (approved):

PR #471, opened 2020-09-24
Add savePolicyDatasetAt()

Vinnl:
# New feature description

This is the first step towards implementing the experimental Access Control Policies proposal. Since this is a low-level API for an in-development proposal, the new functionality is not yet added to the top-level export, and is not yet included in the API documentation and changelog.

This adds `savePolicyDatasetAt()` and `createPolicyDataset()`.

# Checklist

- [x] All acceptance criteria are met.
- [x] Relevant documentation, if any, has been written/updated.
- [x] The changelog has been updated, if applicable.
- [x] New functions/types have been exported in `index.ts`, if applicable.
- [x] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).

PR #479, opened 2020-09-28
Add Access Policy initialisers

Vinnl:
# New feature description

Note: this builds on #471, so best to merge that before reviewing this.

It adds `createPolicy`, `getPolicy`, `getPolicyAll`, `setPolicy` and `removePolicy`.

# Checklist

### DIFF
--- a/src/acp/policy.test.ts
+++ b/src/acp/policy.test.ts
@@ -101,6 +101,16 @@ describe("savePolicyDatasetAt", () => {
 
     expect(mockedFetcher.fetch.mock.calls[0][0]).toBe(policyUrl);
   });
+
+  it("uses the given fetcher if provided", async () => {
+    const mockFetch = jest.fn(window.fetch).mockResolvedValue(new Response());
+
+    await savePolicyDatasetAt(policyUrl, createSolidDataset(), {
+      fetch: mockFetch,
+    });
+
+    expect(mockFetch.mock.calls[0][0]).toBe(policyUrl);
+  });
 });
 
 describe("createPolicy", () => {
--- a/src/acp/policy.test.ts
+++ b/src/acp/policy.test.ts
@@ -34,13 +34,22 @@ import { Response } from "cross-fetch";
 import { rdf, acp } from "../constants";
 import { createSolidDataset } from "../resource/solidDataset";
 import { getUrl, getUrlAll } from "../thing/get";
+import { removeUrl } from "../thing/remove";
 import { setUrl } from "../thing/set";
-import { asUrl, createThing, getThing, setThing } from "../thing/thing";
+import {
+  asUrl,
+  createThing,
+  getThing,
+  getThingAll,
+  setThing,
+} from "../thing/thing";
 import {
   createPolicy,
   getPolicy,
   getPolicyAll,
+  removePolicy,
   savePolicyDatasetAt,
+  setPolicy,
 } from "./policy";
 
 const policyUrl = "https://some.pod/policy-resource";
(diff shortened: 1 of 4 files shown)


src/acp/policy.test.ts
```diff
@@ -0,0 +1,254 @@
...
+        ReturnType<typeof window.fetch>,
+        [RequestInfo, RequestInit?]
+      >;
+    };
+
+    await savePolicyDatasetAt(policyUrl, createSolidDataset());
+
+    expect(mockedFetcher.fetch.mock.calls[0][0]).toBe(policyUrl);
+  });
+});
```
NSeydoux, line 104: There is no test checking explicitly that the provided fetch is used.

src/acp/policy.test.ts
```diff
@@ -0,0 +1,254 @@
...
+
+describe("createPolicy", () => {
+  it("sets creates a Thing of type acp:AccessPolicy", () => {
```
NSeydoux, line 107:
```suggestion
  it("creates a Thing of type acp:AccessPolicy", () => {
```

src/acp/policy.ts
```diff
@@ -0,0 +1,179 @@
...
+ * @param dataset The SolidDataset containing Access Policies to save.
+ * @param options Optional parameter `options.fetch`: An alternative `fetch` function to make the HTTP request, compatible with the browser-native [fetch API](https://developer.mozilla.org/docs/Web/API/WindowOrWorkerGlobalScope/fetch#parameters).
+ */
+export async function savePolicyDatasetAt(
+  url: Url | UrlString,
+  dataset: PolicyDataset,
+  options: Partial<
+    typeof internal_defaultFetchOptions
+  > = internal_defaultFetchOptions
+): ReturnType<typeof saveSolidDatasetAt> {
```
NSeydoux, line 76: Neat

src/acp/policy.ts
```diff
@@ -0,0 +1,179 @@
...
+ * ```
+ *
+ * Initialise a new, empty [[Policy]].
+ *
+ * @param url URL that identifies this Access Policy.
+ */
+export function createPolicy(url: Url | UrlString): Policy {
+  url = internal_toIriString(url);
+  let policyThing = createThing({ url: url });
+  policyThing = setUrl(policyThing, rdf.type, acp.AccessPolicy);
```
NSeydoux, line 97: It seems that we begin to use this construct quite often. Would it be worth to publicly expose a function `createThingTyped`, which takes the type as a mandatory parameter ? On one hand, we want common operations to be as affordable as possible, but on the other once we go down this path we could have a proliferation of functions that just bake in RDF predicates.

src/acp/policy.ts
```diff
@@ -0,0 +1,179 @@
...
+ *
+ * Get all [[Policy]]'s in a given [[PolicyDataset]].
+ *
+ * @param policyResource The Resource that contains Access Policies.
+ */
+export function getPolicyAll(policyResource: PolicyDataset): Policy[] {
+  const foundThings = getThingAll(policyResource);
+  const foundPolicies = foundThings.filter(
+    (thing) =>
+      !isThingLocal(thing) && getUrl(thing, rdf.type) === acp.AccessPolicy
```
NSeydoux, line 141: There is a possibility that an ACP might have several types, and that `getUrl` does not return the one we expect here. I'd suggest using `getUrlAll(thing, rdf.type).includes(acp.AccessPolicy)`.

NSeydoux (approved):

src/acp/policy.ts
Vinnl: Yeah, I was thinking about abstracting out this specific part, but since most of the "code" here is the comments, and how I'd sooner expect this specific function to change separately rather than wanting to update it for everything, I didn't do it for this PR.

That said, I didn't imagine it as a convenience function for developers, which could make sense. I've mentioned it in the ticket about `findThingBy*`, where it might make sense to combine it with a bit of other potential work.

src/acp/policy.ts
Vinnl: Good catch: 2ba0d5b26cbd94bc2a6e5036baf9fb0234e6f015

src/acp/policy.test.ts
Vinnl: Yeah fair enough: 097f37f351974e50b00bf884e3d12e372f5add3d

src/acp/policy.ts
pmcb55: Yeah I agree with the convenience function, although I'd call it `createThingOfType(type: iri)`. I wouldn't worry (yet!) about a proliferation of functions that just bake in RDF predicates - `rdf:type` is one of the most common RDF predicates out there, so having convenience functions for it makes a lot of sense (as does a 'getThingOfType()`).

src/acp/policy.ts
pmcb55: Yep - as with my previous comment, I'd actually just include an RDF-type-aware convenience function - something like 'isThingOfType(thing, type)` or 'hasThingType(thing, type)`.

pmcb55 (approved):
LGTM - couple of comments, but both for 'later'.

src/acp/policy.ts
```diff
@@ -0,0 +1,180 @@
...
+ * ```{note} There is no Access Control Policies specification yet. As such, this
+ * function is still experimental and subject to change, even in a non-major release.
+ * ```
+ *
+ * Initialise a new, empty [[Policy]].
+ *
+ * @param url URL that identifies this Access Policy.
+ */
+export function createPolicy(url: Url | UrlString): Policy {
+  url = internal_toIriString(url);
```
pmcb55, line 95: Is this common JS practice (i.e. reassigning an input parameter)? In Java this would be frowned upon (i.e. as a general rule I always use `final` (Java's equivalent of `const`) for all input params).

Vinnl: I usually avoid it too, didn't do it here since it's essentially still the same thing and it was on the first line, but I'm definitely up for just banning it altogether (possibly one for https://github.com/inrupt/javascript-style-configs). Fixed: 5ac4efb1971b7eb4bd5142fc2c3e63dddc8ef90a

src/acp/policy.ts
```diff
@@ -32,18 +32,26 @@ import {
...
-export type AccessPolicy = Thing;
+export type Policy = ThingPersisted;
 
 /**
  * ```{note} There is no Access Control Policies specification yet. As such, this
  * function is still experimental and subject to change, even in a non-major release.
  * ```
  *
- * Initialise a new empty [[SolidDataset]] to story [[AccessPolicy]]'s in.
+ * Initialise a new empty [[SolidDataset]] to story [[Policy]]'s in.
```
pmcb55, line 54: Typo - 'story' should be 'store'. (Don't really need the word 'in' at the end of the sentence either.)

PR #486, opened 2020-10-01
Implement hasAcr()

Vinnl:
# New feature description

It adds ~~`hasAcr`~~ `hasLinkedAcr`, which checks whether the given Resource exposes Access
Controls via the experimental Access Policy mechanism.

Note that I chose not to name it `hasAccessibleAcr`, as in the acceptance criteria, because the APC proposal indicated that the presence of the `acp:accessControl` header does not guarantee that that Resource is actually accessible.

# Checklist

- [ ] All acceptance criteria are met. (See above.)

### DIFF
--- a/src/acp/control.test.ts
+++ b/src/acp/control.test.ts
@@ -1,0 +1,69 @@
+/**
+ * Copyright 2020 Inrupt Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to use,
+ * copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the
+ * Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,
+ * INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A
+ * PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+ * HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+ * OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+ * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+import { describe, it, expect } from "@jest/globals";
+
+import { hasLinkedAcr, WithLinkedAcpAccessControl } from "./control";
+import { acp } from "../constants";
+import { WithAccessibleAcl, WithResourceInfo } from "../interfaces";
+
+describe("hasLinkedAcr", () => {
+  it("returns true if a Resource exposes a URL to an Access Control Resource", () => {
+    const withLinkedAcr: WithLinkedAcpAccessControl = {
+      internal_resourceInfo: {
+        isRawData: false,
+        sourceIri: "https://some.pod/resource",
+        linkedResources: {
+          [acp.accessControl]: ["https://some.pod/access-control-resource"],
+        },
+      },
+    };
+
+    expect(hasLinkedAcr(withLinkedAcr)).toBe(true);
+  });
+
+  it("returns false if a Resource is governed by Web-Access-Control", () => {
+    const withLinkedAcr: WithAccessibleAcl = {
+      internal_resourceInfo: {
+        isRawData: false,
+        sourceIri: "https://some.pod/resource",
+        linkedResources: {
+          acl: ["https://some.pod/access-control-resource"],
+        },
+        aclUrl: "https://some.pod/access-control-resource",
+      },
+    };
+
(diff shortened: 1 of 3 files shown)


src/acp/control.ts
```diff
@@ -0,0 +1,66 @@
...
+export function hasAcr<Resource extends WithResourceInfo>(
+  resource: Resource
+): resource is WithAcpAccessControl<Resource> {
+  return (
+    hasResourceInfo(resource) &&
+    Array.isArray(
+      resource.internal_resourceInfo.linkedResources[acp.accessControl]
+    ) &&
+    resource.internal_resourceInfo.linkedResources[acp.accessControl].length ===
+      1
```
pmcb55, line 44: Are you missing a test for this condition - i.e. what if the array length is > 1 (is that allowed, what does it mean (explained in the function's TSDoc, I mean)?)...?

Vinnl: Array length > 1 is not allowed by the ACP spec, so we return `false` in that case. I've added a clarification in the TSDoc: 27bcaef29f54703db673ad5edcde2c1acacd937b

Vinnl:
Update: I renamed `hasAcr` to `hasLinkedAcfr`, because we already have a function `hasAcl` that checks whether a Resource was fetched together with its ACL Resources - whereas this just checks whether it _refers_ to such Resources for ACP. (We'll probably add `hasAcp` later to check whether a Resource was fetched together with its Access Control Resources and Access Policy Resources.)

src/acp/control.ts
pmcb55: Cool. So it raises the question of should our client enforce that spec condition - i.e. should we throw an exception (not here perhaps (although maybe here too), but maybe in the code that retrieves the linked resources information), e.g. if a 'broken' server returned multiple?
It just seems like it could be confusing to a client (if they aren't familiar with the nitty-gritty details of the spec) to be told a resource has no ACR when a response (from a buggy/broken/malicious server) actually has 2 or more ACR links.

Vinnl: I'm not sure whether it should throw an exception. As a comparison, if the server would expose the ACR with the wrong link relation, we wouldn't throw an error either - we'd just say that the server does not expose an ACR. In other words, we don't want to anticipate all the billion ways in which a server could wrongly expose an ACR, but just support the one way in which it could correctly do so.

So yes, I think it's totally reasonable for a client to be confused when the server does a confusing thing, i.e. exposing to ACRs when the spec explicitly says there should be just one. Just like they would be confused if it used e.g. the wrong `rel`.

pmcb55: Well, in this case I don't agree. The example of a wrong `rel` is a very different condition I think (i.e. our client would clearly return `false` in that case - as the server has explicitly **_not_** said it has an ACR).
But the _generally_ case of our client verifying server spec compliance, when that's easy and simple for us to do so (as it is in this particular case), I think just makes our library more useful and helpful (and educational, in this case) to developers.
@NSeydoux - what are your thoughts on this (it's a more general question really (i.e. should our library check for spec-compliance when it's easy and simple to do so, and to throw if it detects a clear violation))??

Vinnl: So what are developers to do in this case? Some time, way after implementing, some user is using a non-compliant server, so we/the developer cannot rely on that server properly implementing ACP. Other than just not supporting modifying policies (i.e. what we do now), is the developer going to display an error to the user saying that their server is for some reason exposing multiple links to what it says are ACRs? That doesn't really seem actionable for the user, especially considering how this is really mostly a theoretical situation.

(Note that throwing an error will cause the developer's application to crash unless they explicitly take that into account, so we'd be adding additional failure modes and effort for the developer, whereas what they actually want to do is just not display anything ACP-related in case of a misbehaving server.)

In any case, even if we do decide to throw an error, that would surely be when fetching the Resource, which is not what this PR adds.

pmcb55: Yeah, as I said above too - the check should be when fetching the resource really. And yeah, I agree that throwing an error is probably too much of an imposition on the developer/user. But I think it might be nice to at least log it (or ultimately to have our library record the problem 'somewhere' (like sending a bug report)), but we don't have mechanisms for either of those things right now.
So I guess this boils down to being an example of a super-simple opportunity to detect a spec-violation - and it seems a 'waste' to not take advantage of that somehow (e.g. notify the developer, or the server implementer, etc.) - but yeah, we can leave it as-is for this PR.

NSeydoux: I agree, throwing an error does require the app consuming our library to manage it, when if the server is just non-spec compliant, there is probably nothing they can do. Giving more information to the developer feels like having the spec bleed into the API, which we don't necessarily want, as the purpose of our API is to abstract the spec.

I agree that detecting a spec violation is an opportunity, but since this can only happen at runtime, I don't know to what extent in can be transformed into an action... It almost feels like at this point, non-compliance management would be a whole set of features that we explicitly expose in a dedicated API, rather than bundling them in "practical" features. If one hits an error, testing for compliance might help give feedback to the user, but that sounds like quite an advanced use case, when any app will need a simple approach to access data if it is available in a spec compliant way, and to act as if there was no data otherwise, in order not to be tempted to make any guesses and in fine enable non-spec compliant behaviours.

NSeydoux (approved):

src/acp/control.test.ts
```diff
@@ -0,0 +1,69 @@
...
+ */
+
+import { describe, it, expect } from "@jest/globals";
+
+import { hasLinkedAcr, WithLinkedAcpAccessControl } from "./control";
+import { acp } from "../constants";
+import { WithAccessibleAcl, WithResourceInfo } from "../interfaces";
+
+describe("hasLinkedAcr", () => {
+  it("returns true if a Resource exposes a URL to an Access Control Resource", () => {
```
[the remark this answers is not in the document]
Vinnl: That's indeed not possible. In fact, if a server supports both, it will still only support one at a time _for the entire Pod_.

PR #489, opened 2020-10-05
Ensure no input values are modified anywhere

### LINKED ISSUE #224: The async litDataset methods manipulate the input LitDataset
**Search terms you've used**

immutability, LitDataset, Resource, ACL

**Describe the bug**

The async methods in `litDataset.ts` (or, if #161 gets implemented first, also the Resource-specific async methods) use `Object.assign()` to add metadata returned from the call to the Pod to the input Resource. However, `Object.assign()` also _changes_ its first argument.

Normally, you'd solve that by passing an empty object as the first argument, and shift the other arguments one position. However, because `LitDataset`s actually are a class, that's not sufficient here, IIRC because the prototype chain doesn't get copied over.

As an additional complexity, some of these methods (specifically, `saveAclFor()` and `deleteAclFor()`) only operate on the metadata, and should hence work on both LitDatasets and, after #219, regular files.

Presumably, the solution here is to check whether a Resource is a file, in which case the file should be cloned in whatever way is appropriate, if it's a LitDataset, in which case it should use `clone()` from `rdfjs.ts`, or just metadata, in which case we can just initialise an empty object. Then, `cloneLitStructs` (which should probably be renamed as well, since we no longer use the term `Struct`) should be modified to operate on the resulting clone to copy over the relevant properties from the input Resource.

**To Reproduce**
Call `saveLitDatasetAt()` with a changed `LitDataset`. Afterwards, check `.changeLog` on the input `LitDataset`. Instead of still listing the changes that were applied to it, it is now empty.

**Expected behavior**

It should still list the changes applied before saving it.

Vinnl:
This fixes #224, a long-standing risk where we manipulated input values (`Object.assign` mutates its first argument). Usually, the guard against this is to just pass an empty object to `Object.assign`, but since Datasets and Blobs are not always (depending on the implementation, I suppose) plain JS objects, we couldn't do that - we need an actual instance of their classes with copied data. This adds a function `internal_cloneResource` that uses a heuristic to copy over the properties we've added to such an instance (by only copying over properties with atomic values, i.e. not functions like `Dataset.match`), which is slightly better than what we did before (copy over a list of known properties, e.g. `internal_withResourceInfo`) and thus can be used everywhere.

### DIFF
--- a/src/acl/acl.test.ts
+++ b/src/acl/acl.test.ts
@@ -51,6 +51,7 @@ import {
   deleteAclFor,
   createAcl,
   internal_getContainerPath,
+  hasAcl,
 } from "./acl";
 import {
   WithResourceInfo,
@@ -59,6 +60,7 @@ import {
   AclDataset,
   Access,
   WithAccessibleAcl,
+  WithAcl,
 } from "../interfaces";
 
 function mockResponse(
@@ -403,6 +405,25 @@ describe("getContainerPath", () => {
 
   it("does not prefix a slash if the input did not do so either", () => {
     expect(internal_getContainerPath("container/resource")).toBe("container/");
+  });
+});
+
+describe("hasAcl", () => {
+  it("returns true if a Resource was fetched with its ACL Resources attached", () => {
+    const withAcl: WithAcl = {
+      internal_acl: {
+        resourceAcl: null,
+        fallbackAcl: null,
+      },
+    };
+
+    expect(hasAcl(withAcl)).toBe(true);
+  });
+
+  it("returns false if a Resource was fetched without its ACL Resources attached", () => {
+    const withoutAcl = {};
+
+    expect(hasAcl(withoutAcl)).toBe(false);
   });
 });
 
--- a/src/acl/acl.ts
+++ b/src/acl/acl.ts
@@ -50,6 +50,7 @@ import {
   getSourceUrl,
   internal_defaultFetchOptions,
   getResourceInfo,
+  internal_cloneResource,
 } from "../resource/resource";
 import { addIri } from "..";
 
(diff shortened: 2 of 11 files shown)


pmcb55:
Hi @Vinnl - this is probably my lack of TS / JS experience again, but I'm struggling to fully understand the issue here (is it as simple as our SDK wanting to have pure immutable JS objects, whereas the underlying RDF/JS libraries use JS classes with mutable state?).
The heuristics in the `cloneResource()` also make me nervous (i.e. it 'seems' like there might be a need for something like the `termType` field in the basic RDF/JS classes, like `Literal`, `BlankNode`, `NamedNode`, etc. - i.e. an explicit field providing the exact object type as a string, rather than relying on the presence of specifically named methods).
Anyway, I guess you consider this important to do before GA, so I assume it's really important that I properly understand the actual problem better here...

Vinnl:
> (is it as simple as our SDK wanting to have pure immutable JS objects

yes

> whereas the underlying RDF/JS libraries use JS classes with mutable state?

It's not so much that they're using classes with mutable state, but that there are no native JS facilities to clone a class instance, and that the RDF/JS doesn't even mandate that Datasets actually _are_ class instances, I think.

> The heuristics in the cloneResource() also make me nervous (i.e. it 'seems' like there might be a need for something like the termType field in the basic RDF/JS classes, like Literal, BlankNode, NamedNode, etc. - i.e. an explicit field providing the exact object type as a string, rather than relying on the presence of specifically named methods).

That would be nice, but the RDF/JS spec doesn't specify that. That said, I'm not worried too much about the heuristics for distinguishing Files and SolidDatasets (since we'd expect to get one of those anyway, so they're reliable). The heuristics for identifying our own properties rather than instance functions are somewhat more iffy, but still better than what we had.

Longer term, ideally we'd just use raw JS data structures (would be more efficient too) and provide a quick translation functions that exports it into an RDF/JS dataset proper (we discussed something like that earlier with Emmet), but that's more of a 'cleanup' thing. For now, this PR at least prevents actual bugs (i.e. in which the developer is trying to access stale data but unexpectedly gets data that is out of sync with their View).

pmcb55 (approved):
LGTM - (thanks for the explanation on my comments - I'd really love to get to that 'clean-up' phase quickly (i.e. moving everything into the RDF/JS dataset itself)).

PR #493, opened 2020-10-06
Prevent Jest from injecting globals

Vinnl:
Since 9cc1379d07d627cb2baa58f6c7a19c8de3df1e3b, Jest supports
disabling the injection of globals. This makes tests easier to
follow, and ensures that all types can be properly checked as well.

(~~Unfortunately this also exposed a flaw in Jest's type definitions,
which I've fixed here but which isn't released yet:
https://github.com/facebook/jest/pull/10600~~ Fix has been released, workarounds removed: f0e99c9f135d77d78869606de6286c5e4804fbfc)

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,3 +1,24 @@
+/**
+ * Copyright 2020 Inrupt Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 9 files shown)


pmcb55 (approved):
LGTM

PR #495, opened 2020-10-07
Manage Policies for Access Controls

Vinnl:
# New feature description

It adds functions to add, get and remove policy URLs from Access Controls. Note that I appended `-Url` to the function names, which was not specified in the acceptance criteria. The reason I did that is because you cannot pass in-memory policies, which would be problematic when their URLs would be resolved on saving as relative to the Access Control Resource, even though they will often be located in an Access Policy Resource. Thus, you should already know the URL of the policy before adding it.

I also renamed the addition functions `add*` instead of `apply*` - initially, I called them apply because that's what the Predicates are named, but that refers to applying the Policy to the Resource that the Control is about, not to the Control itself. Thus, `add` seemed more reasonable for adding it to a Control.

# Checklist

- [ ] All acceptance criteria are met. _No, see above._


src/acp/control.test.ts
```diff
@@ -228,3 +245,745 @@ describe("removeAccessControl", () => {
...
+  it("accepts policy URLs as Things with URLs", () => {
+    const policyUrl = "https://some.pod/policy.ttl#some-policy";
+    const policy = createThing({ url: policyUrl });
+    const control = createThing();
+
+    const updatedControl = addPolicyUrl(control, policy);
+
+    expect(getIriAll(updatedControl, acp.apply)).toContain(policyUrl);
+  });
+});
```
NSeydoux, line 290: There is no tests checking that the input object is left unmodified. I don't think it's a big deal, but this type of test is present in other test suites. Should it be added here ?

This comment applies to other tests in this file.

src/acp/control.test.ts
```diff
@@ -228,3 +245,745 @@ describe("removeAccessControl", () => {
...
+    const updatedControl = removeConstantMemberPolicyUrl(control, policy);
+
+    const foundPolicyUrl = getIri(updatedControl, acp.applyMembersConstant);
+
+    expect(foundPolicyUrl).toBeNull();
+  });
+});
+
+describe("removeConstantMemberPolicyUrlAll", () => {
+  it("does not remove other existing Constant Member Policies", () => {
```
NSeydoux, line 963:
```suggestion
  it("removes all Constant Member Policies", () => {
```

NSeydoux (approved):

src/acp/control.test.ts
Vinnl: How did that happen? And how on earth did you manage to spot that? :P 08e89b157e7a9fa607c2c18abaf7aa03a5ee2fd1

src/acp/control.test.ts
Vinnl: That is cleaner:
- d03f3b5c1e5217a4785b08e8c35b8b05821f26e4
- d5357c62596ad1ca22ebde6f7aa8469a101eb53e

Vinnl:
Update: just received word that "Constant" policies are not being implemented atm, and that the name is still in flux too. Hence, I've removed it from this PR: 3cd493daf9666f60444c17f25e637a971d7f7eda.

PR #499, opened 2020-10-08
Add missing overload for asUrl

Vinnl:
This PR fixes a bug in which a call to `asUrl` with a `Thing` (rather than an object which is known to explicitly be a `ThingLocal` or a `ThingPersisted`) would be flagged by TypeScript because the function overload for the generic case was missing.

- [x] I've added a unit test to test for potential regressions of this bug. N/A


pmcb55 (approved):
LGTM

PR #508, opened 2020-10-14
Distinguish between sent and received resourceInfo

Vinnl:
# New feature description

This adds a type WithServerResourceInfo which is like the existing
WithResourceInfo, but adds properties that the server has to tell
us. Specifically, it is data received when sending a request to
that specific Resource (like the URLs of linked Resources), but not
when creating the Resource by sending a POST to its Container.

Since all required properties that were available on
WithResourceInfo in the last release are still available now, this
allows us to make the previous breaking change no longer breaking.
Instead, we explicitly clarify that the return value of our
`save*()` functions are the data we _sent_ to the Pod, which is not
necessarily all data that's actually available at the Pod (e.g.
when someone changed it in between our fetching and saving it).

When functions _can_ obtain data that's set by the server (i.e.
when doing a request directly to the Resource), they now return
WithServerResourceInfo, which is an extension of the existing
WithResourceInfo that (currently) adds a list of Linked Resources.

# Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,18 +6,12 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 The following changes have been implemented but not released yet:
 
-### Breaking changes
-
-- `saveSolidDatasetInContainer`, `saveFileInContainer` and `createContainerInContainer` now return
-  `null` if the newly-created Resource is not actually readable by the current user. Please update
-  your code with a `null` check on the return value before writing to the Resource again, or read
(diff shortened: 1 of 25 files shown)


Vinnl:
Hmm, apparently I missed some tests when running Jest in watch mode - suppose they were cached or something. I'll fix them up.

Vinnl:
OK there we go, sorry about that.

NSeydoux (approved):

NSeydoux:
The distinction between `Resourceinfo` and `ServerResourceinfo` is a really good idea, I think it kind of documents itself, and it really makes the concept impossible to miss, so good thinking

PR #515, opened 2020-10-15
Add get*WithAccessDatasets()

Vinnl:
# New feature description

It add functions to fetch a Resource together with either its ACL or its ACR, depending on which applies to it - or neither, if neither is exposed, of course.

# Checklist

### DIFF
--- a/src/acp/acp.test.ts
+++ b/src/acp/acp.test.ts
@@ -33,8 +33,11 @@ jest.mock("../fetcher.ts", () => ({
 
 import { Response } from "cross-fetch";
 import {
+  getFileWithAccessDatasets,
   getFileWithAcp,
+  getResourceInfoWithAccessDatasets,
   getResourceInfoWithAcp,
+  getSolidDatasetWithAccessDatasets,
   getSolidDatasetWithAcp,
   WithAccessibleAcr,
 } from "./acp";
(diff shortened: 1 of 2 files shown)


src/acp/acp.test.ts
```diff
@@ -746,3 +749,452 @@ describe("getResourceInfoWithAcp", () => {
...
+describe("getSolidDatasetWithAccessDatasets", () => {
+  it("fetches the Resource at the given URL", async () => {
+    const mockedGetSolidDataset = jest.spyOn(
+      SolidDatasetModule,
+      "getSolidDataset"
+    );
+
+    await getSolidDatasetWithAccessDatasets("https://some.pod/resource");
+
+    expect(mockedGetSolidDataset).toHaveBeenCalledTimes(1);
```
pmcb55, line 762: These tests seem to be testing just internal implementation details (i.e. that some **_other_** function was called X times, and with value Y), as opposed to asserting on behaviour. It seems particularly strange when the function under test is a `getXXX()` function and we're not assigning the return value to anything to assert on it's value - i.e. to assert on the expected behaviour of the function itself.
Maybe I'm missing something...?

Vinnl: Yeah I was being a bit lazy here. Seeing as this is a really thin wrapper that delegates to the existing ACL- or ACP-related functions, I didn't feel like copy-pasting (with small adjustments) all those tests again for this wrapper. Especially not considering that that would have to be done three times (for SolidDatasets, Files and ResourceInfo).

My judgement call was that this would be provide enough safety in this case, and that the assurance-effort ratio made this the better choice.

Perhaps I should add the above as a comment to the tests? Alternatively, feel free to copy over the existing tests before I get back on Monday ;)

pmcb55: Hhhmmm... I think perhaps the bigger issue here might be an over avoidance of DRY (for the sake of fully self-contained tests). In my experience, simple wrapper functions should always be simple to test (properly), as I've always favored test code re-use for large, heavily used libraries (e.g. shared mocks, shared fixture setup code, shared matchers, etc.). Just by way of example (from Java-land), but Apache Camel has always had very extensive test support, which they use to test themselves, but then expose for users of their library to also reuse if they want (https://livebook.manning.com/book/camel-in-action/chapter-6/ch06lev1sec1). 
As a basic, simple example in our own case, I see our simple use of a shared Test Vocab (https://github.com/inrupt/solid-common-vocab-rdf/blob/master/inrupt-rdf/Common/CopyOfVocab/inrupt-test.ttl) as being a simple part of shareable test support we should be:
1. using ourselves in our own tests (i.e. for readability and consistency), and...
2. ...offering to others to reuse if they want.
I know this is all a completely separate topic, but I'll mark this discussion as a case-in-point input for that discussion later.

pmcb55 (approved):
Raises a few topics for more general discussion, but not for now. Otherwise this looks good.

NSeydoux (approved):

PR #520, opened 2020-10-17
Expose functions to link rules to policies

NSeydoux:
This is part of the implementation of the new ACP-based access control. Specifically, this adds a couple of functions that make it possible to create/delete relations between policies and rules, based on the three dedicated predicates in the ACP vocabulary: `acp:allOf`, `acp:anyOf`, and `acp:noneOf`. This way, a developer may specify which are the rules that are required/optional/forbidden in order to get some access right to a resource.

Note that the ACP spec still being experimental, these functions are not exported as part of the public API, nor are they mentionned in the changelog or in the documentation.

# Checklist

- [ ] Relevant documentation, if any, has been written/updated. N/A
- [ ] The changelog has been updated, if applicable. N/A
- [ ] New functions/types have been exported in `index.ts`, if applicable. N/A

### DIFF
--- a/src/acp/rule.test.ts
+++ b/src/acp/rule.test.ts
@@ -1,0 +1,704 @@
+/**
+ * Copyright 2020 Inrupt Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to use,
+ * copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the
+ * Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,
+ * INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A
+ * PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+ * HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+ * OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+ * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+import { jest, describe, it, expect } from "@jest/globals";
+import { asIri, createThing } from "../thing/thing";
+import {
+  addForbiddenRuleToPolicy,
+  addOptionalRuleToPolicy,
+  addRequiredRuleToPolicy,
+  getForbiddenRuleOnPolicyAll,
+  getOptionalRuleOnPolicyAll,
+  getRequiredRuleOnPolicyAll,
+  Rule,
+  setForbiddenRuleOnPolicy,
+  setOptionalRuleOnPolicy,
+  setRequiredRuleOnPolicy,
+} from "./rule";
+import { DataFactory } from "n3";
+import { Thing, ThingPersisted } from "../interfaces";
+import { Policy } from "./policy";
+
+const ACP_ANY = "http://www.w3.org/ns/solid/acp#anyOf";
+const ACP_ALL = "http://www.w3.org/ns/solid/acp#allOf";
+const ACP_NONE = "http://www.w3.org/ns/solid/acp#noneOf";
+
+const mockRule = (url: string) =>
+  createThing({
+    url,
+  });
+
+const addAll = (
+  thing: ThingPersisted,
+  predicate: string,
+  objects: ThingPersisted[]
+): void => {
+  objects.forEach((objectToAdd: ThingPersisted) => {
+    thing.add(
+      DataFactory.quad(
+        DataFactory.namedNode(asIri(thing)),
+        DataFactory.namedNode(predicate),
+        DataFactory.namedNode(asIri(objectToAdd))
+      )
+    );
+  });
+};
+
+const mockPolicy = (
+  url: string,
+  rules?: { required?: Rule[]; optional?: Rule[]; forbidden?: Rule[] }
+): Policy => {
+  const mockPolicy = createThing({ url });
+  if (rules?.forbidden) {
+    addAll(mockPolicy, ACP_NONE, rules.forbidden);
+  }
+  if (rules?.optional) {
+    addAll(mockPolicy, ACP_ANY, rules.optional);
+  }
+  if (rules?.required) {
+    addAll(mockPolicy, ACP_ALL, rules.required);
+  }
+  return mockPolicy;
+};
+
+describe("addForbiddenRuleToPolicy", () => {
+  it("adds the rule in the forbidden rules of the policy", () => {
+    const myPolicy = addForbiddenRuleToPolicy(
+      mockPolicy("https://some.pod/policy-resource#policy"),
+      mockRule("https://some.pod/rule-resource#rule")
+    );
+    expect(
+      myPolicy.has(
+        DataFactory.quad(
+          DataFactory.namedNode("https://some.pod/policy-resource#policy"),
+          DataFactory.namedNode(ACP_NONE),
+          DataFactory.namedNode("https://some.pod/rule-resource#rule")
+        )
+      )
+    ).toBe(true);
+  });
+
+  it("does not remove the existing forbidden rules", () => {
+    const mockedPolicy = mockPolicy("https://some.pod/policy-resource#policy", {
+      forbidden: [mockRule("https://some.pod/rule-resource#another-rule")],
+    });
+    const myPolicy = addForbiddenRuleToPolicy(
+      mockedPolicy,
+      mockRule("https://some.pod/rule-resource#a-rule")
+    );
+    expect(
+      myPolicy.has(
+        DataFactory.quad(
+          DataFactory.namedNode("https://some.pod/policy-resource#policy"),
+          DataFactory.namedNode(ACP_NONE),
+          DataFactory.namedNode("https://some.pod/rule-resource#another-rule")
+        )
+      )
+    ).toBe(true);
+  });
+
+  it("does not change the existing required and optional rules", () => {
+    const mockedPolicy = mockPolicy("https://some.pod/policy-resource#policy", {
+      optional: [mockRule("https://some.pod/rule-resource#optional-rule")],
+      required: [mockRule("https://some.pod/rule-resource#required-rule")],
+    });
+    const myPolicy = addForbiddenRuleToPolicy(
+      mockedPolicy,
+      mockRule("https://some.pod/rule-resource#forbidden-rule")
+    );
+    expect(
+      myPolicy.has(
+        DataFactory.quad(
+          DataFactory.namedNode("https://some.pod/policy-resource#policy"),
+          DataFactory.namedNode(ACP_ALL),
+          DataFactory.namedNode("https://some.pod/rule-resource#required-rule")
+        )
+      )
+    ).toBe(true);
+    expect(
+      myPolicy.has(
+        DataFactory.quad(
+          DataFactory.namedNode("https://some.pod/policy-resource#policy"),
+          DataFactory.namedNode(ACP_ANY),
+          DataFactory.namedNode("https://some.pod/rule-resource#optional-rule")
+        )
+      )
+    ).toBe(true);
+  });
+
+  it("does not change the input policy", () => {
+    const myPolicy = mockPolicy("https://some.pod/policy-resource#policy");
+    addForbiddenRuleToPolicy(
+      myPolicy,
+      mockRule("https://some.pod/rule-resource#rule")
+    );
+    expect(myPolicy.size).toEqual(0);
+  });
+});
+
+describe("addOptionalRuleToPolicy", () => {
+  it("adds the rule in the optional rules of the policy", () => {
+    const myPolicy = addOptionalRuleToPolicy(
+      mockPolicy("https://some.pod/policy-resource#policy"),
+      mockRule("https://some.pod/rule-resource#rule")
+    );
+    expect(
+      myPolicy.has(
+        DataFactory.quad(
+          DataFactory.namedNode("https://some.pod/policy-resource#policy"),
+          DataFactory.namedNode(ACP_ANY),
+          DataFactory.namedNode("https://some.pod/rule-resource#rule")
+        )
+      )
+    ).toBe(true);
+  });
+
+  it("does not remove the existing optional rules", () => {
+    const mockedPolicy = mockPolicy("https://some.pod/policy-resource#policy", {
+      optional: [mockRule("https://some.pod/rule-resource#another-rule")],
+    });
+    const myPolicy = addOptionalRuleToPolicy(
+      mockedPolicy,
+      mockRule("https://some.pod/rule-resource#a-rule")
+    );
+    expect(
+      myPolicy.has(
+        DataFactory.quad(
+          DataFactory.namedNode("https://some.pod/policy-resource#policy"),
+          DataFactory.namedNode(ACP_ANY),
+          DataFactory.namedNode("https://some.pod/rule-resource#another-rule")
+        )
+      )
+    ).toBe(true);
+  });
+
+  it("does not change the existing required and forbidden rules", () => {
+    const mockedPolicy = mockPolicy("https://some.pod/policy-resource#policy", {
+      forbidden: [mockRule("https://some.pod/rule-resource#forbidden-rule")],
+      required: [mockRule("https://some.pod/rule-resource#required-rule")],
+    });
+    const myPolicy = addOptionalRuleToPolicy(
+      mockedPolicy,
+      mockRule("https://some.pod/rule-resource#optional-rule")
+    );
+    expect(
+      myPolicy.has(
+        DataFactory.quad(
+          DataFactory.namedNode("https://some.pod/policy-resource#policy"),
+          DataFactory.namedNode(ACP_ALL),
+          DataFactory.namedNode("https://some.pod/rule-resource#required-rule")
+        )
+      )
+    ).toBe(true);
+    expect(
+      myPolicy.has(
+        DataFactory.quad(
+          DataFactory.namedNode("https://some.pod/policy-resource#policy"),
+          DataFactory.namedNode(ACP_NONE),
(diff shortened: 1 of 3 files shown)


src/acp/rule.test.ts
```diff
@@ -0,0 +1,704 @@
...
+  Rule,
+  setForbiddenRuleOnPolicy,
+  setOptionalRuleOnPolicy,
+  setRequiredRuleOnPolicy,
+} from "./rule";
+import { DataFactory } from "n3";
+import { Thing, ThingPersisted } from "../interfaces";
+import { Policy } from "./policy";
+
+const ACP_ANY = "http://www.w3.org/ns/solid/acp#anyOf";
```
pmcb55, line 40: Why are these 3 local constants defined here as strings (and not NamedNodes)? Each of them is used 12 times in this file - 11 of those times they're wrapped in a NameNode call, and the 12th time they could be passed as a NamedNode or a string! I'm not trying to kick off the String vs IRI debate again here (that's a totally different discussion anyway) - it just seems a clear example of making this test code **_less_** readable and unnecessarily verbose for not benefit whatsoever.

src/acp/rule.test.ts
```diff
@@ -0,0 +1,704 @@
...
+const ACP_ALL = "http://www.w3.org/ns/solid/acp#allOf";
+const ACP_NONE = "http://www.w3.org/ns/solid/acp#noneOf";
+
+const mockRule = (url: string) =>
+  createThing({
+    url,
+  });
+
+const addAll = (
```
pmcb55, line 49: Could this helper function be more clearly named as `addRulesToThing()`  (and the 3rd param typed as a `Rule`) - as that seems to be the intent of the function...?

src/acp/rule.test.ts
```diff
@@ -0,0 +1,704 @@
...
+      DataFactory.quad(
+        DataFactory.namedNode(asIri(thing)),
+        DataFactory.namedNode(predicate),
+        DataFactory.namedNode(asIri(objectToAdd))
+      )
+    );
+  });
+};
+
+const mockPolicy = (
```
pmcb55, line 65: This doesn't seem to be an actual 'mock' at all - it's creating an actual Policy dataset. I'd just name it `createPolicy()` (and it seems like a handy little helper function that might be useful in the main source, and not just in tests...?).

src/acp/rule.test.ts
```diff
@@ -0,0 +1,704 @@
...
+} from "./rule";
+import { DataFactory } from "n3";
+import { Thing, ThingPersisted } from "../interfaces";
+import { Policy } from "./policy";
+
+const ACP_ANY = "http://www.w3.org/ns/solid/acp#anyOf";
+const ACP_ALL = "http://www.w3.org/ns/solid/acp#allOf";
+const ACP_NONE = "http://www.w3.org/ns/solid/acp#noneOf";
+
+const mockRule = (url: string) =>
```
pmcb55, line 44: I'm confused about why this would be considered a 'mock rule'. It just seems to be creating an empty thing, so would it not make the code clearer to just use `createThing(url)` in the rest of this code rather than attempting to abstract it into a 'Rule' (unless the idea is to later add to this function to actually do something Rule-specific, such as adding a triple to the Thing it's creating stating it's a Thing of type `acp:Rule` or something...)? Even in that case, I'd still just rename this function to `createRule()`.

src/acp/rule.test.ts
```diff
@@ -0,0 +1,704 @@
...
+    ).toBe(true);
+  });
+
+  it("does not change the input policy", () => {
+    const myPolicy = mockPolicy("https://some.pod/policy-resource#policy");
+    addForbiddenRuleToPolicy(
+      myPolicy,
+      mockRule("https://some.pod/rule-resource#rule")
+    );
+    expect(myPolicy.size).toEqual(0);
```
pmcb55, line 153: I found this test kinda confusing (i.e. was creating a policy, calling an `addXXX()` function, and then asserting on the size of something being zero - so it just **_feels_** a bit jarring). I assume it's just testing that the function-under-test doesn't mutate it's input, which I guess is fine (although it feels kinda strange - i.e. do we test every single function we write to ensure it doesn't mutate it's input?). But just to make the intent clearer, I'd rewrite it as:
```
it("does not change the input policy", () => {
    const myPolicy = mockPolicy("https://some.pod/policy-resource#policy");
    const myPolicySize = myPolicy.size;
    const newPolicy = addForbiddenRuleToPolicy(
      myPolicy,
      mockRule("https://some.pod/rule-resource#rule")
    );
    expect(myPolicy.size).toEqual(myPolicySize);
    expect(newPolicy.size).toEqual(myPolicySize + 1);
};
```
And maybe I'd rewrite the test description as:
```
it("adding rule to policy does not mutate the input policy", ...
```

pmcb55 (approved):
LGTM - just minor comments really. But I am curious about how you **_feel_** about what seems like an incredible proliferation of string constants in this test file (and the fact that it takes 700 lines of test code for probably less than 100 lines (if you ignore comments) of actual code!?

src/acp/rule.test.ts
NSeydoux: There is not particular reason, so they could be changed to be `NamedNodes`, but they are mostly used along other named nodes, such as: 
```
expect(
      myPolicy.has(
        DataFactory.quad(
          DataFactory.namedNode("https://some.pod/policy-resource#policy"),
          DataFactory.namedNode(ACP_NONE),
          DataFactory.namedNode("https://some.pod/rule-resource#rule")
        )
      )
    ).toBe(true);
```
so there would be little value in changing `DataFactory.namedNode(ACP_NONE)` into `ACP_NONE`. To make the statement more compact, we'd have to also declare constants for `DataFactory.namedNode("https://some.pod/policy-resource#policy")` and `DataFactory.namedNode("https://some.pod/rule-resource#rule")`, which I'm happy to do but I wasn't sure would make the tests more readable. i'll give it a try.

src/acp/rule.test.ts
Vinnl: It's creating an actual Policy with fake data in it - sounds like a mock to me?

(It also mutates its input, which we shouldn't be doing in the main source.)

src/acp/rule.test.ts
```diff
@@ -0,0 +1,704 @@
...
+});
+
+describe("setForbiddenRuleOnPolicy", () => {
+  it("sets the provided rules as the forbidden rules for the policy", () => {
+    const myPolicy = setForbiddenRuleOnPolicy(
+      mockPolicy("https://some.pod/policy-resource#policy"),
+      [
+        mockRule("https://some.pod/rule-resource#a-rule"),
+        mockRule("https://some.pod/rule-resource#another-rule"),
+      ]
```
Vinnl, line 314: I'm not sure if we want to support setting multiple rules at the same time yet, because it might lead to API inconsistencies. Specifically, for e.g. `setInteger()` and the likes, we don't support setting multiple integers at the same time yet either, because we're not sure yet what the API will look like for lists vs. multiple triples with the same predicate.

(Also, `addForbiddenRuleToPolicy` also doesn't support multiple Rules at the same time, I think?)

src/acp/rule.test.ts
NSeydoux: The function name is changed in https://github.com/inrupt/solid-client-js/pull/525. Even if it is indeed only used here to add rules to policy, it is actually more generic than that. Should the name reflect how it is used here, or what it does in general ?

src/acp/rule.test.ts
```diff
@@ -0,0 +1,704 @@
...
+  it("does not change the input policy", () => {
+    const myPolicy = mockPolicy("https://some.pod/policy-resource#policy");
+    setRequiredRuleOnPolicy(myPolicy, [
+      mockRule("https://some.pod/rule-resource#rule"),
+    ]);
+    expect(myPolicy.size).toEqual(0);
+  });
+});
+
+describe("getForbiddenRulesOnpolicyAll", () => {
```
Vinnl, line 607:
```suggestion
describe("getForbiddenRuleOnPolicyAll", () => {
```

src/acp/rule.test.ts
NSeydoux: The intent was indeed to add rule-specific behaviour in this function, which is done in https://github.com/inrupt/solid-client-js/pull/525.

However, for this comment as for the previous one, it is true that these functions might be useful as part of our public API. I'll create a ticket to this intent, so that it does'nt clog this (already quite large) PR.

Vinnl (changes requested):
Overall looks good, but I'd have the `set*()` functions accept only a single rule for now for consistency, and add `remove*()` functions. What do you think?

(As a note: I'd decided to allow myself to use the basic solid-client functions in the ACP stuff, since it's layered on top of that and thus could theoretically also be abstracted out into a separate library. Hope that's OK with you.)

src/acp/rule.test.ts
NSeydoux: I updated it in 556fa15647a426615498bb8ea4d01f4922b4edd3. I kept checking the size of the newly created policy out, because that is already tested for in the test verifying that the function actually adds the rule to the policy, so it would make this test somewhat redundant and testing for several things.

src/acp/rule.test.ts
Vinnl: I'd always go with how it's intended to be used - otherwise you might wrongly assume it is and will remain more generic than it is, re-use it elsewhere in the codebase, and then suddenly break that if you modify it for the use case here.

src/acp/rule.test.ts
NSeydoux: Right, I figured that there would be more use cases involving setting multiple rules at once compared to setting e.g. integers, but I didn't think about the issue of API consistency when it comes to lists. I'll fix this.

NSeydoux: Fixed in 802bbdf0a94e9fbcb7e41840d2001d499a94add7

NSeydoux:
Added the `remove*` in b01bc1ba2de22f3facffdae12218d42b51d04a3c

src/acp/rule.test.ts
NSeydoux: That makes sense. I'll rename these in the next PR, to avoid conflicts.

NSeydoux:
Overall @pmcb55, I agree it feels like there is too much test code for the size of the change. I think @Vinnl proposal to "test for implementation" in that we verify that the lower-level solid-client API is called, rather than doing a full-sized spec of the API when it is such a thin layer, may be a good resolution to that problem.

Vinnl (approved):
That was quick :)

Btw, the size of the test code seems fine to me; it's the number of tests that makes it so long. It's debatable whether the amount of effort to write them outweighs the benefits of having them - some of them are rather unlikely to break (e.g. `removeRequired*` remove optional rules), but then again the effort to add/maintain them isn't too great either.

src/acp/rule.test.ts
NSeydoux: The similar code snippet, using named nodes constants, would be change as in the following: 
**Before**
```
const myPolicy = addOptionalRuleToPolicy(
      mockPolicy("https://some.pod/policy-resource#policy"),
      mockRule("https://some.pod/rule-resource#rule")
    );
    expect(
      myPolicy.has(
        DataFactory.quad(
          DataFactory.namedNode("https://some.pod/policy-resource#policy"),
          DataFactory.namedNode(ACP_ANY),
          DataFactory.namedNode("https://some.pod/rule-resource#rule")
        )
      )
    ).toBe(true);
```
**After**
```
it("adds the rule in the forbidden rules of the policy", () => {
    const myPolicy = addForbiddenRuleToPolicy(
      mockPolicy(MOCKED_POLICY_IRI),
      mockRule(MOCKED_RULE_IRI)
    );
    expect(
      myPolicy.has(
        DataFactory.quad( MOCKED_POLICY_IRI, ACP_NONE, MOCKED_RULE_IRI )
      )
    ).toBe(true);
  });
```
It is indeed more compact, but I'm not sure it is more readable. What do you think ?

src/acp/rule.test.ts
pmcb55: Well, I certainly always think of a mock as mocking behaviour (i.e. "**_So in simple words, mocking is creating objects that simulate the behavior of real objects._**"). This is just test data in a perfectly valid, un-mocked Thing object. Personally I'd just rename this to `createTestPolicy()`. (Likewise, if I had a test WebID value, I wouldn't call that `const mockWebId = new Iri('https://...);`, I'd just call it a `testWebId`.)
And yeah, I'm fine not making a general helper function, but now I'm just curious - how is this implementation mutating the input (seems it's just reading that input from what I can see)...?

src/acp/rule.test.ts
pmcb55: Yeah, I think it's much more readable. Just to add too though, the `MOCKED_POLICY_IRI` etc. values I'd actually move into the existing Inrupt test vocabulary (https://github.com/inrupt/solid-common-vocab-rdf/blob/master/inrupt-rdf/Common/CopyOfVocab/inrupt-test.ttl)  (and as above, I wouldn't refer to them as MOCKs either).

src/acp/rule.test.ts
Vinnl: Hmm OK, but then we'd have to go over the entire code base - everywhere we've been using test data so far, we've referred to it as mock data. Maybe for consistency leave it as is, and we can go over the entire codebase some time to change that.

As for the mutation: `addAll` calls `thing.add`, which is an RDF/JS Dataset method that mutates the Dataset itself.

pmcb55: Just on the mutation thing - sure, it's mutating the instance of the object returned from `createThing()`, but that's a local function variable, not an **_input_** to the function itself. So it seems fine for this function to be treated as a black box (so it can do what it likes internally), but it's not mutating it's input at all, and therefore it's fine, no...?

Vinnl: Ehm yep you're right, `mockPolicy` should be safe, `addAll` is not. (Though in a production implementation I'd still prefer it to be built on the primitives we ship ourselves.)

pmcb55: Ok, cool (it kills me if I think I'm missing something :) !). On the 'building on our own primitives' question - yeah, I think that very often comes down to a subjective choice. In this particular case though, I actually really like that we are building on a primitive-primitive - i.e. RDF/JS itself is the primitive that our library sits on top of, so in fact I really like that we're proving out that we can use it directly with our `solid-client-js` primitives (arguably we should be proving that out in dedicated tests, rather than it being scattered randomly across our tests, but you get my point!).
It kinda highlights a really good question though - i.e. why did Nic find it **_easier_** to use RDF/JS primitives directly in this test code **_instead_** of `solid-client-js` primitives, which by definition should be 'easier' than RDF/JS....!? (I'm not saying this is highlighting a problem or anything, in fact I really like that we have the choice here, and that our tests are illustrating the ability to exercise that choice, but perhaps our `solid-client-js` SDK is still missing something that Nic could have used here even more naturally...?)

Vinnl: @NSeydoux correct me if I'm wrong, but I'm assuming he used the primitives here because that's what we've been doing for the tests for our own primitives, for exactly the reason you mention: to prove that we can use those directly.

However, given that we've already proven that in the unit tests for our primitives, IMO it'd be fine to not do that again in our code that is layered on top of our primitives. (Which is why I asked whether it's OK that I did that in my ACP-related PRs.

NSeydoux: Exactly as Vincent says: mock data is created using RDFJS primitives not because they are simpler to use, but because we've been using them across the codebase to show interoperability. Also, I feel like using these primitives make the test a tiny bit more DAMP, because we rebuild our primitives "from scratch", so we don't make hidden assumptions on anything our library may add on top of the rdfjs datasets. But I agree that we are now dealing with a higher level of abstraction in the library, layered on top of our own primitives, so it should be fine to use these as well.

pmcb55: Ok cool - so we're all the same page then I think (although I'd refer to these examples as 'creating test data', and not 'creating mock data', as the word 'mock' (certainly in my head, and I'm pretty sure across the industry) is associated with mocking **_behaviour_**, and not related to pure data).
So yeah, the choice of using low-level primitives (e.g. RDF/JS constructs), or our higher-level primitives (e.g. Things), will always be a somewhat subjective choice, and I'm fine with either really, but in general I suppose the idea would be to try to limit the number of abstractions to one. So in this particular case, if we agree that ACP is a 'layer on top', then sure, using `solid-client-js` constructs here (as opposed to RDF/JS constructs) would be totally fine.

pmcb55 (approved):
LGTM - and great discussion!

src/acp/rule.ts
```diff
@@ -0,0 +1,235 @@
...
+import { acp } from "../constants";
+import { SolidDataset, ThingPersisted, UrlString } from "../interfaces";
+import { addIri } from "../thing/add";
+import { getIriAll } from "../thing/get";
+import { removeAll, removeIri } from "../thing/remove";
+import { setIri } from "../thing/set";
+import { Policy } from "./policy";
+
+export type RuleDataset = SolidDataset;
+export type Rule = ThingPersisted;
```
Vinnl, line 31: I just realised: when separating Policies from Controls, I made it so that you didn't give the actual _Policies_ to the Controls, but their _URLs_. This to indicate that you have to make sure that the Policies are saved somewhere, not necessarily in the same Resource as the Controls themselves.

I think something similar should apply here. Even though usually a Rule will (hopefully) be located in the same Resource as the Policies, it might not be. So maybe we should update our functions to say `addRequiredRuleUrlToPolicy()` rather than `addRequiredRuleToPolicy()`, just like we have `addPolicyUrl()` rather than `addPolicy()`.

(And I suppose for consistency, the latter should be renamed to `addPolicyUrlToControl()` as well? Though in a separate PR.)

And then maybe this should become:

```suggestion
export type RuleUrl = ThingPersisted | Url | UrlString;
```

Vinnl (changes requested):
Just thought of another inconsistency between this and adding Policies to Controls. Either this or that should be aligned with the other - let me know which you think is preferable :)

src/acp/rule.ts
pmcb55: Firstly, I think I prefer `addRequiredRuleReferenceToPolicy()` over `addRequiredRuleUrlToPolicy()` (so, yeah, that would probably ripple across lots more name re-alignment too, as separate topic/PR), but yeah I agree with the principle of having explicitly named functions to add a Policy/Rule directly, and to add a reference to a Policy/Rule. And yeah, ditto for the Policy and the Access Control Resource.

NSeydoux: Fixed in eb3d09e413b6b6d9ca69a9df7741411e6d914740

src/acp/rule.test.ts
NSeydoux: Fixed in 5f1e895c4a9284fba126ba97dce97d20e312d00f

Vinnl (approved):
It's a shame that change wasn't there initially, as it could've simplified writing a bunch of tests (i.e. just passing the URL rather than having to set up an entire Rule), but ah well :P

Vinnl:
I'll fix the conflicts.

PR #525, opened 2020-10-19
Add functions to manage agents/groups in rules

NSeydoux:
This adds a set of functions enabling manipulation of the agents a rule applies to (adding/removing one, setting a batch of agents, listing the agents a rule applies to). Additionally to agents and groups identified by IRI, a couple of functions are dedicated to manipulating specific agent classes, namely authenticated agents and public agents.

Note that this is still part of an experimental API, and as such it is not publicly exposed in index.ts, nor is it mentioned in the changelog.
- [ ] The changelog has been updated, if applicable. **N/A**
- [ ] New functions/types have been exported in `index.ts`, if applicable. **N/A**

### DIFF
--- a/src/acp/rule.test.ts
+++ b/src/acp/rule.test.ts
@@ -19,37 +19,72 @@
  * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
-import { jest, describe, it, expect } from "@jest/globals";
-import { asIri, createThing } from "../thing/thing";
+import { describe, it, expect } from "@jest/globals";
+import { asIri, createThing, setThing } from "../thing/thing";
 import {
+  addAgentToRule,
   addForbiddenRuleToPolicy,
+  addGroupToRule,
   addOptionalRuleToPolicy,
   addRequiredRuleToPolicy,
+  createRule,
+  getAgentForRuleAll,
   getForbiddenRuleOnPolicyAll,
+  getGroupForRuleAll,
   getOptionalRuleOnPolicyAll,
   getRequiredRuleOnPolicyAll,
   removeForbiddenRuleFromPolicy,
   removeOptionalRuleFromPolicy,
   removeRequiredRuleFromPolicy,
+  getRule,
+  hasAuthenticatedInRule,
+  hasPublicInRule,
+  removeAgentFromRule,
+  removeGroupFromRule,
   Rule,
+  setAgentInRule,
+  setAuthenticatedForRule,
   setForbiddenRuleOnPolicy,
+  setGroupInRule,
   setOptionalRuleOnPolicy,
+  setPublicForRule,
   setRequiredRuleOnPolicy,
 } from "./rule";
 import { DataFactory } from "n3";
-import { Thing, ThingPersisted } from "../interfaces";
+import { Thing, ThingPersisted, UrlString, WebId } from "../interfaces";
 import { Policy } from "./policy";
+import { createSolidDataset } from "../resource/solidDataset";
+import { setUrl } from "../thing/set";
 
 const ACP_ANY = "http://www.w3.org/ns/solid/acp#anyOf";
 const ACP_ALL = "http://www.w3.org/ns/solid/acp#allOf";
 const ACP_NONE = "http://www.w3.org/ns/solid/acp#noneOf";
 
-const mockRule = (url: string) =>
-  createThing({
-    url,
-  });
+const ACP_RULE = "http://www.w3.org/ns/solid/acp#Rule";
+const ACP_AGENT = "http://www.w3.org/ns/solid/acp#agent";
+const ACP_GROUP = "http://www.w3.org/ns/solid/acp#group";
+const ACP_PUBLIC = "http://www.w3.org/ns/solid/acp#PublicAgent";
+const ACP_AUTHENTICATED = "http://www.w3.org/ns/solid/acp#AuthenticatedAgent";
 
-const addAll = (
+const RDF_TYPE = "http://www.w3.org/1999/02/22-rdf-syntax-ns#type";
+
+const addAllIri = (
+  thing: ThingPersisted,
+  predicate: string,
+  objects: UrlString[]
+): void => {
+  objects.forEach((objectToAdd: UrlString) => {
+    thing.add(
+      DataFactory.quad(
+        DataFactory.namedNode(asIri(thing)),
+        DataFactory.namedNode(predicate),
+        DataFactory.namedNode(objectToAdd)
+      )
+    );
+  });
+};
+
+const addAllThings = (
   thing: ThingPersisted,
   predicate: string,
   objects: ThingPersisted[]
@@ -65,19 +100,65 @@ const addAll = (
   });
 };
 
+const mockRule = (
+  url: string,
+  content?: {
+    agents?: WebId[];
+    groups?: UrlString[];
+    public?: boolean;
+    authenticated?: boolean;
+  }
+): Rule => {
+  const mockedRule = createThing({
+    url,
+  });
+  mockedRule.add(
+    DataFactory.quad(
+      DataFactory.namedNode(asIri(mockedRule)),
+      DataFactory.namedNode(RDF_TYPE),
+      DataFactory.namedNode(ACP_RULE)
+    )
+  );
+  if (content?.agents) {
+    addAllIri(mockedRule, ACP_AGENT, content.agents);
+  }
+  if (content?.groups) {
+    addAllIri(mockedRule, ACP_GROUP, content.groups);
+  }
+  if (content?.public) {
+    mockedRule.add(
+      DataFactory.quad(
+        DataFactory.namedNode(asIri(mockedRule)),
+        DataFactory.namedNode(ACP_AGENT),
+        DataFactory.namedNode(ACP_PUBLIC)
+      )
+    );
+  }
+  if (content?.authenticated) {
+    mockedRule.add(
+      DataFactory.quad(
+        DataFactory.namedNode(asIri(mockedRule)),
+        DataFactory.namedNode(ACP_AGENT),
+        DataFactory.namedNode(ACP_AUTHENTICATED)
+      )
+    );
+  }
+  return mockedRule;
+};
+
 const mockPolicy = (
   url: string,
   rules?: { required?: Rule[]; optional?: Rule[]; forbidden?: Rule[] }
 ): Policy => {
   const mockPolicy = createThing({ url });
   if (rules?.forbidden) {
-    addAll(mockPolicy, ACP_NONE, rules.forbidden);
+    addAllThings(mockPolicy, ACP_NONE, rules.forbidden);
   }
   if (rules?.optional) {
-    addAll(mockPolicy, ACP_ANY, rules.optional);
+    addAllThings(mockPolicy, ACP_ANY, rules.optional);
   }
   if (rules?.required) {
-    addAll(mockPolicy, ACP_ALL, rules.required);
+    addAllThings(mockPolicy, ACP_ALL, rules.required);
   }
   return mockPolicy;
 };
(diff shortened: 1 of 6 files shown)


src/acp/rule.test.ts
```diff
@@ -702,3 +783,778 @@ describe("getRequiredRulesOnPolicyAll", () => {
...
+  });
+});
+
+describe("setAgentInRule", () => {
+  it("sets the given agents for the rule", () => {
+    const rule = mockRule("https://my.pod/rule-resource#rule");
+    const result = setAgentInRule(rule, [
+      "https://my.pod/profile#me",
+      "https://your.pod/profile#you",
+    ]);
```
Vinnl, line 867: Like in #520, I wouldn't make this accept an array.

src/acp/rule.test.ts
```diff
@@ -702,3 +783,778 @@ describe("getRequiredRulesOnPolicyAll", () => {
...
+      authenticated: true,
+    });
+    const groups = getGroupForRuleAll(rule);
+    expect(groups).not.toContainEqual("https://my.pod/profile#me");
+    expect(groups).not.toContainEqual(ACP_AUTHENTICATED);
+    expect(groups).not.toContainEqual(ACP_PUBLIC);
+  });
+});
+
+describe("setGroupInRule", () => {
```
Vinnl, line 1157: Just wondering: would it be a good idea to have consistency of either `getGroupForRuleAll()`/`setGroupForRule()` or `getGroupInRuleAll()`/`setGroupInRule()`? Or do the different prepositions just make more sense in the different context? (Not being a native speaker getting me down here...)

src/acp/rule.ts
```diff
@@ -197,3 +211,283 @@ export function setForbiddenRuleOnPolicy(
...
 }
+
+/**
+ * ```{note} There is no Access Control Policies specification yet. As such, this
+ * function is still experimental and subject to change, even in a non-major release.
+ * ```
+ *
+ * Initialise a new empty [[SolidDataset]] to store [[Rule]]'s in.
+ */
+export const createRuleDataset = createSolidDataset;
```
Vinnl, line 222: Thinking: by exposing this function, we are stimulating people to keep their rules in a separate Resource. Would it be better to encourage people to add them to their APRs, i.e. local to the policies that refer to them?

src/acp/rule.ts
```diff
@@ -197,3 +211,283 @@ export function setForbiddenRuleOnPolicy(
...
+ * that are part of a group the [[Rule]] applies to, nor will it include specific agent
+ * classes, such as authenticated or public agents.
+ *
+ * @param rule The rule from which agents are read.
+ * @returns A list of the WebIDs of agents included in the rule.
+ * @since Unreleased
+ */
+export function getAgentForRuleAll(rule: Rule): WebId[] {
+  return getIriAll(rule, acp.agent).filter(
+    (agent: string) =>
```
Vinnl, line 277:
```suggestion
    (agent: WebId) =>
```

src/acp/rule.ts
```diff
@@ -197,3 +211,283 @@ export function setForbiddenRuleOnPolicy(
...
+      agent !== acp.PublicAgent && agent !== acp.AuthenticatedAgent
+  );
+}
+
+/**
+ * ```{note} There is no Access Control Policies specification yet. As such, this
+ * function is still experimental and subject to change, even in a non-major release.
+ * ```
+ *
+ * Overrite the agents the [[Rule]] applies to with the provided agents.
```
Vinnl, line 287:
```suggestion
 * Override the agents the [[Rule]] applies to with the provided agents.
```
or did you mean

```suggestion
 * Overwrite the agents the [[Rule]] applies to with the provided agents.
```

Either works :P

src/acp/rule.ts
```diff
@@ -197,3 +211,283 @@ export function setForbiddenRuleOnPolicy(
...
+ * @since Unreleased
+ */
+export function setAgentInRule(rule: Rule, agents: WebId[]): Rule {
+  // Preserve the special agent classes authenticated and public, which we
+  // don't want to overwrite with this function.
+  // NB: This raises the question wether the named agents and agent classes should
+  // be linked to with the same predicate.
+  const isPublic = hasPublicInRule(rule);
+  const isAuthenticated = hasAuthenticatedInRule(rule);
+  let result = cloneThing(rule);
```
Vinnl, line 301: Given that `removeAll` and `setIri` already return a clone, there's no need to do so manually.

```suggestion
  let result: Rule;
```

src/acp/rule.ts
```diff
@@ -197,3 +211,283 @@ export function setForbiddenRuleOnPolicy(
...
+  if (agents.length === 0) {
+    result = removeAll(rule, acp.agent);
+  } else {
+    result = setIri(rule, acp.agent, agents[0]);
+    agents.slice(1).forEach((agent) => {
+      result = addIri(result, acp.agent, agent);
+    });
```
Vinnl, line 308: I think this might be a bit easier to follow when not having to mentally split the agents array into two groups:

```suggestion
    result = removeAll(rule, acp.agent);
    agents..forEach((agent) => {
      result = addIri(result, acp.agent, agent);
    });
```

src/acp/rule.ts
Vinnl: Btw, this and the other comments below also apply to groups.

Vinnl (changes requested):
Implementation-wise this looks good, but like #520, I'd prefer not to pass an array to the `set*()` functions.

Also, hopefully `acp:access` won't be used to public and authenticated agents anymore, it's a bit clunky :)

src/acp/rule.ts
```diff
@@ -0,0 +1,493 @@
...
+ * Get the [[Rule]] with the given URL from an [[RuleDataset]].
+ *
+ * @param ruleResource The Resource that contains the given [[Rule]].
+ * @param url URL that identifies this [[Rule]].
+ * @returns The requested [[Rule]], if it exists, or `null` if it does not.
+ */
+export function getRule(
+  ruleResource: RuleDataset,
+  url: Url | UrlString
+): Policy | null {
```
megoth, line 254: I reckon the returned type here is wrong? Should be `Rule`?

src/acp/rule.test.ts
NSeydoux: Fixed in b67fa299daec8a7a7fe05d5d50ba370b824e25ad

src/acp/rule.test.ts
NSeydoux: @pmcb55 , could you have a look at the function names in the API ? I initially thought the prepositions would make sense in their context, but I'd prefer have that double-checked :)

src/acp/rule.ts
NSeydoux: I think reuse would be encouraged if the rules were kept in a separate resource: I'd imagine a user wanting to create only one "family" or "dog enthusiasts" rules, and manipulate agents in these, rather than having mostly resource-scoped rules, but I could see how that requires extra trust to the apps you install, if each app may mess up your permissions because they are centralised in a few resources... But even without sharing across apps, for one app having a dedicated resource to list groups with whom data is shared seems easier to manage than having resource-specific groups. What do you think ?

src/acp/rule.ts
NSeydoux: It looks like I meant both ^^

src/acp/rule.ts
NSeydoux: Since we don't take an array of agents, it simplifies the function quite a bit.

src/acp/rule.ts
NSeydoux: Fixed in 59e7de6daf35fcf97bc5f343d88b19575ee6717d

src/acp/rule.ts
NSeydoux: Fixed in e46e4f1f323ac62998356ad505726775a0b29467

src/acp/rule.ts
NSeydoux: Fixed in 59e7de6daf35fcf97bc5f343d88b19575ee6717d

src/acp/rule.ts
NSeydoux: Fixed in e46e4f1f323ac62998356ad505726775a0b29467

src/acp/rule.ts
NSeydoux: Fixed in 59e7de6daf35fcf97bc5f343d88b19575ee6717d

src/acp/rule.ts
Vinnl: I'm not sure if reuse makes sense. Sure, the "family" or "dog enthusiasts" use case makes sense, but that would be covered by a group. A rule, instead, is more for the use case "I want to allow family and disallow dog enthusiasts", which I'm not sure if you really want to re-use those across Policies? Best use case I can think of is "Alice's birthday" being an "allOf <friends>; noneOf <alice>", but even then, I think there'd be _one_ policy called "Alice's birthday", not a rule that's used in multiple policies, and especially not a rule that's used in multiple policies defined in different resources.

Perhaps we should just ask Emmet whether Rules are intended to be reused?

src/acp/rule.ts
```diff
@@ -73,6 +87,7 @@ export function removeRequiredRuleFromPolicy(
  * ```
  *
  * Overwrites the rule refining the scope of a given the [[Policy]]. If an agent
```
Vinnl, line 89: Forgot to remove the old version :) (Also on line 159.)

```suggestion
```

src/acp/rule.ts
```diff
@@ -233,3 +249,268 @@ export function setForbiddenRuleOnPolicy(policy: Policy, rule: Rule): Policy {
...
+ * @param rule The rule for which agents are set.
+ * @param agents The list of agents the rule should apply to.
+ * @returns A copy of the input rule, applying to a different set of agents.
+ * @since Unreleased
+ */
+export function setAgentInRule(rule: Rule, agent: WebId): Rule {
+  // Preserve the special agent classes authenticated and public, which we
+  // don't want to overwrite with this function.
+  // NB: This raises the question wether the named agents and agent classes should
+  // be linked to with the same predicate.
```
Vinnl, line 336: It's not clear what future readers of the code are supposed to do with this comment, so I'd just leave it out, and bring it up in the relevant forums. (Which I think you've already done, right?)

```suggestion
```

Vinnl (approved):
:+1: no major concerns for me anymore.

When merging, could you make sure that the replacing of URLs with constants is a distinct commit?

src/acp/rule.ts
Vinnl: Ah wait, no, Rules are even just groups of groups and agents. Just like... Groups. They're just groups specific to ACPs, I guess? I guess maybe then reuse makes sense?

src/acp/rule.ts
NSeydoux: Good point, fixed in 5edb865bdaaa40d15ea658d0d23f10e3bb1e7f08

src/acp/rule.ts
NSeydoux: Good catch, fixed in 62a05e84592e8cf4e65882f578ef1282565c1894

src/acp/rule.test.ts
pmcb55: Yeah, Vinnie's right - `getGroupForRuleAll()` / `setGroupForRule()` are more appropriately named.

src/acp/rule.ts
pmcb55: Yeah, I'd certainly agree with Nic on the reusability of Rules, although the LucidChart architecture diagram doesn't represent a Rule Resource (just Rule statements). So probably a good idea to get Emmet's input on that question (I'll ping him now).

src/acp/rule.ts
```diff
@@ -233,3 +247,266 @@ export function setForbiddenRuleOnPolicy(policy: Policy, rule: Rule): Policy {
...
+ * ```
+ *
+ * Overwrite the agents the [[Rule]] applies to with the provided agents.
+ *
+ * @param rule The rule for which agents are set.
+ * @param agents The list of agents the rule should apply to.
+ * @returns A copy of the input rule, applying to a different set of agents.
+ * @since Unreleased
+ */
+export function setAgentInRule(rule: Rule, agent: WebId): Rule {
```
pmcb55, line 330: Yeah, this should be named `setAgentForRule()`.

src/acp/rule.ts
```diff
@@ -233,3 +247,257 @@ export function setForbiddenRuleOnPolicy(policy: Policy, rule: Rule): Policy {
...
+ * @param rule The rule being modified.
+ * @param hasPublic A boolean indicating whether the rule should apply or not to any authenticated agent.
+ * @returns A copy of the rule, updated to apply/not apply to any authenticated agent.
+ * @status Unreleased
+ */
+export function setAuthenticatedForRule(
+  rule: Rule,
+  authenticated: boolean
+): Rule {
+  if (!authenticated) {
```
pmcb55, line 499: Being very pedantic, but on simple 'if' statements like this, I tend to favour testing positive conditions not negative (i.e. just remove the 'not' and reverse the clauses). And for conditionals **_this simple_**, I'd just use the ternary operator:
```
return authenticated : addIri(rule, acp.agent, acp.AuthenticatedAgent) : removeIri(rule, acp.agent, acp.AuthenticatedAgent);
```

NSeydoux: Fixed as part of 9107fe466f0f0d056bdb8146e6e14f59347f0213

src/acp/rule.ts
NSeydoux: I aligned all the names to the template setXForY in 9107fe466f0f0d056bdb8146e6e14f59347f0213, and we can have a deeper naming conversation as part of the dedicted ticket this sprint.

src/acp/rule.test.ts
NSeydoux: Fixed in 9107fe466f0f0d056bdb8146e6e14f59347f0213

PR #526, opened 2020-10-19
Make Access Control-functions work on WithAcrs

Vinnl:
If Solid ever gets updated to be able to send both a Resource and
its Linked Resources in a single Response, we'll have API
compatibility if we treat a Resource and its Linked Resources as a
single object.

We've done that for ACLs so far, but I forgot to treat ACRs in the
same way. This commit fixes that: operations on ACRs now take a
Resource with an attached ACR as parameter.

### DIFF
--- a/src/acp/control.test.ts
+++ b/src/acp/control.test.ts
@@ -42,10 +42,12 @@ import { acp, rdf } from "../constants";
 import { WithAccessibleAcl, WithServerResourceInfo } from "../interfaces";
 import { getIri, getIriAll } from "../thing/get";
 import { createThing, getThing, setThing } from "../thing/thing";
-import { mockAcrFor } from "./mock";
+import { addMockAcrTo, mockAcrFor } from "./mock";
 import { setIri, setUrl } from "../thing/set";
 import { DataFactory } from "n3";
 import { addIri } from "../thing/add";
+import { createSolidDataset } from "../resource/solidDataset";
+import { mockSolidDatasetFrom } from "../resource/mock";
 
 describe("hasLinkedAcr", () => {
   it("returns true if a Resource exposes a URL to an Access Control Resource", () => {
@@ -111,9 +113,13 @@ describe("getAccessControl", () => {
       mockAcrFor("https://some.pod/resource"),
       accessControl
     );
+    const resourceWithAcr = addMockAcrTo(
+      mockSolidDatasetFrom("https://arbitrary.pod/resource"),
+      accessControlResource
+    );
 
     const foundAccessControl = getAccessControl(
-      accessControlResource,
+      resourceWithAcr,
       accessControlUrl
     );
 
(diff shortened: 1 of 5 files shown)


src/acp/control.test.ts
```diff
@@ -111,12 +113,13 @@ describe("getAccessControl", () => {
       mockAcrFor("https://some.pod/resource"),
       accessControl
     );
-
-    const foundAccessControl = getAccessControl(
-      accessControlResource,
-      accessControlUrl
+    const withAcr = addMockAcrTo(
```
pmcb55, line 116: I'd just name this `resourceWithAcr`.

src/acp/control.ts
```diff
@@ -167,34 +172,62 @@ export function getAccessControlAll(
...
  * ```
  *
- * Insert an [[AccessControl]] into an [[AccessControlResource]], replacing previous instances of that Access Control.
+ * Insert an [[AccessControl]] into the [[AccessControlResource]] attached to a Resource, replacing
+ * previous instances of that Access Control.
  *
- * @param accessControlResource The Access Control Resource to insert an Access Control into.
- * @param accessControl The Access Control to insert into the given Access Control Resource.
- * @returns A new Access Control Resource equal to the given Access Control Resource, but with the given Access Control.
+ * @param withAccessControlResource A Resource with the Access Control Resource into which to insert an Access Control attached.
```
pmcb55, line 178: Not sure what the final word 'attached' is referring to here - i.e. just remove that final word, or:
```
 * @param withAccessControlResource A Resource with the Access Control Resource into which to insert the specified Access Control.
```

src/acp/control.ts
```diff
@@ -167,34 +172,62 @@ export function getAccessControlAll(
...
+  const acr = getAcr(withAccessControlResource);
+  const updatedAcr = removeThing(acr, accessControl);
+  const updatedResource = setAcr(withAccessControlResource, updatedAcr);
+  return updatedResource;
+}
+
+function getAcr(resource: WithAccessibleAcr): AccessControlResource {
+  if (!hasAccessibleAcr(resource)) {
+    throw new Error(
+      "Cannot work with Access Controls on a Resource that does not have an Access Control Resource attached."
```
pmcb55, line 215: Should we be trying to identify the Resource that does not have an ACR, i.e.:
```
"Cannot work with Access Controls on the Resource [${resource.toString()}] that does not have an Access Control Resource attached."
```
(or some meaningful equivalent of `.toString()` for a Resource)...?

pmcb55 (approved):
LGTM - just minor comments really, but lingering concerns over how intuitive all this will be to the 'average developer'...!

src/acp/control.ts
Vinnl: All references to "attached" have been removed: 93404f52fe8ab0e31bef93bf206b19c2214de6ba

src/acp/control.ts
Vinnl: Yeah good idea: fed669c1543c4f8a21fa3c03ae9fe83d62064dff

Vinnl:
Yeah not at all, average or not - it just takes too much time to learn how all this works. Hopefully a higher-level API will serve them better.

PR #527, opened 2020-10-19
No longer automatically fetch attached Policies

Vinnl:
Since they're not server-managed and will therefore never be
returned by the server together with the actual data in a single
HTTP Response, the Access Policy Resources should not be
automatically fetched by us.

Instead, we provide a convenience function that will give the
developer a list of all the existing Policy Resources that are
being referenced in an ACR, so they can fetch them themselves (and
potentially deal with errors that come up when they aren't
accessible to the current user).

### DIFF
--- a/src/acp/acp.test.ts
+++ b/src/acp/acp.test.ts
@@ -32,24 +32,25 @@ jest.mock("../fetcher.ts", () => ({
 }));
 
 import { Response } from "cross-fetch";
-import {
-  getFileWithAccessDatasets,
-  getFileWithAcp,
-  getResourceInfoWithAccessDatasets,
-  getResourceInfoWithAcp,
-  getSolidDatasetWithAccessDatasets,
-  getSolidDatasetWithAcp,
-  WithAccessibleAcr,
-} from "./acp";
 import { acp, rdf } from "../constants";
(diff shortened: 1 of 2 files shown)


src/acp/acp.test.ts
```diff
@@ -32,21 +32,22 @@ jest.mock("../fetcher.ts", () => ({
...
-import { acp, rdf } from "../constants";
 import * as SolidDatasetModule from "../resource/solidDataset";
 import * as FileModule from "../resource/nonRdfData";
 import * as ResourceModule from "../resource/resource";
-import { mockSolidDatasetFrom } from "../resource/mock";
-import { File, UrlString, WithServerResourceInfo } from "../interfaces";
-import { AccessControlResource } from "./control";
+import {
+  getFileWithAcr,
+  getReferencedPolicyUrls,
```
pmcb55, line 40: Minor point - but for consistency, should this be named `getReferencedPolicyUrlAll()` (i.e. is this literally the only function name with a plural)...?

pmcb55 (approved):
LGTM

src/acp/acp.test.ts
Vinnl: Good point: f585e8f2293a5e1184f5436fe2148f4ae06a54d8

PR #529, opened 2020-10-20
Fix headers issue in browser environment

NSeydoux:
There is an incompatibility between the cross-fetch headers and the native ones, which creates errors when using this in a browser-based app.
This is a temporary solution, until the root cause of the issue is identified.

- [ ] I've added a unit test to test for potential regressions of this bug. **Currently N/A, only testable in a browser**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ The following changes have been implemented but not released yet:
 
 ### Bugs fixed
 
-- When writing non-RDF data, the request headers were incorrectly set, resulting in incorrect headers.
+- When writing non-RDF data, the request headers were incorrectly set.
 
 The following sections document changes that have been released already:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 The following changes have been implemented but not released yet:
 
+### Bugs fixed
+
+- When writing non-RDF data, the request headers were incorrectly set, resulting in incorrect headers.
+
 The following sections document changes that have been released already:
 
 ## [0.6.1] - 2020-10-15
--- a/src/resource/nonRdfData.test.ts
+++ b/src/resource/nonRdfData.test.ts
@@ -76,10 +76,10 @@ describe("flattenHeaders", () => {
       callback("text/turtle", "Content-Type");
     };
     const flatHeaders = flattenHeaders(myHeaders);
-    expect(Object.entries(flatHeaders)).toEqual([
-      ["accept", "application/json"],
-      ["Content-Type", "text/turtle"],
-    ]);
+    expect(flatHeaders).toEqual({
+      accept: "application/json",
+      "Content-Type": "text/turtle",
+    });
   });
 
   it("transforms an incoming string[][] array into a flat headers structure", () => {
@@ -88,10 +88,10 @@ describe("flattenHeaders", () => {
       ["Content-Type", "text/turtle"],
     ];
     const flatHeaders = flattenHeaders(myHeaders);
-    expect(Object.entries(flatHeaders)).toEqual([
-      ["accept", "application/json"],
-      ["Content-Type", "text/turtle"],
-    ]);
+    expect(flatHeaders).toEqual({
+      accept: "application/json",
+      "Content-Type": "text/turtle",
+    });
   });
 });
 
--- a/src/resource/nonRdfData.test.ts
+++ b/src/resource/nonRdfData.test.ts
@@ -59,12 +59,6 @@ describe("flattenHeaders", () => {
     const myHeaders = new Headers();
     myHeaders.append("accept", "application/json");
     myHeaders.append("Content-Type", "text/turtle");
-    // The following needs to be ignored because `node-fetch::Headers` and
-    // `lib.dom.d.ts::Headers` don't align. It doesn't break the way we
-    // use them currently, but it's something that must be cleaned up
-    // at some point.
-    // eslint-disable-next-line
-    // @ts-ignore
     const flatHeaders = flattenHeaders(myHeaders);
     expect(Object.entries(flatHeaders)).toContainEqual([
       "accept",
--- a/src/resource/nonRdfData.test.ts
+++ b/src/resource/nonRdfData.test.ts
@@ -38,9 +38,56 @@ import {
   saveFileInContainer,
   overwriteFile,
   getFileWithAcl,
+  flattenHeaders,
 } from "./nonRdfData";
 import { Headers, Response } from "cross-fetch";
 import { WithResourceInfo } from "../interfaces";
+
+describe("flattenHeaders", () => {
+  it("returns an empty object for undefined headers", () => {
+    expect(flattenHeaders(undefined)).toEqual({});
+  });
+
+  it("returns well-formed headers as-is", () => {
+    const headers: Record<string, string> = {
+      test: "value",
+    };
+    expect(flattenHeaders(headers)).toEqual(headers);
+  });
+
+  it("transforms an incoming Headers object into a flat headers structure", () => {
+    const myHeaders = new Headers();
+    myHeaders.append("accept", "application/json");
+    myHeaders.append("content-type", "text/turtle");
+    // The following needs to be ignored because `node-fetch::Headers` and
+    // `lib.dom.d.ts::Headers` don't align. It doesn't break the way we
+    // use them currently, but it's something that must be cleaned up
+    // at some point.
+    // eslint-disable-next-line
+    // @ts-ignore
+    const flatHeaders = flattenHeaders(myHeaders);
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 The following changes have been implemented but not released yet:
 
+### Bugs fixed
+
+- When writing non-RDF data, the request headers were incorrectly set, resulting in incorrect headers.
```
Vinnl, line 11: This seems a bit redundant :P

```suggestion
- When writing non-RDF data, the request headers were incorrectly set.
```

or posisbly

```suggestion
- When writing non-RDF data, the request headers were incorrectly set, resulting in the save failing (when?).
```

pmcb55 (approved):
Wow - this seems to be a mess of complexity. Is the best thing to do here to **_not_** try and support both browser and NodeJS in the one codebase, in relation to HTTP headers (i.e. just have one TS type definition for browsers, and one for NodeJS)...?
Anyways, for now I'm sure this is the quickest way forward - so LGTM (kinda!)

src/resource/nonRdfData.ts
```diff
@@ -243,6 +243,36 @@ export async function overwriteFile(
...
+  if (typeof headersToFlatten.forEach !== "function") {
+    // FIXME: This will break when passed a string[][]
+    //        (as shown when the type assertions are removed).
+    return headersToFlatten as Record<string, string>;
+  }
+
+  (headersToFlatten as Headers).forEach((value: string, key: string) => {
+    flatHeaders[key] = value;
+  });
+  return flatHeaders;
```
Vinnl, line 273: The damn `string[][]` form isn't even documented anywhere. Also, I think it's technically an `Array<[string, string]>`, i.e. every element has two values.

Anyway, maybe it's best to deal with this right now as well? (And maybe also avoid `forEach`, since [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Headers) doesn't have a page for it and doesn't even list it in its compatibility table.)

```suggestion
  // If the headers are already a Record<string, string>,
  // they can directly be returned.
  if (typeof headersToFlatten.entries !== "function" && !Array.isArray(headersToFlatten)) {
    return headersToFlatten;
  }

  const flatHeaders: Record<string, string> = {};
  const headerKeys: Array<[string, string]> = (typeof headersToFlatten.entries !== "function")
    ? headersToFlatten
    : Array.from(headersToFlatten.entries());
    
  headerKeys.forEach(([key, value]) => {
    flatHeaders[key] = value;
  });
  return flatHeaders;
```

NSeydoux:
Well at some point it used to work, so I suspect there's a cleaner fix that we can apply when we identify the root cause of the issue. Occam's razor would suggest that there's a bug in our codebase, rather than there being a fundamental flaw in header management that makes TS, browser and NodeJS incompatible, but we can investigate after GA.

src/resource/nonRdfData.ts
```diff
@@ -262,7 +292,7 @@ async function writeFile(
     ...defaultGetFileOptions,
     ...options,
   };
-  const headers = new Headers(config.init?.headers ?? {});
+  const headers = new CrossHeaders(config.init?.headers ?? {});
```
Vinnl, line 295: In case cross-fetch's Headers conflict with native fetch, might be better to avoid it:

```suggestion
  const headers = flattenHeaders(config.init?.headers ?? {});
```

and then have the subsequent functions that take `headers` as an argument only accept a `Record<string, string>`.

src/resource/nonRdfData.test.ts
```diff
@@ -38,10 +38,57 @@ import {
...
+  it("transforms an incoming Headers object into a flat headers structure", () => {
+    const myHeaders = new Headers();
+    myHeaders.append("accept", "application/json");
+    myHeaders.append("content-type", "text/turtle");
+    // The following needs to be ignored because `node-fetch::Headers` and
+    // `lib.dom.d.ts::Headers` don't align. It doesn't break the way we
+    // use them currently, but it's something that must be cleaned up
+    // at some point.
+    // eslint-disable-next-line
+    // @ts-ignore
```
Vinnl, line 67: Can't we just fix that by having `flattenHeaders` accept both cross-fetch's and the regular `Headers`? It should flatten them in the same, compatible way, so it sounds like that should work fine?

Vinnl: And if not, we'd probably better use `@ts-expect-error`, so that we get notified when we can remove this: https://devblogs.microsoft.com/typescript/announcing-typescript-3-9-beta/#ts-expect-error-comments

src/resource/nonRdfData.test.ts
```diff
@@ -38,10 +38,57 @@ import {
...
+    const flatHeaders = flattenHeaders(myHeaders);
+    expect(Object.entries(flatHeaders)).toEqual([
+      ["accept", "application/json"],
+      ["content-type", "text/turtle"],
+    ]);
```
Vinnl, line 72: Is there a reason you're converting this back to `Array<[string, string]>`, rather than

```suggestion
    expect(flatHeaders).toEqual({
      "accept", "application/json",
      "content-type", "text/turtle",
    });
```

?

Vinnl (changes requested):
Looks good, but I'd like to avoid issues with passing cross-fetch's Header to native fetch as well, as well as issues with `string[][]` headers.

src/resource/nonRdfData.ts
```diff
@@ -243,6 +243,36 @@ export async function overwriteFile(
   return Object.assign(blobClone, { internal_resourceInfo: resourceInfo });
 }
 
+/**
+ * @hidden
+ * This function feels unnecessarily complicated, but is required in order to
+ * have Headers according to type definitions in both Node and browser environments.
```
Vinnl, line 249: Btw, "type definitions" should have nothing to do with this, as they have no runtime effect. I'm guessing that `fetch` does an `instanceof` check, and since the global `Headers` are apparently not equal to the `Headers` from cross-fetch`, it treats it as a plain object.

CHANGELOG.md
NSeydoux: The headers were really incorrectly incorrect.

src/resource/nonRdfData.ts
NSeydoux: I had to make a few twists on your suggestion, but the idea is in f7363578f85359d64cb48d726fd1847b9c323d1f

src/resource/nonRdfData.test.ts
NSeydoux: You're right, that's not needed anymore. Removed in 1934686aab5c9f96dfdfbec3d623fe336e6c9ea4.

src/resource/nonRdfData.test.ts
NSeydoux: Maybe there was, but I don't remember what it was, and the alternative looks better, fixed in 1c86607ffeae8c20582177afbf4aba6b0021e586

src/resource/nonRdfData.test.ts
```diff
@@ -38,10 +38,63 @@ import {
...
+      callback: (value: string, key: string) => void
+    ): void => {
+      callback("application/json", "accept");
+      callback("text/turtle", "Content-Type");
+    };
+    const flatHeaders = flattenHeaders(myHeaders);
+    expect(Object.entries(flatHeaders)).toEqual([
+      ["accept", "application/json"],
+      ["Content-Type", "text/turtle"],
+    ]);
```
Vinnl, line 82: The same change could be made here :P 

```suggestion
    expect(flatHeaders).toEqual({
      "accept", "application/json",
      "Content-Type", "text/turtle",
    });
```

src/resource/nonRdfData.test.ts
```diff
@@ -38,10 +38,63 @@ import {
...
+  });
+
+  it("transforms an incoming string[][] array into a flat headers structure", () => {
+    const myHeaders: string[][] = [
+      ["accept", "application/json"],
+      ["Content-Type", "text/turtle"],
+    ];
+    const flatHeaders = flattenHeaders(myHeaders);
+    expect(Object.entries(flatHeaders)).toEqual([
```
Vinnl, line 91: And here.

Vinnl (approved):

src/resource/nonRdfData.test.ts
Vinnl: :+1: two more places were the same applies I suppose :)

src/resource/nonRdfData.test.ts
NSeydoux: Oops cd6fcff7f4e8450c8c1168087d342ca4a48a669a

src/resource/nonRdfData.test.ts
NSeydoux: Maybe they could ^^ cd6fcff7f4e8450c8c1168087d342ca4a48a669a

PR #547, opened 2020-10-26
Delete docs_request.md

nicolasmondada:
There is already a docs feedback button on the docs site


NSeydoux (approved):
Should the main README be updated to point out the existence of this documentation feedback ? I assume the documentation feedback creates a helpdesk ticket, is that correct ?

Vinnl:
@nicolasmondada Do you know if there's a link for the docs feedback form? I can only open it in a modal on the docs site, unfortunately, but it is possible for us to provide links to other places in the "New issue" form, so we could replace this with that link for people (like me) who would be inclined to go directly to a project's issue tracker, rather than look for a feedback button on the docs site.

nicolasmondada:
> Should the main README be updated to point out the existence of this documentation feedback ? I assume the documentation feedback creates a helpdesk ticket, is that correct ?

The best way to report problems is the service desk feedback form referenced to in the readme. The documentation lives on docs.inrupt.com (we point to that in the readme too) where there is another way to report docs issues. 
Screenshot 2020-10-27 at 13 01 07

> @nicolasmondada Do you know if there's a link for the docs feedback form? I can only open it in a modal on the docs site, unfortunately, but it is possible for us to provide links to other places in the "New issue" form, so we could replace this with that link for people (like me) who would be inclined to go directly to a project's issue tracker, rather than look for a feedback button on the docs site.
We have the link to the service desk feedback form in the readme.

Vinnl:
@nicolasmondada Right, but that's only helpful if you're currently _at the docs_ and have seen that button. If you're used to open source projects and just visit GitHub issues to report these, then it would be helpful to be pointed to the correct place to leave your feedback. See e.g. https://github.com/inrupt/solid-client-js/pull/554/files, where I'd like to have included a pointer to the Docs service desk.

PR #548, opened 2020-10-27
Update feature_request.md

nicolasmondada:


NSeydoux (approved):

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -1,27 +1,38 @@
 ---
 name: Feature request
-about: Suggest an idea for this project
+about: Suggest an idea for a new feature
 title: ""
-labels: "enhancement"
```
Vinnl, line 5: `enhancement` was a default GitHub label (that we've already used in the past as well), so I think it might be best to leave it as-is, in case GitHub does clever things with it. (For example, it also has a `good first issue` label, and it has pages where it directs people who want to contribute to open source to projects that use that label.)

Vinnl (changes requested):

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -1,27 +1,38 @@
...
+labels: "new feature"
 assignees: ""
 ---
 
-**Search terms you've used**
+<!-- 
+Thank you for reporting an issue.
 
+Please fill in as much of the template below as you're able. Feel free to delete any section you want to skip.
```
Vinnl, line 12: This is good, but I'd still encourage people to fill in as much as they can.

```suggestion
Please fill in as much of the template below as you're able. Feel free to delete any section you don't think applies, but be aware that the more comprehensive your description, the easier it is to take it into account. 
```

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -1,27 +1,38 @@
...
-**Is your feature request related to a problem? Please describe.**
+## Suggestion
+
+<!-- A summary of what you'd like to see added or changed -->
 
-<!-- A clear and concise description of what the problem is. Ex. I'm always frustrated when [...] -->
+## Expected functionality/enhancement
 
-**Describe the solution you'd like**
+<!-- How should the new feature work -->
```
Vinnl, line 24: To give us some leeway in the implementation, i.e. if there are issues with the proposed solution in the context of the larger codebase, but an alternative solution would still satisfy their use case.

```suggestion
<!-- How would you like the new feature work -->
```

Vinnl (changes requested):

Vinnl (approved):

PR #553, opened 2020-10-27
Add browser-based end-to-end tests

Vinnl:
# New feature description

This adds a top-level export `acp_lowlevel_preview` that ~~reckless developers~~ people can import to get all the ACP functions, and an end-to-end test running in the browser (because they need to be authenticated) that uses them to create a Policy Resource, make it publicly accessible, and delete it again.

~~Still to do: use the correct ESS credentials, and fix the Windows test (it's using the wrong slash when switching directories...). But it can already be reviewed.~~ Done

~~Also, the build is failing due to inrupt.net having issues, but you can see a working run here: https://github.com/inrupt/solid-client-js/runs/1316068235?check_suite_focus=true~~

See /src/e2e-browser/README.md for a general description of the setup.

# Checklist


.codesandbox/sandbox/src/end-to-end-test.ts
```diff
@@ -0,0 +1,42 @@
...
+loginButton.addEventListener("click", async (e) => {
+  const idp = (document.getElementById("idp") as HTMLInputElement).value;
+  await session.login({
+    oidcIssuer: idp,
+    redirectUrl: document.location.href,
+  });
+});
+
+if (!session.info.isLoggedIn) {
+  session.handleIncomingRedirect(window.location.href).then(async () => {
```
NSeydoux, line 23: I think this is alway going to be true, since the session cannot be logged in before `handleincomingRedirect` has been called. In any case, if the session is already logged in, calling `handleincomingRedirect` just returns the current session information, so I think the outer `if` can be removed.

NSeydoux (approved):
I'm not very familiar with TestCafé, so I'm not 100% confident I didn't miss anything, but overall this looks good, and isolating the low-level helpers from the test logic makes it easy to read.

.codesandbox/sandbox/package.json
```diff
@@ -1,14 +1,15 @@
...
+  "description": "Example project calling out to solid-client",
   "main": "index.html",
   "scripts": {
     "start": "parcel index.html --open",
     "build": "parcel build index.html"
   },
   "dependencies": {
-    "@inrupt/solid-client": "0.1.0",
+    "@inrupt/solid-client": "../../",
+    "@inrupt/solid-client-authn-browser": "*",
```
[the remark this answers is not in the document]
Vinnl: Yeah, but maybe I should just limit it to version 0.2.2 until 1.x, to avoid breaking these, I suppose.

.codesandbox/sandbox/src/end-to-end-test.ts
```diff
@@ -0,0 +1,42 @@
+import * as SolidClient from "@inrupt/solid-client";
+import {
+  Session,
+  getClientAuthenticationWithDependencies,
+} from "@inrupt/solid-client-authn-browser";
+import { getHelpers } from "./end-to-end-test-helpers";
+
+const session = new Session({
+  clientAuthentication: getClientAuthenticationWithDependencies({}),
```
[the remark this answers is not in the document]
Vinnl: RIP

.codesandbox/sandbox/src/end-to-end-test.ts
Vinnl: Hmm that's great... I copied that from here though: https://docs.inrupt.com/developer-tools/javascript/client-libraries/tutorial/authenticate/#example

Vinnl: Ah, I presume that is a remnant from when sessions could survive between refreshes, in which case the outer `if` was useful?

.codesandbox/sandbox/src/end-to-end-test-helpers.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  async function setAprPublicRead() {
+    const resourceWithApr = await acp.getSolidDatasetWithAcr(aprUrl, {
+      fetch: session.fetch,
+    });
+    if (!acp.hasAccessibleAcr(resourceWithApr)) {
+      throw new Error(
+        `The test Resource at [${getSourceUrl(
+          resourceWithApr
+        )}] does not appear to have an Access Control Resource. Please check the Pod setup.`
```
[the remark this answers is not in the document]
Vinnl: bef20b0a9a64d928c04eee6ead2750db3d6584c8

.codesandbox/sandbox/package.json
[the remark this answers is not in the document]
Vinnl: Alright, it's now `0.2.2 - 1.x`. (Assuming we're not doing any more breaking changes for v1.)

Vinnl:
The e2e tests now also run in Windows (against Chrome, Firefox and Edge) and MacOS (Safari only, with [a somewhat cumbersome workaround](https://devexpress.github.io/testcafe/documentation/guides/continuous-integration/github-actions.html#step-2---create-a-job) - but at least we have Safari).

src/e2e-browser/acp.testcafe.ts
```diff
@@ -0,0 +1,83 @@
...
+import { config } from "dotenv-flow";
+import { resolve } from "path";
+import { essUser, essUserLogin } from "./roles";
+import type { getHelpers } from "../../.codesandbox/sandbox/src/end-to-end-test-helpers";
+
+// E2eHelpers is a global defined in .codesandbox/sandbox/src/end-to-end-helper.
+// Since code inside of ClientFunction is executed in that context in the browser,
+// that variable is available to it - but as far as TypeScript is concerned,
+// it is executed in the context of this test file.
+// Hence, we just declare this variable to be of the  type here.
```
pmcb55, line 32: Missing the actual type here (although I tend to avoid putting actual specifics in comments if possible, so I'd just allude to the nature of the type (in English) in the comment, not the actual type name itself):
```
// Hence, we just declare this variable to be of the ????? type here.
```

Vinnl: The type is `ReturnType<typeof getHelpers>` - don't know why I inserted a double space there. Possibly I inteded to type "to be of the same type" - would that help?

pmcb55: Yeah sure, that's more readable alright. To be honest though, I still don't understand the overall comment, but I certainly appreciate the obvious effort you put into writing it (and the very extensive README). I need to read up on TestCafe's `ClientFunction` thing I guess, and also on CodeSandbox, and then carefully re-read through this whole PR, as I suspect there's lots of cool stuff going on here! Cheers.

Vinnl: Yeah it's all a bit hairy. Maybe as a helpful metaphor, imagine that the test code was written in Java. The code inside `ClientFunction` would then still be JavaScript, because it's not actually test code.

The difference is that in Java, the JS would have to be written inside a string, which the test runner would then inject into the page - and then of course transform the return value back into data that Java understands. Since both are in JS, however, TestCafe can take the AST of the contents of `ClientFunction`, serialise that to inject it into the page, and then inject the resulting value into the test again.

PR #570, opened 2020-11-02
DOCS-208: update matomo tracking code

kay-kim:
Use new tracking code for docs

### DIFF
--- a/docs/api/_templates/layout.html
+++ b/docs/api/_templates/layout.html
@@ -14,16 +14,16 @@
 
   <!-- Matomo -->
   <script type="text/javascript">
-    var _paq = window._paq || [];
+    var _paq = window._paq = window._paq || [];
     /* tracker methods like "setCustomDimension" should be called before "trackPageView" */
     _paq.push(['trackPageView']);
     _paq.push(['enableLinkTracking']);
     (function() {
-      var u="https://mat.inrupt.com/";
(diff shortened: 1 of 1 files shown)


docs/api/_templates/layout.html
```diff
@@ -14,16 +14,16 @@
...
-    var _paq = window._paq || [];
+    var _paq = window._paq = window._paq || [];
     /* tracker methods like "setCustomDimension" should be called before "trackPageView" */
     _paq.push(['trackPageView']);
     _paq.push(['enableLinkTracking']);
     (function() {
-      var u="https://mat.inrupt.com/";
-      _paq.push(['setTrackerUrl', u+'piwik.php']);
-      _paq.push(['setSiteId', '12']);
+      var u="https://inrupt.matomo.cloud/";
```
Vinnl, line 22: Hmm, if we're sending this to a third-party (matomo.cloud), even if they're relatively respectful of privacy, maybe we should be extra cautious not to send it if the user has explicitly opted out of tracking?

```suggestion
      if (navigator.doNotTrack && navigator.doNotTrack === "1") {
        return;
      }
      var u="https://inrupt.matomo.cloud/";
```

Vinnl (approved):

docs/api/_templates/layout.html
kay-kim: Thanks for the suggestion.  Will update all our matomo code.

PR #583, opened 2020-11-04
Rename acp_lowlevel_preview to acp_v1

Vinnl:
Somewhat of a compromise: it's no longer a constant and
in-your-face reminder that it's a preview, it doesn't look like a
regular module (because it's an object that contains all
ACP-related functions at once), and still allows us to iterate on
it without affecting the rest of our API's.


NSeydoux (approved):
I thought it was agreed that this was not v1 for the ACP API, but that's a tradeoff I can live with :)

Vinnl:
Yeah same here. In this case it's not in the shape of a semantic version though, so I think we're fine.

PR #589, opened 2020-11-05
Pin GH action dependencies to major versions

acoburn:
# New feature description

When referencing a [GitHub action dependency](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepsuses), it is possible to pin that to a particular commit, a particular release or just a major version. Pinning to a major version has several advantages:

1. It requires less work to keep the dependencies up-to-date (note that the [Annotations](https://github.com/inrupt/solid-client-js/actions/runs/347281220) section contains many errors due to outdated dependencies)
2. It means that, generally, your workflow will continue to function, provided that the maintainer of the dependency adheres to semantic versioning
3. It is recommended and common practice:

> Using the specific major action version allows you to receive critical fixes and security patches while still maintaining compatibility. It also assures that your workflow should still work.

## Checklist

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -20,7 +20,7 @@ jobs:
     steps:
     - name: Create GitHub Deployment
       id: create-deployment
-      uses: octokit/request-action@v2.x
+      uses: octokit/request-action@v2
       with:
         route: POST /repos/:repository/deployments
         repository: ${{ github.repository }}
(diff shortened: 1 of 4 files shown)


Vinnl (approved):
Ideally we'd have them pinned to exact versions and Dependabot would submit PRs with updates, but in lieu of that, major versions are a fine compromise between staying up-to-date and builds not randomly breaking, I think (especially considering that all the Actions we use are by GitHub themselves).

acoburn:
> Ideally we'd have them pinned to exact versions

I completely agree with this. The downside is that in practice, it is a maintenance headache to keep them current and there is no automation to check for updates.

Also, I take it that the CI tooling won't run when PRs come from external repositories.

Vinnl:
Actually, it seems that Dependabot _can_ do GitHub Actions: https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/configuration-options-for-dependency-updates#package-ecosystem

(Dependabot has access to the repository itself, so that shouldn't be a problem.)

I'll take a look at that.

Vinnl:
https://github.com/inrupt/solid-client-js/pull/590

acoburn:
> Actually, it seems that Dependabot can do GitHub Actions

ah, that's great! Feel free to close this PR in favor of that other update.

acoburn:
Closing in favor of #590

PR #601, opened 2020-11-09
Expose HTTP status on failed requests

### LINKED ISSUE #436: Expose status codes and error messages on errors
**Search terms you've used**

HTTP status codes, error message.

**Is your feature request related to a problem? Please describe.**

When calling a function that reads or writes data (e.g. `getSolidDataset` or `overwriteFile`), the server can return an error, which solid-client then throws and exposes in the error message. However, the error message cannot be relied upon to reliably get access to what in particular went wrong.

The error object that gets thrown in functions that make HTTP requests also exposes a property containing at least the status code and the status message, and possibly the entire Response object.

Manually parsing the error message for the expected status code.

**Additional context**

N/A

Vinnl:
# New feature description

This exposes the status code and status text of failed HTTP requests to the caller handling the thrown errors.

Fixes #436.

# Checklist

### DIFF
--- a/src/resource/nonRdfData.ts
+++ b/src/resource/nonRdfData.ts
@@ -83,7 +83,7 @@ export async function getFile(
   const response = await config.fetch(url, config.init);
   if (internal_isUnsuccessfulResponse(response)) {
     throw new FetchError(
-      `Fetching the File failed: ${response.status} ${response.statusText}.`,
+      `Fetching the File failed: \`${response.status}\` \`${response.statusText}\`.`,
       response
     );
   }
@@ -157,7 +157,7 @@ export async function deleteFile(
 
   if (internal_isUnsuccessfulResponse(response)) {
     throw new FetchError(
-      `Deleting the file at \`${url}\` failed: ${response.status} ${response.statusText}.`,
+      `Deleting the file at \`${url}\` failed: \`${response.status}\` \`${response.statusText}\`.`,
       response
     );
   }
@@ -189,7 +189,7 @@ export async function saveFileInContainer<FileExt extends File>(
 
   if (internal_isUnsuccessfulResponse(response)) {
     throw new FetchError(
-      `Saving the file in \`${folderUrl}\` failed: ${response.status} ${response.statusText}.`,
+      `Saving the file in \`${folderUrl}\` failed: \`${response.status}\` \`${response.statusText}\`.`,
       response
     );
   }
@@ -236,7 +236,7 @@ export async function overwriteFile(
 
   if (internal_isUnsuccessfulResponse(response)) {
     throw new FetchError(
-      `Overwriting the file at \`${fileUrlString}\` failed: ${response.status} ${response.statusText}.`,
+      `Overwriting the file at \`${fileUrlString}\` failed: \`${response.status}\` \`${response.statusText}\`.`,
       response
     );
   }
--- a/src/resource/nonRdfData.ts
+++ b/src/resource/nonRdfData.ts
@@ -38,6 +38,8 @@ import {
   getResourceInfo,
   isRawData,
   internal_cloneResource,
+  internal_isUnsuccessfulResponse,
+  FetchError,
 } from "./resource";
 import { type } from "rdf-namespaces/dist/dc";
 
@@ -79,9 +81,10 @@ export async function getFile(
   };
   const url = internal_toIriString(input);
   const response = await config.fetch(url, config.init);
-  if (!response.ok) {
-    throw new Error(
-      `Fetching the File failed: ${response.status} ${response.statusText}.`
+  if (internal_isUnsuccessfulResponse(response)) {
+    throw new FetchError(
+      `Fetching the File failed: ${response.status} ${response.statusText}.`,
+      response
     );
   }
   const resourceInfo = internal_parseResourceInfo(response);
@@ -152,9 +155,10 @@ export async function deleteFile(
     method: "DELETE",
   });
 
-  if (!response.ok) {
-    throw new Error(
-      `Deleting the file at \`${url}\` failed: ${response.status} ${response.statusText}.`
+  if (internal_isUnsuccessfulResponse(response)) {
+    throw new FetchError(
+      `Deleting the file at \`${url}\` failed: ${response.status} ${response.statusText}.`,
+      response
     );
   }
 }
@@ -183,9 +187,10 @@ export async function saveFileInContainer<FileExt extends File>(
   const folderUrlString = internal_toIriString(folderUrl);
   const response = await writeFile(folderUrlString, file, "POST", options);
 
-  if (!response.ok) {
-    throw new Error(
-      `Saving the file in \`${folderUrl}\` failed: ${response.status} ${response.statusText}.`
+  if (internal_isUnsuccessfulResponse(response)) {
+    throw new FetchError(
+      `Saving the file in \`${folderUrl}\` failed: ${response.status} ${response.statusText}.`,
+      response
     );
   }
 
@@ -229,9 +234,10 @@ export async function overwriteFile(
   const fileUrlString = internal_toIriString(fileUrl);
   const response = await writeFile(fileUrlString, file, "PUT", options);
 
-  if (!response.ok) {
-    throw new Error(
-      `Overwriting the file at \`${fileUrlString}\` failed: ${response.status} ${response.statusText}.`
+  if (internal_isUnsuccessfulResponse(response)) {
+    throw new FetchError(
+      `Overwriting the file at \`${fileUrlString}\` failed: ${response.status} ${response.statusText}.`,
+      response
     );
   }
 
--- a/src/resource/resource.ts
+++ b/src/resource/resource.ts
@@ -344,7 +344,7 @@ export function isPodOwner(
  */
 export class FetchError extends Error {
   public readonly statusCode: number;
-  public readonly statusText: string;
+  public readonly statusText?: string;
 
   constructor(message: string, errorResponse: Response & { ok: false }) {
     super(message);
--- a/src/resource/resource.ts
+++ b/src/resource/resource.ts
@@ -71,7 +71,7 @@ export async function getResourceInfo(
   const response = await config.fetch(url, { method: "HEAD" });
   if (internal_isUnsuccessfulResponse(response)) {
     throw new FetchError(
-      `Fetching the metadata of the Resource at \`${url}\` failed: ${response.status} ${response.statusText}.`,
+      `Fetching the metadata of the Resource at \`${url}\` failed: \`${response.status}\` \`${response.statusText}\`.`,
       response
     );
   }
--- a/src/resource/resource.ts
+++ b/src/resource/resource.ts
@@ -69,9 +69,10 @@ export async function getResourceInfo(
   };
 
   const response = await config.fetch(url, { method: "HEAD" });
-  if (!response.ok) {
-    throw new Error(
-      `Fetching the metadata of the Resource at \`${url}\` failed: ${response.status} ${response.statusText}.`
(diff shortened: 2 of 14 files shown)


NSeydoux (approved):

src/resource/nonRdfData.ts
```diff
@@ -152,9 +155,10 @@ export async function deleteFile(
     method: "DELETE",
   });
 
-  if (!response.ok) {
-    throw new Error(
-      `Deleting the file at \`${url}\` failed: ${response.status} ${response.statusText}.`
+  if (internal_isUnsuccessfulResponse(response)) {
+    throw new FetchError(
+      `Deleting the file at \`${url}\` failed: ${response.status} ${response.statusText}.`,
```
pmcb55, line 160:
```suggestion
      `Deleting the file at [${url}] failed: [${response.status}] [${response.statusText}].`,
```

src/resource/nonRdfData.ts
```diff
@@ -79,9 +81,10 @@ export async function getFile(
   };
   const url = internal_toIriString(input);
   const response = await config.fetch(url, config.init);
-  if (!response.ok) {
-    throw new Error(
-      `Fetching the File failed: ${response.status} ${response.statusText}.`
+  if (internal_isUnsuccessfulResponse(response)) {
+    throw new FetchError(
+      `Fetching the File failed: ${response.status} ${response.statusText}.`,
```
pmcb55, line 86: Just to be consistent across all our code (not just all our JavaScript code, but our Java and Python code too), I'd delimit the variable placeholders:
```suggestion
      `Fetching the File failed: [${response.status}] [${response.statusText}].`,
```

src/resource/nonRdfData.ts
```diff
@@ -183,9 +187,10 @@ export async function saveFileInContainer<FileExt extends File>(
   const folderUrlString = internal_toIriString(folderUrl);
   const response = await writeFile(folderUrlString, file, "POST", options);
 
-  if (!response.ok) {
-    throw new Error(
-      `Saving the file in \`${folderUrl}\` failed: ${response.status} ${response.statusText}.`
+  if (internal_isUnsuccessfulResponse(response)) {
+    throw new FetchError(
+      `Saving the file in \`${folderUrl}\` failed: ${response.status} ${response.statusText}.`,
```
pmcb55, line 192:
```suggestion
      `Saving the file in [${folderUrl}] failed: [${response.status}] [${response.statusText}].`,
```

src/resource/nonRdfData.ts
```diff
@@ -229,9 +234,10 @@ export async function overwriteFile(
   const fileUrlString = internal_toIriString(fileUrl);
   const response = await writeFile(fileUrlString, file, "PUT", options);
 
-  if (!response.ok) {
-    throw new Error(
-      `Overwriting the file at \`${fileUrlString}\` failed: ${response.status} ${response.statusText}.`
+  if (internal_isUnsuccessfulResponse(response)) {
+    throw new FetchError(
+      `Overwriting the file at \`${fileUrlString}\` failed: ${response.status} ${response.statusText}.`,
```
pmcb55, line 239:
```suggestion
      `Overwriting the file at [${fileUrlString}] failed: [${response.status}] [${response.statusText}].`,
```

src/resource/resource.ts
```diff
@@ -69,9 +69,10 @@ export async function getResourceInfo(
   };
 
   const response = await config.fetch(url, { method: "HEAD" });
-  if (!response.ok) {
-    throw new Error(
-      `Fetching the metadata of the Resource at \`${url}\` failed: ${response.status} ${response.statusText}.`
+  if (internal_isUnsuccessfulResponse(response)) {
+    throw new FetchError(
+      `Fetching the metadata of the Resource at \`${url}\` failed: ${response.status} ${response.statusText}.`,
```
pmcb55, line 74: Ditto

src/resource/nonRdfData.ts
pmcb55: Actually I guess you'll suggest updating all occurrences of the backtick delimiter with `[` and `]` later in one specific PR, but personally, since this is brand new code, I'd just start using that convention now, and fix the rest of the code in that later PR (i.e. I wouldn't see a particularly strong need to maintain consistency within just this codebase), and it just means less updates to make in that subsequent clean-up PR.

src/resource/resource.ts
```diff
@@ -338,6 +339,27 @@ export function isPodOwner(
   return podOwner === webId;
 }
 
+/**
+ * Extends the regular JavaScript error object with access to the status code and status message.
+ */
+export class FetchError extends Error {
```
pmcb55, line 345: I'd have expected this class to be in it's own source file, i.e. `FetchError.ts`.

src/resource/resource.ts
```diff
@@ -338,6 +339,27 @@ export function isPodOwner(
...
+  public readonly statusCode: number;
+  public readonly statusText: string;
```
pmcb55, line 347: I would have thought having the entire response object available here might be useful too (e.g. what if the server response contained a detailed explanation of the error (perhaps in RDF) in the body of the response?).

src/resource/nonRdfData.ts
pmcb55: It seems every instance of `new FetchError()` in this PR is duplicating the same text at the end of the error message string (i.e. `at [${url}] failed: [${response.status}] [${response.statusText}]`). So instead I'd just add a parameter to the  constructor to add it automatically, e.g. 
```
constructor(message: string, errorResponse: Response & { ok: false }, appendDetails:boolean = true) {
  super(`${message}${appendDetails  ? ` Failed at URL [${url}]: [${response.status}] [${response.statusText}].` : ''}`);
```
Messages would need to change accordingly, i.e. this would just become `throw new FetchError("Deleting file.", response);`.

src/resource/nonRdfData.ts
Vinnl: I did not touch the actual error message, but I added 63523897f74308b0e8a6968d353ef907f304a6b0 to add them now. I did use the delimiters we use elsewhere in the code.

src/resource/resource.ts
Vinnl: Just because it's a class? We don't use separate files for all the functions related to Resources either?

src/resource/resource.ts
Vinnl: Yeah it's listed in the ticket as something to consider, but I preferred to leave it out for now because I'd like to avoid people manually parsing server responses, which will then likely only work with the server they happen to be testing against. If there's something standardised in the responses, we should expose it in a way that will work for every Solid server. (And e.g. the Solid spec should specify that a detail explanation of the error should be included in RDF using some vocabulary.)

Vinnl:
@NSeydoux @pmcb55 Sorry, I submitted this PR too early and converted it to a draft, because I'd forgotten two things that needed to be added as well.

Could you also review d3fac7e1a7d3b8ab76b4662d0e846965ec1cf85c and fcbbda2bfe886b9eef310c03bdda900b5c956b8b?

src/resource/nonRdfData.ts
pmcb55: | I did use the delimiters we use elsewhere in the code.

But we've all agreed (including Jack, as tech lead for all front-end development at Inrupt) to use '[' and ']', and not backticks as delimiters, right? So why continue to use backticks on any new code we write (i.e. when someone is going to have to revert all of these to use '[' and ']')...?

src/resource/resource.ts
pmcb55: No, not **_just_** because it's a class, but because it seems to be a very general, generic, standalone bit of code that I assume could be easily reused across much of this codebase, or in fact any Inrupt codebase. Having it reside within the `resource.ts` source file would, I think, makes any developer think it is somehow be related to 'resources'. In fact, looking at the `resource.ts` file itself now, it seems to be a kinda grab-bag of utility functions related to resources (so I'd probably rename it `resourceUtil.ts`, if that is it's actual intent), but that seems to reinforce my opinion of moving this Fetch-specific Error class into it's own source file, as I wouldn't see it merely as a util class for use with resources, I'd see it as a general-purpose Error class for use with fetch, any fetch, and therefore seeing it reside in a file named `FetchError.ts` would be my expectation.

src/resource/resource.ts
pmcb55: Hmmm... Ok. Personally I would still have err'ed on the side of providing the full response object. I still consider this entire library fairly low-level, so I'd favor flexibility and access over protecting developers from foot-guns (as that's what higher-level abstractions try to do). But I'm Ok to leave this for now and let's see what happens...

src/resource/nonRdfData.ts
Vinnl: For consistency - we work through the whole codebase in one fell swoop in a future ticket to update the delimiters everywhere, rather than use one type in some places, and another type in other places.

src/resource/resource.ts
Vinnl: Currently, the organisation is as follows:

```
src
├─  resource/ # Contains everything related to fetching data
| ├─ solidDataset.ts # Specific to fetching SolidDatasets
| ├─ nonRdfData.ts # Specific to fetching non-RDF data
| ├─ resource.ts # Generic data fetching code
```

By that organisation, it seems logical to have errors arising from fetching data of any kind inside `/resource/resource.ts`?

src/resource/resource.ts
Vinnl: Yeah, definitely something we can reconsider when we have higher-level stuff that people use, instead of solid-client directly.

NSeydoux:
Oh nice, I didn't know `Response` provided a default status text, in this case no problem :)

src/resource/nonRdfData.ts
ajacksified: I'd prefer internal consistency, which we can upgrade all at once.

pmcb55: Okey dokey - but I reckon my comments absolve me from being the lucky dev chosen to implement that 'switch to brackets everywhere' PR :) !
(BTW, I notice now that the GitHub 'Suggestions' button is greyed out now, and the helpful tooltip explains why (comment is outdated). But what does having no 'Suggestions' button at all mean (or is the greyed-out-and-tooltip thing a brand new update, like in the past 10 mins!)...?):
image

Vinnl: Haha sure, feel free to create the ticket and then mention those comments here :) 

I think adding a suggestion only works when commenting on a change that's still in the PR. (Which is really annoying if you want to comment on how something affects some other piece of code, but alas...)

pmcb55: Sure, I had already created the ticket  (SDK-1391), but I've updated it to point here too :).

src/resource/resource.ts
pmcb55: I guess it's subjective - i.e. it's not really a big deal (so I'm happy to leave this one), but personally I'd rename `resource.ts` to `resourceUtil.js` (as that seems to better reflect it's intent), and provide the `FetchError` class in either the existing `/fetcher.ts` file, or it's own file in a project `Util` folder (in fact, perhaps even a generic Inrupt JavaScript Util project (but I'm not sure if we have enough common utility JavaScript code across the various Inrupt teams yet to justify that step, but consistent error handling would be a very good candidate for starting...)).

Vinnl: Haha OK, that's indeed my definitely subjective and unreasonably strong dislike of anything called "util" or that is the kind of grab-bag of hardly-related too-generic functionality that it usually refers to, which I will resist for longer than a reasonable person would consider worth it :P

(But of course, if this does turn out to be something that we use beyond the resource-related functionality, we can always move it elsewhere.)

PR #605, opened 2020-11-11
Update Makefile and index.rst for docs

kay-kim:
Update makefile's target [migrate] to copy over everything in api/ directory - rather than keep up with directories under api.

### DIFF
--- a/docs/api/Makefile
+++ b/docs/api/Makefile
@@ -33,8 +33,10 @@ migrate: clean
    # in case someone forgets to backslash after build/
 
 	cp -R $(SOURCEDIR)/index.rst $(SOURCECOPYDIR)
-	cp -R $(SOURCEDIR)/api $(SOURCECOPYDIR)
-
+	cp -R $(SOURCEDIR)/api/modules/ $(SOURCECOPYDIR)/modules/
+	cp -R $(SOURCEDIR)/api/classes/ $(SOURCECOPYDIR)/classes/
+	cp -R $(SOURCEDIR)/api/README.md $(SOURCECOPYDIR)/
+	
 html: Makefile migrate
 
 # To clean up refs in markdown, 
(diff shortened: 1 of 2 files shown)


docs/api/Makefile
```diff
@@ -43,9 +42,9 @@ html: Makefile migrate
 # Also, on macOS, need -i '' but not for Linux
 
 ifdef IS_MAC
-	find $(SOURCECOPYDIR)/modules/ -type f -name "*.md" -exec sed -i '' -e 's?\.md#?\.html#?g' {} \;
+	find $(SOURCECOPYDIR)api/modules/ -type f -name "*.md" -exec sed -i '' -e 's?\.md#?\.html#?g' {} \;
```
Vinnl, line 45: Given that it used to say `/modules`, shouldn't we prepend a slash to `api/modules` as well? (Same question for the `else` leg.)

```suggestion
	find $(SOURCECOPYDIR)/api/modules/ -type f -name "*.md" -exec sed -i '' -e 's?\.md#?\.html#?g' {} \;
```

Vinnl (approved):
I'm not sure if I understand what's going on/why this is needed, but it looks fine to me (assuming the answer to the question I asked is "no" :) ).

docs/api/Makefile
kay-kim: Oh - SOURCECOPYDIR already ends with a / and on a mac, the extra slash of build/source//api  causes it to error.

Of course, I now see that I'm only changing in the api modules/ files and if files in the classes/ directory link to other files, this sed won't change it.

kay-kim: But - adding a "." to the "find  . $(SOURCECOPYDIR)"  should allows us to build with the extra "/" .  Am making the change now.

kay-kim: Oh - actually - just realized I clobbered logic I shouldn't have during the copy phase.

PR #612, opened 2020-11-16
Set all serialised date parts at once

Vinnl:
Previously, we set each date part individually (i.e. first the
year, then the month, etc.). However, if any of these parts result
in an invalid/unrepresentable date (e.g. because the year is
earlier than the earliest year that can be represented by
JavaScript's Date object), the end result will also be invalid,
even though the parts we set later would make them valid again
(e.g. because the timezone pushes the date back into the range that
is representable by Date).

To fix this, this commit sets all Date parts at the same time.

This bug was discovered by our property-based tests, which tested more edge cases with fast-check's new version. Hence, this PR should make #609 mergeable.

### DIFF
--- a/src/datatypes.test.ts
+++ b/src/datatypes.test.ts
@@ -224,6 +224,32 @@ describe("deserializeDatetime", () => {
     expect(deserializeDatetime("1990-11-12T10:00:00+14:00")).toEqual(
       expectedDateWithMaxPositiveTimezone
     );
+
+    const expectedEarliestRepresentableDate = new Date(-8640000000000000);
+    expect(deserializeDatetime("-271821-04-20T00:00:00.000Z")).toEqual(
+      expectedEarliestRepresentableDate
+    );
+    // Same date, earlier timezone
+    expect(deserializeDatetime("-271821-04-20T01:00:00.000-01:00")).toEqual(
+      expectedEarliestRepresentableDate
+    );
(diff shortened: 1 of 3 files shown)


src/datatypes.test.ts
```diff
@@ -224,6 +224,44 @@ describe("deserializeDatetime", () => {
     expect(deserializeDatetime("1990-11-12T10:00:00+14:00")).toEqual(
       expectedDateWithMaxPositiveTimezone
     );
+
+    const expectedEarliestRepresentableDate = new Date(-8640000000000000);
```
pmcb55, line 228: I'd just make this (and the positive version) file-level constants with descriptive names - i.e. the same values, plus and minus, are used in a couple of different contexts (i.e. in Earliest, Latest, Earlier Timezone, in Later Timezone), e.g.
```
const EARLIEST_POSSIBLE_VALID_DATE = new Date(-8640000000000000);
const LATEST_POSSIBLE_VALID_DATE = new Date(8640000000000000);
```

pmcb55 (approved):
LGTM

src/datatypes.test.ts
Vinnl: Huh, good point, those shouldn't have been different variables - the point of those tests is to represent the _same_ value with different serialisations. Fixed now.

PR #623, opened 2020-11-18
Support ES modules in Node 14 and 12

Vinnl:
# New feature description

This is a somewhat big PR that, as you can see by the number of commits, does quite a number of things, all in service of being able to load our modules as ES modules in Node (and, incidentally, to explicitly define which modules can be imported in Webpack 5 as well).

Specifically:

- CI checks now run in Node 14 too.
- When we import CommonJS modules (which all our dependencies currently still are, in Node), we no longer use named imports, which are not supported by CommonJS. Except for...
- N3, which annoying exports ES modules with named imports to browsers, but CommonJS modules with only a default export for Node. This means that `import { Parser } from "n3"` would break in Node, but `import n3 from "n3"` would break in Webpack. I worked around that via dynamic import - a comment in the code and the associated commit message explain this in more detail.
- Added a fourth build output (after CommonJS modules, `.js` ES modules for bundlers, and a UMD build for people wanting to include a single script): `.mjs` ES modules for Node.
- Add an export map detailing which submodules we export. I've kept the top-level export to avoid introducing a breaking change for people upgrading their Node/Webpack versions. (And since the filenames matter for this, I renamed the file containing File-related API's to `file.ts`.)
- I added minimal projects that import functions from solid-client in different environments. After publishing a preview release, that release is added to those projects in CI and then run/built, to see if that breaks. I hope that they cover the most common usages.
- I added the [package-check](https://docs.skypack.dev/package-authors/package-checks) to check for best practices, such as publishing ES modules.

# Checklist


pmcb55 (approved):
LGTM (having the demo of this today really helped set the context for reviewing this, thankfully!).

Vinnl:
Great. I also added bad8e5007cc4b1b49ed4718c50cfcd39b0f7619c, which moves everything that's prefixed with `internal_` into its own modules (suffixed `.internal.ts`), so that people on Node 12+ and Webpack 5 will not be able to import them. (And so we can automatically exclude them from our docs, so this would've been useful long ago.)

PR #627, opened 2020-11-19
Run end-to-end tests against the published package

Vinnl:
This was actually easier than I thought. What I did was simply run a search+replace on the Node end-to-end test files to replace the imports from `../index` with imports from `@inrupt/solid-client`, and then force-installed the just published package as a dependency of this package - even though it's the same package, this does not seem to cause a problem.

I verified that this does indeed run the published code by running an `rm -rf node_modules/@inrupt/solid-client` before running the tests, after which they failed, as expected.


pmcb55 (approved):
LGTM (very nice idea, just so elegantly simple - great job :) !).

PR #628, opened 2020-11-19
Throw error when not passed a Thing

Vinnl:
Although it's not really doable to verify all the types we encounter at runtime, passing the result of `getThing()` to one of our Thing-related functions without doing a null check is probably common enough that explicitly checking and adding a helpful error message is probably worth the effort. This was triggered by @pmcb55 encountering the cryptic runtime error `thing is not iterable` when working without TS for a demo app.

### DIFF
--- a/src/thing/add.ts
+++ b/src/thing/add.ts
@@ -21,7 +21,11 @@
 
 import { Literal, NamedNode, Quad_Object } from "rdf-js";
 import { Thing, UrlString, Url } from "../interfaces";
-import { internal_cloneThing, internal_toNode } from "./thing.internal";
+import {
+  internal_cloneThing,
+  internal_toNode,
+  internal_throwIfNotThing,
+} from "./thing.internal";
 import {
   asNamedNode,
   serializeBoolean,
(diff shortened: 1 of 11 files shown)


src/thing/add.ts
```diff
@@ -46,11 +50,12 @@ import { DataFactory } from "../rdfjs";
  * @param url URL to add to `thing` for the given `property`.
  * @returns A new Thing equal to the input Thing with the given value added for the given Property.
  */
-export const addUrl: AddOfType<Url | UrlString | Thing> = (
+export const addUrl: AddOfType<Url | UrlString | Thing> = function addUrl(
```
ajacksified, line 53: Why not `export function addUrl(`?

Vinnl: This allows us to reuse the `AddOfType` type definition, to make sure the signature stays aligned with the other `add<type>` functions.

pmcb55: Not sure I follow - I think Jack meant:
```
export function addUrl(...): AddOfType<Url | UrlString | Thing> {
```
Isn't that still reusing the `AddOfType` type definition?
My concern on this code change is the duplication of the function name itself (i.e. `addUrl` in this case), and that if a developer used the 'Refactor -> Rename' option in an IDE, the IDE might only rename one of those two, leading to untold confusion (I don't know that an IDE would do that, but I suspect it might).
I think(!) you're only doing this function name duplication so that the exception message can include the function name (which would indeed be a good (but not **_essential_**) thing to include in the exception message), but is there not a cleaner way to (reliably) do that in JavaScript without duplicating the function name like it is here?
(I know there are very general problems with minifiers renaming functions too, so not sure how that might affect all this too...?)

Vinnl: That suggestion would define the return type, not the type of the function.

You raise a good point though: the duplication of function names has a risk of getting outdated. I could remove the use of `AddOfType`, since the functions getting misaligned isn't a major risk anyway, but looking at this with fresh eyes I think my idea of including the function name in the message is misguided anyway: that's what the stack trace is for, and that will even tell you the _actual_ function you called. So I'll just remove that, and resubmit.

PR #634, opened 2020-11-23
Throw error when not passed a Thing

Vinnl:
Recreation of #628 incorporating [the feedback](https://github.com/inrupt/solid-client-js/pull/628#discussion_r527039402) of Jack and Pat that the assignment of a named function was odd (which it was), and removing the manual semi-recreation of the stack trace, which also was pretty weird. It now simply throws an error in the functions operating on a Thing if not passed a Thing.

### DIFF
--- a/src/thing/thing.internal.ts
+++ b/src/thing/thing.internal.ts
@@ -40,7 +40,7 @@ import {
   LocalNode,
   ThingPersisted,
 } from "../interfaces";
-import { isThingLocal, asUrl, isThing } from "./thing";
+import { isThingLocal, asUrl, isThing, ThingExpectedError } from "./thing";
 
 /** @hidden For internal use only. */
 export function internal_getReadableValue(value: Quad_Object): string {
@@ -168,6 +168,6 @@ export function internal_filterThing<T extends Thing>(
  */
 export function internal_throwIfNotThing(thing: Thing): void {
   if (!isThing(thing)) {
-    throw new Error(`Expected a Thing, but received: \`${thing}\`.`);
+    throw new ThingExpectedError(thing);
   }
 }
--- a/src/thing/thing.internal.ts
+++ b/src/thing/thing.internal.ts
@@ -40,7 +40,7 @@ import {
   LocalNode,
   ThingPersisted,
 } from "../interfaces";
-import { isThingLocal, asUrl } from "./thing";
+import { isThingLocal, asUrl, isThing } from "./thing";
 
 /** @hidden For internal use only. */
 export function internal_getReadableValue(value: Quad_Object): string {
@@ -161,4 +161,13 @@ export function internal_filterThing<T extends Thing>(
   }
   (filtered as ThingPersisted).internal_url = (thing as ThingPersisted).internal_url;
   return filtered as T;
+}
+
+/**
+ * @hidden
+ */
+export function internal_throwIfNotThing(thing: Thing): void {
+  if (!isThing(thing)) {
+    throw new Error(`Expected a Thing, but received: \`${thing}\`.`);
+  }
 }
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -148,6 +148,10 @@ import {
   mockThingFrom,
   addMockResourceAclTo,
   addMockFallbackAclTo,
+  // Error classes:
+  SolidClientError,
+  FetchError,
+  ThingExpectedError,
   // Preview API's exported for early adopters:
   acp_v1,
   // Deprecated functions still exported for backwards compatibility:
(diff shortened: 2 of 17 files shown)


NSeydoux (approved):

src/thing/thing.internal.ts
```diff
@@ -162,3 +162,12 @@ export function internal_filterThing<T extends Thing>(
   (filtered as ThingPersisted).internal_url = (thing as ThingPersisted).internal_url;
   return filtered as T;
 }
+
+/**
+ * @hidden
+ */
+export function internal_throwIfNotThing(thing: Thing): void {
+  if (!isThing(thing)) {
+    throw new Error(`Expected a Thing, but received: \`${thing}\`.`);
```
ajacksified, line 171: Maybe a good spot to start implementing typed errors?

Vinnl: What kind of type would you expect here, and what would it be used for?

ajacksified: Even more specific, although I'm not against inheritance - `ThingExpectedError(value)`, which could inherit from a `SolidClientError` so that client code can do things based on error type rather than checking the string.

Vinnl: OK, so just a new class per error we throw. As long as they all inherit from SolidClientError (which is nothing more than a non-modifying wrapper) and not from each other (to avoid [MurderRobotDogs](https://medium.com/humans-create-software/composition-over-inheritance-cb6f88070205) I'm fine with that - I'll add another commit.

Vinnl: Added: bf706f538b9b9ae55e0e7be6dfdcaa3d55873567

ajacksified (approved):

src/thing/thing.internal.ts
pmcb55: @ajacksified @Vinnl Arrgghhh! I wouldn't have done that - i.e. I really dislike seeing a proliferation of error classes (in Java or JavaScript) that do literally nothing but extend `Error` (or `RuntimeException`) as a means of allowing '_other code do things based on the error type_'. The intent is perfectly fine, but not using the type system to do so at that level of granularity. Instead I prefer just a single generic `InruptError` class (so it's still distinguishable from standard `Error` instances at that level of granularity), but that also allows an error identifier in it's constructor. That way we can distinguish not by Error **_type_** (although we still can at the `Error` vs `InruptError` level), but by error **_identifier_**, which I think is far more appropriate (since identifiers can be IRIs, thereby globally unique and dereferenceable).
We've started down that road already in `solid-client-authn-js` here: https://github.com/inrupt/solid-client-authn-js/blob/master/packages/core/src/errors/InruptError.ts (there's clean up to do on the constructor by moving the HTTP Response-related params out, and there's lot more ideas like allowing nested Errors, and storing parameters (to allow client code do things based on error ID **_and_** parameter values, for instance), but the basic idea is there already).
**_NOTE_**: I already demo'ed this `InruptError` class in the team demos just last week - so maybe those demo sessions aren't as useful as maybe we thought... :) !

ajacksified: It is a very common pattern in javascript (and most other languages) to use error types. `error instanceof SpecificError` is simpler than `error.someArbitraryPropertyIdSpecialToInrupt === someIriHardcodedSomewhere`. There's no reason each class couldn't define an identifier- but we should follow common javascript conventions in a javascript library.

As an aside, multiple types of errors also prevents a proliferation of nullable constructor arguments (as you can see with the _four_ nullable or defaulted arguments in InruptError, which will only grow as the number of error types increase.)

Vinnl: Don't worry Pat, I was planning to cut a release today but can hold off on that for a bit.

That said, I agree with Jack: we can certainly add an IRI to every error (though by all means, give me a concrete IRI that we can use today and that will never change - I wouldn't dare to come up with one), but I don't see why that would need to be separate. In fact, we could define every error class we have to explicitly carry that IRI, always.

And yes, I've only taken a small glance at that InruptError class, but I didn't like it very much for exactly the reason Jack mentioned. It feels overly generic, and I don't see any clear advantages that would make that worth it.

pmcb55: :) Oh, the deja vu (i.e. the number of times I heard the same pushback, with lightbulbs only going off later). Let's take this offline, as I didn't realise there was that much of a disconnect.

Vinnl: I'd much prefer to see it written down though, where I can refer back to it later and think it through leisurely. I've found that's by far the best way for me to internalise best practices.

pmcb55: @Vinnl - oh yeah, sure thing. When I said 'offline', I just meant 'off-this-line'. I am writing it up and will share when written, i.e. 'soon'...

PR #638, opened 2020-11-24
Document when intermediary Containers are created

Vinnl:


NSeydoux (approved):

src/resource/file.ts
```diff
@@ -170,6 +170,10 @@ type SaveFileOptions = GetFileOptions & {
  * Saves a file in a folder associated with the given URL. The final filename may or may
  * not be the given `slug`.
  *
+ * The Container at the given URL should already exist; if it does not, you can initialise it first
```
kay-kim, line 173: Q: Should we mention that if it doesn't, the function throws an error?

Vinnl: Yeah good point, done in b71c158cd941eb37215464f04b500f5701046bb0.

PR #640, opened 2020-11-25
Streamline CD

Vinnl:
For some reason I'd set up the Webpack and Parcel import verification jobs to run _after_ the Node ones, and to exclude other Node versions in a rather cumbersome way. This makes them separate jobs that only run in Node 14.

Additionally, all jobs that rely on the preview npm package being published now wait for 60s before installing it. Hopefully that is long enough for npm to make the package available, because we've been having intermittent failures in that regard, where the package sometimes couldn't be installed immediately after publication.

I also added a timestamp to the published version, so that re-running the job doesn't try to overwrite the previously published package. (For some reason GitHub Actions doesn't have an env var that indicates the job run.)


pmcb55 (approved):
LGTM - but did you consider trying to check the exit code from the `npm install --force ...` calls, and then maybe retrying after a short timeout (and then maybe checking and retying a 2nd time after a much longer timeout)...? If we could do that, then we might be able to report on each manual retry attempt we make (e.g. sending a static POST message to some `https://inrupt.com/buildProcessMonitoringServer/` endpoint (e.g. `Retrying to install [${pacakge}], after timeout of {${timeout}] seconds... result [{success | failed}]`, so we could watch what's going on over time...)?

Vinnl:
@pmcb55 That's a good idea, I'll look into that tomorrow.

Vinnl:
Unfortunately my Bash-fu apparently isn't good enough to make that work in a reasonable time frame, so I'll keep this as-is for now, and maybe we can look into that later if the delay becomes problematic.

PR #642, opened 2020-11-26
Remove access control indirection

Vinnl:
# New feature description

This adds ACP functions to add Policy URLs to Resources with ACRs directly, rather than having to manually check whether Access Controls already exist and adding one if they do not. This is because individual Access Controls don't do anything: whether there are ten or just one, all their statements will be concatenated together to determine the applicable policies.

The old API's remain available unchanged on the exported `acp_v1` object; the updated API's are available on the new `acp_v2` object.

# Checklist

- [x] Relevant documentation, if any, has been written/updated. (Prose docs will be updated in a separate PR on the docs repo.)


pmcb55 (approved):
LGTM

src/acp/v1.ts
```diff
@@ -0,0 +1,119 @@
...
+  control: Control
+): ResourceExt {
+  const acr = internal_getAcr(withAccessControlResource);
+  const updatedAcr = removeThing(acr, control);
+  const updatedResource = internal_setAcr(
+    withAccessControlResource,
+    updatedAcr
+  );
+  return updatedResource;
+}
```
NSeydoux, line 119: Why isn't this function in `control.internal.ts` ?

NSeydoux (approved):
Looks good to me, only a minor question but that's just out of curiosity :).

src/acp/v1.ts
Vinnl: It's the only one the newer API's do not use internally. (I was doubting whether I wanted to re-use the other ones in the first place, since they're so trivial and their names overlap, or if I should just move all old functions into `v1.ts`. I'm still not sure about that :P )

PR #643, opened 2020-11-26
Export ACL-related API's from ACL module

Vinnl:
This just moves code around; no functionality is changed.

With the export map now making the module in which an API is
defined relevant to the consumer, it makes sense to keep all our
ACL-related API's in the ACL module. (This isn't a breaking change yet, because we haven't yet pushed a release with the exports map - it would be afterwards, though.)

The only remnant is some references to ACLs in the Resource
metadata, but that can be removed once the ACL-related API's have
been updated to look at the generic Linked Resources metadata.


pmcb55 (approved):
LGTM (kinda hard to see what's going on exactly ('cos the PR is so big), but it does appear to just be moving files around, so I guess that's fine!).

PR #647, opened 2020-11-30
Make e2e tests work again by using the regular Pod environment

Vinnl:
Sign-up there works again, and it's expected to be stable, whereas
the dev environment might be undergoing e.g. performance tests, causing instability in our pipelines.


pmcb55 (approved):
LGTM

PR #675, opened 2020-12-15
Add authenticated end-to-end tests in Node

Vinnl:
Now that we have an auth client for Node, we can do authenticated end-to-end tests in Node, which are far less brittle than the browser-based ones!

This PR adds end-to-end tests that exercise many of the same code paths as the existing ones, but now against ESS. NSS can follow as soon as https://github.com/solid/node-solid-server/issues/1533 is fixed and deployed somewhere. It also prevents errors when developers do not specify a content type, because the end-to-end tests ran into that error.

The tests run against both pod.inrupt.com and pod-compat.inrupt.com, with WAC tests only running against the latter.

After this is merged, the secrets starting with `TESTCAFE_ESS_PROD` can be removed from the repository's secrets - they've been renamed to `E2E_TEST_ESS_PROD`.

# Checklist


src/resource/file.ts
```diff
@@ -305,7 +319,15 @@ async function writeFile(
...
-  headers["Content-Type"] = file.type;
+  const fileType =
+    typeof file === "object" &&
+    file !== null &&
+    typeof (file as Blob).type === "string" &&
+    (file as Blob).type.length > 0
+      ? (file as Blob).type
+      : undefined;
+  headers["Content-Type"] =
+    options.contentType ?? fileType ?? "application/octet-stream";
```
NSeydoux, line 330: This (rather intricate) piece of logic is duplicated. Should it be isolated in its own function ?

src/e2e-node/e2e.test.ts
```diff
@@ -291,3 +299,242 @@ describe.each([
...
+        fetch: session.fetch,
+      });
+
+      const firstSavedDataset = await getSolidDataset(datasetUrl, {
+        fetch: session.fetch,
+      });
+      const firstSavedThing = getThing(
+        firstSavedDataset,
+        datasetUrl + "#e2e-test-thing"
+      )!;
```
NSeydoux, line 377: I never noticed this type of operators, and just discovered their existence ^^. That looks convenient !

NSeydoux (approved):
It all looks good, but I'm cutting a release of `solid-clent-authn-node` which supports the sessions WebID right now, so maybe this could wait on the release to be merged, so that some of the workarounds don't have to be included ?

src/resource/file.ts
Vinnl: Hmm yeah good point, it started out rather innocently, but kinda grew into this monster without me noticing. I'll refactor it.

src/e2e-node/e2e.test.ts
Vinnl: I'd be a bit reluctant with it though, it's basically a type assertion, i.e. telling TypeScript that you know a value is not going to be `null` - an actual null check is of course safer. But a runtime error here is no problem, since we want tests to fail if this doesn't hold anyway.

src/resource/file.ts
Vinnl: ecb193fc7cdba8d0935e1a08adc58fc93f6a0f62

PR #677, opened 2020-12-16
WIP: Attempt to fix CodeSandbox

Vinnl:
Before this, actually running CodeSandboxes resulted in the
following error:

> Could not fetch dependencies, please try again in a couple
> seconds: Could not fetch
> https://cdn.jsdelivr.net/gh/../..//package.json

I thought that was just CodeSandbox being unstable, but I think the cause might be on our side. This PR is to get CodeSandbox to build this branch again to see whether that fixes it - it does not need review yet.

### DIFF
--- a/.codesandbox/sandbox/package.json
+++ b/.codesandbox/sandbox/package.json
@@ -8,7 +8,7 @@
     "build": "parcel build index.html"
   },
   "dependencies": {
-    "@inrupt/solid-client": "../../",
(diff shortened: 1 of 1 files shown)


Vinnl:
Oh never mind, PRs are not the problem (that's when CodeSandbox provides its own build) - it's when using the repo directly as a CodeSandbox, which I can test without a PR.

PR #682, opened 2020-12-17
Migrate ACP end-to-end tests to Node

Vinnl:
Now that we can send authenticated requests in Node, we can
manipulate Access Control Policies there. Since browser-based tests
are relatively brittle [1] and pretty involved (the interaction
between the test script and the in-browser code, i.e. the
ClientFunctions), we limit those to a small smoke test that
verifies that we can read and manipulate data while logged in.

[1] https://testing.googleblog.com/2017/04/where-do-our-flaky-tests-come-from.html

### DIFF
--- a/src/e2e-node/e2e.test.ts
+++ b/src/e2e-node/e2e.test.ts
@@ -560,7 +560,10 @@ describe.each(serversUnderTest)(
         let publicReadPolicy = acp.createPolicy(
           policyResourceUrl + "#policy-publicRead"
         );
-        publicReadPolicy = acp.addRequiredRuleUrl(publicReadPolicy, publicRule);
+        // Note: we should think of a better name for "optional", as this isn't really optional.
+        //       At least one "optional" rule should apply, and since this is the only rule for this
+        //       policy, it will in practice be required.
+        publicReadPolicy = acp.addOptionalRuleUrl(publicReadPolicy, publicRule);
         publicReadPolicy = acp.setAllowModes(publicReadPolicy, {
           read: true,
           append: false,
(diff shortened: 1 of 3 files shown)


src/e2e-node/e2e.test.ts
```diff
@@ -536,5 +538,159 @@ describe.each(serversUnderTest)(
...
+        policyResourceUrl: UrlString,
+        session: Session
+      ) {
+        let publicRule = acp.createRule(policyResourceUrl + "#rule-public");
+        publicRule = acp.setPublic(publicRule, true);
+
+        let publicReadPolicy = acp.createPolicy(
+          policyResourceUrl + "#policy-publicRead"
+        );
+        publicReadPolicy = acp.addRequiredRuleUrl(publicReadPolicy, publicRule);
```
NSeydoux, line 562: would it make sense if the rule containing the public agent was an optional one (`acp:any`) instead of a required one (`acp:all`) ? I think in this specific context it would not change anything, so I don't think it's necessary to change the current implementation, it's mostly a question to check that I understand what's going on ^^

NSeydoux (approved):

src/e2e-node/e2e.test.ts
Vinnl: Yeah that's one problem I have with the current API: I called it `Optional`, but it's not optional at all - at least one of those rules should be valid.

It would be good though if both variants would be checked, so I'll change one of the required rules to be "optional".

Vinnl: 7e2b9f5ffae53f5ba16e11aba6852f0fb6aec3ec

PR #683, opened 2020-12-17
Automatically retry e2e tests on MacOS in CI

Vinnl:
Hopefully this assuages some of our CI woes... It commonly fails with network errors (logged as "Failed to find a DNS-record for the resource at"), which appears to be particular to the combination of GitHub Actions and TestCafe, and additionally most often happens on MacOS. Since all that is out of our control, automatically retrying seems best.


NSeydoux (approved):

NSeydoux:
It seems particularly in the air right now to run things in quarantine mode

PR #685, opened 2020-12-18
Add response to FetchError

ajacksified:
# Add HTTP response to fetchError

Allow more in-depth debugging of request failures. Especially handy for usage outside of the browser.

# Checklist


src/resource/resource.ts
```diff
@@ -181,12 +181,14 @@ export function isPodOwner(
  * @since 1.2.0
  */
 export class FetchError extends SolidClientError {
+  public readonly response: Response & { ok: false };
   public readonly statusCode: number;
   public readonly statusText?: string;
```
NSeydoux, line 186: Isn't it redundant to have both ? should we consider making `statusCode` and `statusText` into accessors for the `response` properties for retro-compatibility, and eventually deprecate these ?

NSeydoux (approved):

src/resource/resource.ts
ajacksified: Yes, they should be deprecated -I just didn't want to create a breaking change with this. I can swap them to getters, though.

pmcb55 (approved):

src/resource/resource.ts
```diff
@@ -181,12 +181,18 @@ export function isPodOwner(
  * @since 1.2.0
  */
 export class FetchError extends SolidClientError {
-  public readonly statusCode: number;
-  public readonly statusText?: string;
```
Vinnl, line 185: Removing these makes this a breaking change :(

Vinnl (changes requested):

src/resource/resource.ts
Vinnl: Whoops, never mind - didn't realise they were getters, not methods :)

Vinnl: Although then - shouldn't `reponse` be a getter as well?

src/resource/mock.test.ts
```diff
@@ -21,6 +21,9 @@
 
 import { describe, it, expect } from "@jest/globals";
 
+import * as crossFetch from "cross-fetch";
+const { Response } = crossFetch;
```
Vinnl, line 25: As a nitpick and/or for my education: any reason this isn't just

```suggestion
import { Response } from "cross-fetch";
```

?

Vinnl (approved):

src/resource/resource.ts
Vinnl: Ah right, it's just to read the property from `reponse`, not to actually make it readonly. So much for my conversation with myself ;)

Anyway, if you do want to deprecate these in favour of reading them from the response object, the `@deprecated` TSDoc annotation will mark it as such (preferably combined with a comment describing its replacement).

src/resource/mock.test.ts
ajacksified: Copied it from another file to be consistent 🤷‍♂️

src/resource/resource.ts
ajacksified: Agreed - though if we do deprecate it, I'd prefer to mark it deprecated at the next minor release. It's a handy enough shortcut that maybe it's worth discussion.

src/resource/mock.test.ts
Vinnl: Ah, I see: https://github.com/inrupt/solid-client-js/commit/658d5901e7b1fb4f5ebb1fae1c889ce521a382fd

For posterity: that was because cross-fetch is a CommonJS module, and thus named imports will break when called from an ES module Node project. Not relevant here, but not problematic either - just good to know :)

PR #693, opened 2020-12-23
Clarify when to use the *InContainer() functions

Vinnl:
A developer was unclear about whether to use `saveSolidDatasetInContainer` rather than just `saveSolidDatasetAt`; this PR adds some clarification on why you'd use the former, ever.

### DIFF
--- a/src/resource/file.ts
+++ b/src/resource/file.ts
@@ -139,6 +139,16 @@ type SaveFileOptions = WriteFileOptions & {
  * be rejected. You can initialise it first using [[createContainerAt]], or directly save the file
  * at the desired location using [[overwriteFile]].
  *
+ * This function is primarily useful if the current user does not have access to change existing files in
+ * a Container, but is allowed to add new files; in other words, they have Append, but not Write
+ * access to a Container. This is useful in situations where someone wants to allow others to,
+ * for example, send notifications to their Pod, but not to view or delete existing notifications.
+ * You can pass a suggestion for the new Resource's name, but the server may decide to give it
+ * another name — for example, if a Resource with that name already exists inside the given
+ * Container.
+ * If the user does have access to write directly to a given location, [[overwriteFile]]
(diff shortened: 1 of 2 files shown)


src/resource/file.ts
```diff
@@ -139,6 +139,16 @@ type SaveFileOptions = WriteFileOptions & {
  * be rejected. You can initialise it first using [[createContainerAt]], or directly save the file
  * at the desired location using [[overwriteFile]].
  *
+ * This function is primarily useful if the current user does not access to change existing files in
```
pmcb55, line 142:
```suggestion
 * This function is primarily useful if the current user does not have access to change existing files in
```

src/resource/file.ts
```diff
@@ -139,6 +139,16 @@ type SaveFileOptions = WriteFileOptions & {
...
+ * a Container, but is allowed to add new files; in other words, they have Append, but not Write
+ * access to a Container. This is useful in situations where someone wants to allow others to e.g.
```
pmcb55, line 144: Personally I don't like the use of 'e.g.' in the middle of a sentence (it's Ok at the end of a sentence, or in brackets), so I'd just tweak this to:
```suggestion
 * access to a Container. This is useful in situations where someone wants to allow others to, for example,
```

src/resource/file.ts
```diff
@@ -139,6 +139,16 @@ type SaveFileOptions = WriteFileOptions & {
...
+ * send notifications to their Pod, but not to view or delete existing ones.
```
pmcb55, line 145:
```suggestion
 * send notifications to their Pod, but not to view or delete existing notifications.
```

pmcb55 (approved):
Looks good - just some suggested tweaks.

src/resource/file.ts
Vinnl: Incorporated this and the other suggestions in all three instances of this doc.

PR #699, opened 2020-12-29
Run import tests after regular releases as well

Vinnl:
After publishing a preview release, we automatically try whether
importing from the just-published packages work in a variety of
environments. This also enables that for regular releases.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -90,3 +90,116 @@ jobs:
         state: failure
       env:
         GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+    - name: Waiting for npm CDNs to update...
+      run: |
+        echo "Giving npm some time to make the newly-published package available…"
+        sleep 5m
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -90,3 +90,116 @@ jobs:
         state: failure
       env:
         GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+    - name: Waiting for npm CDNs to update...
+      run: |
+        echo "Giving npm some time to make the newly-published package available…"
+        sleep 5m
+        echo "Done — hopefully that was enough time for the follow-up jobs to install the just-published package, to verify that everything looks OK."
```
pmcb55, line 97: Just to emphasise that we're only done waiting, and not 'done' for the whole CI or anything:
```suggestion
        echo "Done waiting — hopefully that was enough time for the follow-up jobs to install the just-published package, to verify that everything looks OK."
```

pmcb55 (approved):

PR #703, opened 2020-12-30
Add get contained resource url all

ABhat00:
Added getContainedResourceUrlAll: a function that returns the urls for each resource contained in the given container

# Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The following changes have been implemented but not released yet:
 
 ## [Unreleased]
 
+- 'getContainedResourceUrlAll': a function that returns the urls for each resource contained in
+  the given container
+
 The following sections document changes that have been released already:
 
 ## [1.2.0] - 2020-12-02
--- a/src/resource/resource.test.ts
+++ b/src/resource/resource.test.ts
@@ -40,6 +40,7 @@ import {
   isContainer,
   isRawData,
   getContentType,
+  getContainedResourceUrlAll,
 } from "./resource";
 import { internal_cloneResource } from "./resource.internal";
 import {
@@ -47,8 +48,14 @@ import {
   IriString,
   WithServerResourceInfo,
   SolidClientError,
+  SolidDataset,
+  UrlString,
 } from "../interfaces";
 import { dataset } from "../rdfjs";
+import { mockContainerFrom, mockSolidDatasetFrom } from "./mock";
+import { dct, ldp, rdf } from "rdf-namespaces";
+import { solidDatasetAsMarkdown } from "./solidDataset";
+import { createThing, addUrl, setThing, addDatetime } from "..";
 
 function mockResponse(
   body?: BodyInit | null,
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -6,11 +6,17 @@ The following changes have been implemented but not released yet:
 
 ## [Unreleased]
 
+======
+
```
Vinnl, line 10: I presume that this is an artefact of a merge conflict?

```suggestion
```

CHANGELOG.md
```diff
@@ -6,11 +6,17 @@ The following changes have been implemented but not released yet:
...
 ### New features
 
 - If you know the content type of a file to upload via `overwriteFile` or `saveFileInContainer`,
   you can now manually set it using the `contentType` property in their `options` parameters.
 
+- 'getContainedResourceUrlAll': a function that returns the urls for each resource contained in
+  the given container
+  =
+
```
Vinnl, line 19: Likewise:

```suggestion
- 'getContainedResourceUrlAll': a function that returns the urls for each resource contained in
  the given container

```

(Also removed the whitespace in between the bulleted list, IIRC it won't display them as a single list otherwise.)

src/resource/resource.ts
```diff
@@ -29,13 +29,17 @@ import {
...
 } from "../interfaces";
 import { internal_toIriString } from "../interfaces.internal";
 import { fetch } from "../fetcher";
 import {
   internal_isUnsuccessfulResponse,
   internal_parseResourceInfo,
 } from "./resource.internal";
+import { getThing } from "../thing/thing";
+import { getIriAll } from "../thing/get";
+import { ldp } from "rdf-namespaces";
```
Vinnl, line 42: This is only a devDependency for convenience in the test, but we don't want to add it to our bundle just for the couple of URLs we use. In lieu of #114, we're just keeping them in `src/constants.ts`. (Though it looks like you're not using this import in the first place, so removing it is even safer :) )

```suggestion
```

src/resource/resource.ts
```diff
@@ -176,6 +180,29 @@ export function isPodOwner(
   return podOwner === webId;
 }
 
+/**
+ *  Given an solidDataset, fetch the urls of all contained resources.
```
Vinnl, line 184: Using `[[` and `]]` (or `{@link` and `}`), TypeDoc will include a link to its definition:

```suggestion
 *  Given a [[SolidDataset]], fetch the urls of all contained resources.
```

src/resource/resource.ts
```diff
@@ -176,6 +180,29 @@ export function isPodOwner(
...
+ *  If the solidDataset given is not a container, or is missing resourceInfo, throw an error.
+ *
+ * @param solidDataset The container from which to fetch all contained resource urls
+ * @returns A list of urls, each of which points to a contained resource of the given solidDataset
+ * @since 1.2.0
```
Vinnl, line 189: We'll update this version number when we make a new release, because the new version number depends on what else goes into that release.

```suggestion
 * @since Not released yet.
```

src/resource/resource.ts
```diff
@@ -176,6 +180,29 @@ export function isPodOwner(
...
+ */
+
+export function getContainedResourceUrlAll(
```
Vinnl, line 192: Since this function only applies to SolidDatasets (i.e. not to Files), it would be better placed in `src/resource/SolidDataset.ts`. Functions in this file apply to either :)

(The reason this matters is because that affects where it will be listed in the documentation, and also from which module people import the file when using [package exports](https://webpack.js.org/guides/package-exports/).)

src/resource/resource.ts
```diff
@@ -176,6 +180,29 @@ export function isPodOwner(
...
+  solidDataset: SolidDataset & WithResourceInfo
+): UrlString[] {
+  const contains = "http://www.w3.org/ns/ldp#contains";
+  const container = getThing(solidDataset, getSourceIri(solidDataset));
+  if (container) {
+    return getIriAll(container, contains);
+  } else {
+    // I'm making the assumption here that a dataset is a container if and only if it has a thing at
+    // its root with metadata about the resources contained - this might not be accurate
```
Vinnl, line 201: It is :) https://www.w3.org/TR/2015/REC-ldp-20150226/#h-ldpc-http_post

> a containment triple MUST be added to the state of the LDPC whose subject is the LDPC URI, whose predicate is ldp:contains and whose object is the URI for the newly created document (LDPR)

```suggestion
```

That said, I would simply return an empty array - the given Resource does not contain other Resources. We already have `isContainer()` to check if it's also a Container (might be useful to mention that in the API docs though).

src/resource/resource.test.ts
```diff
@@ -701,3 +708,68 @@ describe("cloneResource", () => {
...
+      });
+      childListing = addUrl(childListing, rdf.type, ldp.Resource);
+      childListing = addDatetime(childListing, dct.modified, new Date());
+
+      childrenIndex = addUrl(childrenIndex, ldp.contains, childListing);
+      mockContainer = setThing(mockContainer, childListing);
+    });
+
+    mockContainer = setThing(mockContainer, childrenIndex);
+    childrenIndex = addUrl(childrenIndex, rdf.type, ldp.Container);
```
Vinnl, line 734: Technically it doesn't matter since `getContainedResourceUrlAll()` doesn't check the `rdf:type` anyway, but this second line now doesn't do anything. (I'd also assume it to be next to the line in which you added the `BasicContainer` type, although both could also just be removed.)

```suggestion
    childrenIndex = addUrl(childrenIndex, rdf.type, ldp.Container);
    mockContainer = setThing(mockContainer, childrenIndex);
```

src/resource/resource.test.ts
```diff
@@ -701,3 +708,68 @@ describe("cloneResource", () => {
...
+
+  it("gets all urls for contained resources from a container", () => {
+    const containerUrl = "https://arbitrary.pod/container/";
+    const containedThings = ["resource1", "resource2", "resource3"];
+    const container = initContainer(containerUrl, containedThings);
+    const expectedReturnUrls = containedThings.map(
+      (thingName) => `${containerUrl}${thingName}.ttl`
+    );
+
+    expect(getContainedResourceUrlAll(container)).toStrictEqual(
```
Vinnl, line 747: Heh, I didn't know about `toStrictEqual`, but looks like that makes more sense to use in most cases where we use `toEqual`. TIL!

Vinnl (changes requested):
I added some inline comments, mostly about particulars to our particular code setup - mostly for information though, I'll apply them myself and fix up the Git history as I do so. (I'll submit a new PR and tag you in it so you can give it a final look if you want, and so I'm not introducing conflicts for you.)

Vinnl:
Alright, #705 incorporates the requested changes - let me know if it looks like what you'd expect, or if I didn't do your PR justice. And thanks for the contribution! :)

ABhat00:
#705 looks good! Thanks for the feedback, I appreciate it

PR #721, opened 2021-01-06
Docs: missing redirects

kay-kim:
Just adding a handful of redirects.


Vinnl (approved):
Ah sorry, I was under the impression that the old URLs weren't live yet, and thus didn't need a redirect.

Do you just have a very sharp eye, or do you have a clever way to detect missing redirects? Would be good if I could catch them myself in the future :)

kay-kim:
Oh - heh - when committing to the gh-branch for publishing, I just check if there are deletes.

PR #735, opened 2021-01-13
Feat/1375 has inaccessible policies

NSeydoux:
This implements internal_hasInaccessiblePolicies, an internal method
that detects if an ACR contains any external resources, either Policies
or Rules.

### DIFF
--- a/src/access/acp.test.ts
+++ b/src/access/acp.test.ts
@@ -20,19 +20,177 @@
  */
 
 import { describe, it } from "@jest/globals";
+import { AccessControlResource } from "../acp/control";
+import { addMockAcrTo, mockAcrFor } from "../acp/mock";
+import { createPolicy, setPolicy } from "../acp/policy";
+import { acp, rdf } from "../constants";
+import { UrlString, WithServerResourceInfo } from "../interfaces";
+import { mockSolidDatasetFrom } from "../resource/mock";
+import { createSolidDataset } from "../resource/solidDataset";
+import { addIri } from "../thing/add";
+import { createThing, setThing } from "../thing/thing";
+import { internal_hasInaccessiblePolicies } from "./acp";
+
+const defaultMockPolicies = {
+  policies: ["https://some.pod/policies#policy"],
+  memberPolicies: ["https://some.pod/policies#memberPolicy"],
+  acrPolicies: [] as string[],
+  memberAcrPolicies: [] as string[],
+};
+
+type Rule = {
+  allOf: string[];
+  anyOf: string[];
+  noneOf: string[];
+};
+
+function mockAcr(
+  accessTo: UrlString,
+  policies = defaultMockPolicies,
+  rules?: Record<string, Rule>
+): AccessControlResource {
+  let control = createThing({ name: "access-control" });
+  control = addIri(control, rdf.type, acp.AccessControl);
+  policies.policies.forEach((policyUrl) => {
+    control = addIri(control, acp.apply, policyUrl);
+  });
+  policies.memberPolicies.forEach((policyUrl) => {
+    control = addIri(control, acp.applyMembers, policyUrl);
+  });
+
+  const acrUrl = accessTo + "?ext=acr";
+  let acrThing = createThing({ url: acrUrl });
+  policies.acrPolicies.forEach((policyUrl) => {
+    acrThing = addIri(acrThing, acp.access, policyUrl);
+  });
+  policies.memberAcrPolicies.forEach((policyUrl) => {
+    acrThing = addIri(acrThing, acp.accessMembers, policyUrl);
+  });
+
+  let acr: AccessControlResource & WithServerResourceInfo = Object.assign(
+    mockSolidDatasetFrom(acrUrl),
+    {
+      accessTo: accessTo,
+    }
+  );
+  acr = setThing(acr, control);
+  acr = setThing(acr, acrThing);
+
+  if (rules !== undefined) {
+    Object.keys(rules).forEach((policyUrl) => {
+      let policyThing = createThing({ url: policyUrl });
+      rules[policyUrl].allOf.forEach((allOfUrl) => {
+        policyThing = addIri(policyThing, acp.allOf, allOfUrl);
+        let ruleThing = createThing({ url: allOfUrl });
+        ruleThing = addIri(ruleThing, rdf.type, acp.Rule);
+        acr = setThing(acr, ruleThing);
+      });
+      rules[policyUrl].anyOf.forEach((anyOfUrl) => {
+        policyThing = addIri(policyThing, acp.anyOf, anyOfUrl);
+        let ruleThing = createThing({ url: anyOfUrl });
+        ruleThing = addIri(ruleThing, rdf.type, acp.Rule);
+        acr = setThing(acr, ruleThing);
+      });
+      rules[policyUrl].noneOf.forEach((noneOfUrl) => {
+        policyThing = addIri(policyThing, acp.noneOf, noneOfUrl);
+        let ruleThing = createThing({ url: noneOfUrl });
+        ruleThing = addIri(ruleThing, rdf.type, acp.Rule);
+        acr = setThing(acr, ruleThing);
+      });
+      acr = setThing(acr, policyThing);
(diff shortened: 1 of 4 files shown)


src/access/acp.ts
```diff
@@ -0,0 +1,45 @@
...
+import { asUrl } from "../thing/thing";
+
+export function internal_hasInaccessiblePolicies(
+  resource: WithAccessibleAcr & WithResourceInfo
+): boolean {
+  const sourceIri = getSourceIri(resource);
+  const policyUrls = getPolicyUrlAll(resource);
+  const ruleUrls = getRuleAll(internal_getAcr(resource)).map((rule) =>
+    asUrl(rule)
+  );
```
Vinnl, line 37: I think this is not exactly what we need here: Rules with URLs that are not the ACR's URL are fine (though they probably won't exist in practice), as long as they're not _referenced_ from Policies that are referenced from the ACR's controls.

So I suppose we need three more tests:

- One that verifies that unreferenced Rules with other URLs do not result in this function returning `true`.
- One that verifies that unreferenced Policies with other URLs do not result in this function returning `true`.
- One that verifies that references to Rules with other URLs that do not exist in this Resource still results in this function returning `true`.

All three of those would currently fail, I think.

Vinnl (changes requested):
Great, you did all the hard work w.r.t. mocking for the ticket I'm working on :) 

Unfortunately, I think the current behaviour is wrong; see my inline comment. (Also, I presume you're going to squash these together before merging? :) )

src/access/acp.test.ts
```diff
@@ -0,0 +1,196 @@
...
+    const resourceWithAcr = addMockAcrTo(
+      plainResource,
+      mockAcr("https://some.pod/resource", {
+        policies: [],
+        memberAcrPolicies: [],
+        acrPolicies: [],
+        memberPolicies: [],
+      })
+    );
+    expect(internal_hasInaccessiblePolicies(resourceWithAcr)).toEqual(false);
```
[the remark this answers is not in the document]
NSeydoux: Right, but does using `toEqual` have a negative impact on performance ? My understanding was basically that `toBe` compared references, and `toEqual` values (deep equality for objects), and it made more sense to me to compare values here than references, even if the result is the same I guess for raw types. Does that make sense, or is using `toBe` a better practice ?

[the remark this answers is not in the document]
Vinnl: It is sometimes said that JavaScript is pass-by-value for primitive types, and pass-by-reference for objects. That's incorrect: JavaScript is just pass-by-value. It just happens that the value, when we think we're being passed an object, _is_ a reference.

`.toBe` does compare values; however, in the case of objects, those values are references - and they can be different references, even if both objects those references point to have similar values. So what `toEqual` does is follow those references, and then recursively loop over the properties of both objects to compare _their_ values. So yes, it can have a performance impact.

tl;dr in practice it does indeed not matter here, but since, as you say, you want to compare values, `toBe` better captures the intent. (And also Matthieu recently highlighted that Jest also has a `toStrictEqual`, which not only checks whether the properties of both objects have the same values, but also the same types. So in the cases where you _do_ want to use `toEqual`, I think you'd almost always want to use `toStringEqual` just like you'd prefer `===` over `==`.

src/access/acp.test.ts
```diff
@@ -118,6 +125,7 @@ describe("hasInaccessiblePolicies", () => {
     );
     expect(internal_hasInaccessiblePolicies(resourceWithAcr)).toEqual(false);
   });
+
```
Vinnl, line 128: Haha, I didn't want to "complain" about this missing line before, but I did notice it :P

src/access/acp.test.ts
```diff
@@ -132,7 +140,59 @@ describe("hasInaccessiblePolicies", () => {
     expect(internal_hasInaccessiblePolicies(resourceWithAcr)).toEqual(false);
   });
 
-  it("returns false if the ACR only contains reference to Rules in the same Resource", () => {
+  it("returns true if the ACR references an active Policy in a different Resource", () => {
+    const plainResource = mockSolidDatasetFrom("https://some.pod/resource");
+    const resourceWithAcr = addMockAcrTo(
+      plainResource,
+      mockAcr("https://some.pod/resource", {
+        policies: ["https://some.pod/anoter-resource?ext=acr#policy"],
```
Vinnl, line 148: Small typo, also present in the other test:

```suggestion
        policies: ["https://some.pod/another-resource?ext=acr#policy"],
```

src/access/acp.ts
```diff
@@ -20,24 +20,48 @@
...
-  const ruleUrls = getRuleAll(internal_getAcr(resource)).map((rule) =>
-    asUrl(rule)
-  );
-  // If either a policy or a rule are not defined in the ACR, return false
+
+  // Collect all policies that apply to the resource or its ACR (aka active)
+  const activePolicyUrls = getPolicyUrlAll(resource);
+  getAcrPolicyUrlAll(resource).forEach((policyUrl) => {
+    activePolicyUrls.push(policyUrl);
+  });
```
Vinnl, line 43: More a matter of style, but this saves some manual looping:

```suggestion
  const activePolicyUrls = getPolicyUrlAll(resource).concat(getAcrPolicyUrlAll(resource));
```

src/access/acp.ts
```diff
@@ -20,24 +20,48 @@
...
+        ruleUrls.push(activeRuleUrl)
+      );
+      getUrlAll(policyThing, acp.allOf).forEach((activeRuleUrl) =>
+        ruleUrls.push(activeRuleUrl)
+      );
+      getUrlAll(policyThing, acp.noneOf).forEach((activeRuleUrl) =>
+        ruleUrls.push(activeRuleUrl)
+      );
+    }
+  });
```
Vinnl, line 61: Again a matter of style (so feel free to ignore!), and `.flat()` is not supported in IE so would need addition to the list of polyfills mentioned in the README, but possibly interesting:

```suggestion
  const ruleUrls: string[] = activePolicyUrls.map(policyUrl => {
    const acr = internal_getAcr(resource);
    const policyThing = getThing(acr, policyUrl);
    if (policyThing === null) {
      return [];
    }
    return getUrlAll(policyThing, acp.anyOf)
      .concat(getUrlAll(policyThing, acp.allOf))
      .concat(getUrlAll(policyThing, acp.noneOf))
  }).flat();
```

Vinnl (approved):

src/access/acp.test.ts
NSeydoux: Resolved in 611a7c6a4a79997029573d28feaef0c8ef75523d

src/access/acp.ts
NSeydoux: Yeah it looks better, I remember thinking "How nice would it be to have an `addAll` method, or if `push` accepted a whole array", and left it there while I used concat three lines lower ^^.

Resolved in 82181eda8eee05b632e1f18511d407532666e438

src/access/acp.ts
NSeydoux: Again, very good suggestion, thanks ! It does look more readable and less convoluted. 4646c9c9066b76a173b5c9591a46485bd7bdf3dd

src/access/acp.test.ts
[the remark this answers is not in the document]
NSeydoux: Ok, thanks, that clarifies it :)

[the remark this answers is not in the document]
NSeydoux: Resolved in 93d6919951a89e7ea5f5854548dd9ff18ef77df5

src/access/acp.ts
NSeydoux: Well nevermind, `.flat` isn't supported in Node10 ^^

NSeydoux: Blessed be small commits 4b8093de4910785f3f321665c3d4a72d1a291b45

pmcb55 (approved):

pmcb55 (approved):

src/access/acp.ts
Vinnl: Aargh. Just four more months... :)

PR #740, opened 2021-01-15
Fix/docs for overwritefile

pmcb55:
# New feature description
Added documentation about the `contentType` option for write/save file methods.

# Checklist

### DIFF
--- a/src/resource/file.ts
+++ b/src/resource/file.ts
@@ -37,8 +37,18 @@ import {
   internal_parseResourceInfo,
 } from "./resource.internal";
 
-type GetFileOptions = {
+/**
+ * Options when fetching a file from a Pod.
+ *
+ * Available options:
+ * - `fetch`: A custom `fetch` function with the same signature as
+ *   [`window.fetch`](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/fetch).
+ *   This will be used to execute the actual requests. This option can be used to, for example,
+ *   attach credentials to requests that need authentication.
+ */
+export type GetFileOptions = {
   fetch: typeof window.fetch;
+  /** @internal */
   init: UploadRequestInit;
 };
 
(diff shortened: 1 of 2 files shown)


src/resource/file.ts
```diff
@@ -191,7 +213,24 @@ export async function saveFileInContainer<FileExt extends File | Buffer>(
   return Object.assign(blobClone, resourceInfo);
 }
 
+/**
+ * ```{note} The Web Access Control specification is not yet finalised. As such, this
+ * type is still experimental and subject to change, even in a non-major release.
```
Vinnl, line 218: I don't think this has anything to do with WAC? :) 

```suggestion
 * ```{note} This type is still experimental and subject to change, even in a
 * non-major release.
```

src/resource/file.ts
```diff
@@ -135,6 +148,15 @@ type SaveFileOptions = WriteFileOptions & {
  * Saves a file in a folder associated with the given URL. The final filename may or may
  * not be the given `slug`.
  *
+ * If you know the [[https://developer.mozilla.org/en-US/docs/Glossary/MIME_type | media type]]
```
Vinnl, line 151: Are you sure this syntax works (and if so, where is it defined)? In the end it's going to be parsed by Sphinx's markdown parser, so I know the following syntax should work - though if the current syntax works, that's fine too of course.

This also applies to the other links in this PR.

```suggestion
 * If you know the [media type](https://developer.mozilla.org/en-US/docs/Glossary/MIME_type)
```

Vinnl (approved):
Make sure to check the syntax for links and remove the part about WAC, but otherwise looks fine.

src/resource/file.ts
Vinnl: Ah of course I can just check that it looks fine: https://lit-pod-e7sms1j08.vercel.app/modules/resource_file.html#overwritefile

(Although the link to `[[options]]` does not. Anyway, I'm in the process of fixing up this PR, so I'll fix it.)

PR #751, opened 2021-01-20
Rework ACR mock functions: allow Rule initialising

Vinnl:
This reworks the mock functions added by Nicolas in #735 to also allow adding actual data to the Rules. This allows you to initialise an ACR as such:

```
const resourceWithAcr = mockResourceWithAcr(
  "https://some.pod/resource",
  "https://some.pod/resource?ext=acr",
  {
    policies: {
      "https://some.pod/resource?ext=acr#policy": {
        allOf: {
          "https://some.pod/other-rule-resource#rule": {
            [acp.agent]: "https://some.pod/profile#me",
          },
        },
      },
    },
    memberAcrPolicies: {},
    acrPolicies: {},
    memberPolicies: {},
  }
);
```

This isn't actually necessary for the `hasInaccessiblePolicies` function, but it will be useful for the functions that actually read ACP access. I'm working on those right now, but to lower the review burden, I'm submitting this as a separate PR first.

Additionally, I found out that the function `removeRule()` was still missing from the low-level API, so I added that too.

### DIFF
--- a/src/access/acp.test.ts
+++ b/src/access/acp.test.ts
@@ -20,349 +20,388 @@
  */
 
 import { describe, it } from "@jest/globals";
+import { WithAccessibleAcr } from "../acp/acp";
 import { AccessControlResource } from "../acp/control";
-import { addMockAcrTo, mockAcrFor } from "../acp/mock";
-import { createPolicy, setPolicy } from "../acp/policy";
-import { acp, rdf } from "../constants";
-import { UrlString, WithServerResourceInfo } from "../interfaces";
+import { internal_createControl } from "../acp/control.internal";
+import { addMockAcrTo } from "../acp/mock";
+import { createPolicy, removePolicy, setPolicy } from "../acp/policy";
+import {
+  addForbiddenRuleUrl,
+  addOptionalRuleUrl,
+  addRequiredRuleUrl,
+  createRule,
+  Rule,
+} from "../acp/rule";
+import { acp } from "../constants";
+import {
+  IriString,
+  UrlString,
+  WithResourceInfo,
+  WithServerResourceInfo,
+} from "../interfaces";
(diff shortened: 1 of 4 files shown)


src/access/acp.test.ts
```diff
@@ -20,349 +20,386 @@
...
+import { setThing } from "../thing/thing";
 import { internal_hasInaccessiblePolicies } from "./acp";
 
-type Policy = {
-  policies: string[];
-  memberPolicies: string[];
-  acrPolicies: string[];
-  memberAcrPolicies: string[];
+// Key: actor relation (e.g. agent), value: actor (e.g. a WebID)
+type MockRule = Record<UrlString, UrlString[]>;
```
NSeydoux, line 48: Would this be valid ?
```suggestion
type MockRule = Record<acp.agent | acp.group, UrlString[]>;
```

NSeydoux (approved):

NSeydoux:
Can I rebase https://github.com/inrupt/solid-client-js/commit/b912a3afdf91cf41861f1aafba5f739e31640fa6 on this to adapt the mock functions accordingly, or do you anticipate additional changes ?

src/access/acp.test.ts
Vinnl: Hmm good question - it might be. I'll give that a try.

Vinnl:
@NSeydoux I do anticipate additional changes (in fact, I've already made them in my branch), so I'd wait until the rest of the work is complete.

src/access/acp.test.ts
Vinnl: Very cool, it works: 6776b298

Now we get autocompletion when defining the mock object, which is very helpful given how complex they are. Thanks!

PR #762, opened 2021-01-25
Feat/1379 get access all

NSeydoux:
This implements an internal function, `internal_getActorAccessAll`, on which the public API functions `get*AccessAll` will be based. This is also based on changes in #753 , and should only be merged after the other PR has been merged (I'll leave it in draft in the meantime). 

An upcoming PR will add the relevant functions to the public API.

# Checklist

### DIFF
--- a/src/access/acp.test.ts
+++ b/src/access/acp.test.ts
@@ -3104,9 +3104,7 @@ describe("internal_getActorAccessAll", () => {
           memberAcrPolicies: {},
         }
       );
-      expect(internal_getActorAccessAll(resourceWithAcr, actor)).toStrictEqual(
-        {}
-      );
+      expect(internal_getActorAccessAll(resourceWithAcr, actor)).toBeNull();
     }
   );
 
@@ -3228,7 +3226,7 @@ describe("internal_getActorAccessAll", () => {
   );
 
   it.each([acp.agent, acp.group])(
-    "returns an empty object if an external policy is present",
+    "returns null if an external policy is present",
     (actor) => {
       const resourceWithAcr = mockResourceWithAcr(
         "https://some.pod/resource",
--- a/src/access/acp.test.ts
+++ b/src/access/acp.test.ts
@@ -3137,6 +3137,7 @@ describe("internal_getActorAccessAll", () => {
       internal_getActorAccessAll(resourceWithAcr, acp.group)
     ).toStrictEqual({});
   });
+
   it("does not return access given to groups for agents", () => {
     const resourceWithAcr = mockResourceWithAcr(
       "https://some.pod/resource",
@@ -3220,6 +3221,36 @@ describe("internal_getActorAccessAll", () => {
         }
       );
 
+      expect(internal_getActorAccessAll(resourceWithAcr, actor)).toStrictEqual(
+        {}
+      );
+    }
+  );
+
+  it.each([acp.agent, acp.group])(
+    "returns an empty object if an external policy is present",
+    (actor) => {
+      const resourceWithAcr = mockResourceWithAcr(
+        "https://some.pod/resource",
+        "https://some.pod/resource.acr",
+        {
+          policies: {
+            "https://some.pod/another-resource.acr#policy": {
+              anyOf: {
+                "https://some.pod/resource.acr#rule": {
+                  [actor]: ["https://some.pod/some-actor"],
+                },
+              },
+              allow: {
+                read: true,
+              },
+            },
+          },
+          memberPolicies: {},
+          acrPolicies: {},
+          memberAcrPolicies: {},
+        }
+      );
       expect(internal_getActorAccessAll(resourceWithAcr, actor)).toStrictEqual(
         {}
       );
--- a/src/access/acp.test.ts
+++ b/src/access/acp.test.ts
@@ -3088,3 +3088,38 @@ describe("getActorAccess", () => {
     });
   });
 });
+
+describe("getAgentAccessAll", () => {
+  // it("returns an empty list if no individual agent is given access", async () => {
+  //   const plainResource = mockSolidDatasetFrom("https://some.pod/resource");
+  //   const resourceWithAcr = addMockAcrTo(
+  //     plainResource,
+  //     mockAcr("https://some.pod/resource")
+  //   );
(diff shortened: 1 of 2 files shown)


NSeydoux:
One global question for reviewers: in the current implementation, the Actors which are present in the rule, but not given any access (e.g. missing from an `allOf` Rule) are still listed in the result set, instead of not being present. I think it gives some information that can be useful to know that one is present in an ACR without any access granted, but I don't have a strong opinion on this and could see why we'd only want to include Actors to which at least one policy in the ACR do apply, allowing or denying some modes.

src/access/acp.test.ts
```diff
@@ -2618,3 +2619,569 @@ describe("getActorAccess", () => {
     });
   });
 });
+
+describe("internal_getActorAccessAll", () => {
+  it.each([acp.agent, acp.group])(
```
Vinnl, line 2624: Given that this hits the same code path and the code doesn't (nor shouldn't it) distinguish between the different types of actors, as far as I'm concerned this isn't necessary :) (Can't harm, but just in case it's going to cause you more work in finishing this up.)

Vinnl (commented):
> One global question for reviewers: in the current implementation, the Actors which are present in the rule, but not given any access (e.g. missing from an allOf Rule) are still listed in the result set, instead of not being present. I think it gives some information that can be useful to know that one is present in an ACR without any access granted, but I don't have a strong opinion on this and could see why we'd only want to include Actors to which at least one policy in the ACR do apply, allowing or denying some modes.

There are two options I'd be fine with:

- List actors without access, _but_ only list actors that are actually reachable from the ACR's Controls. (Currently you list every actor mentioned in a Rule, even if that Rule isn't referred to in a Policy, or if that Policy is not referred to from a Control.)
- Remove actors for which every type of access is `undefined` (but not `false`).

src/access/acp.ts
```diff
@@ -236,3 +243,54 @@ function ruleAppliesTo(
...
+
+export function internal_getActorAccessAll(
+  resource: WithResourceInfo & WithAcp,
+  actorRelation: typeof acp.agent | typeof acp.group
+): Record<string, Access> {
+  if (
+    !hasAccessibleAcr(resource) ||
+    internal_hasInaccessiblePolicies(resource)
+  ) {
+    return {};
```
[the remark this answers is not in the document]
NSeydoux: Good point, this is fixed here 5e951f4dcb1df70b173bef4ce9a5e83ae34b2d99, and added to the TSdoc comment

src/access/acp.test.ts
NSeydoux: Yeah I only realized after I changed it that it didn't cover anything specific. No harm in leaving it in I guess.

NSeydoux:
True, it's unnecessary to list actors from rules that aren't referenced from the ACR Control. This is fixed in 5e951f4dcb1df70b173bef4ce9a5e83ae34b2d99.

src/access/acp.test.ts
```diff
@@ -2618,3 +2619,567 @@ describe("getActorAccess", () => {
...
+      {
+        policies: {
+          "https://some.pod/resource.acr#policy": {
+            anyOf: {
+              "https://some.pod/resource.acr#rule": {
+                [acp.agent]: ["https://some.pod/profile#agent"],
+              },
+            },
+            allow: {
+              append: true,
```
Vinnl, line 2656: Btw, no biggie (I haven't stuck to this in earlier unit tests either), but generally I think it is a good idea to only vary one property that you're explicitly testing and leave the rest exactly the same. So if you want to test whether it also works for `append`, that would be a separate test. This makes the tests more predictable in the sense that some tests won't suddenly break on an unrelated change whereas others will.

Again, no need to fix this here, but maybe something to keep in mind for the future.

src/access/acp.ts
```diff
@@ -29,30 +29,31 @@ import {
...
 
+function getActiveRuleAll(
+  resource: WithAccessibleAcr & WithResourceInfo,
+  policyUrlAll: UrlString[]
+): UrlString[] {
   // Collect all the rules referenced by the active policies.
   const ruleUrls: string[] = [];
-  activePolicyUrls.forEach((policyUrl) => {
+  policyUrlAll.forEach((policyUrl) => {
     const acr = internal_getAcr(resource);
```
Vinnl, line 57: Reeeaaally minor :)

```suggestion
  const acr = internal_getAcr(resource);
  policyUrlAll.forEach((policyUrl) => {
```

src/access/acp.ts
```diff
@@ -236,3 +251,69 @@ function ruleAppliesTo(
 ): boolean {
   return rule !== null && getIriAll(rule, actorRelation).includes(actor);
 }
+
+/**
+ * Get a set of all actors mentionned in an ACR by active Rules (i.e. that are
+ * references by Policies referenced by the ACR Control, and therefore that
```
Vinnl, line 257:
```suggestion
 * referenced by Policies referenced by the ACR Control, and therefore that
```

src/access/acp.ts
```diff
@@ -236,3 +251,69 @@ function ruleAppliesTo(
...
+        (iri) =>
+          !([acp.PublicAgent, acp.CreatorAgent] as string[]).includes(iri) || actorRelation != acp.agent
+      )
+      .forEach((iri) => actors.add(iri));
+  });
+  return actors;
+}
+
+/**
+ * Iterate through all the actors active for an ACR, and list all of their access.
```
Vinnl, line 291: Might be good to echo some of the caveats from `getActorAccess` here, i.e. that this doesn't return the effective access.

```suggestion
 * Iterate through all the actors active for an ACR, and list all of the access granted or denied explicitly to them.
```

src/access/acp.ts
```diff
@@ -236,3 +251,69 @@ function ruleAppliesTo(
...
+ * @param resource The resource for which we want to list the access
+ * @param actorRelation The type of actor we want to list access for
+ * @returns A map with each actor access indexed by their WebID, or null if some
```
Vinnl, line 294: Since groups don't have a WebID:

```suggestion
 * @returns A map with each actor access indexed by their identifying URL, or null if some
```

Vinnl (approved):
Added some pedantry, but looks good.

PR #764, opened 2021-01-25
Add actor-specific get<Actor>Access() functions

Vinnl:
# New feature description

This adds ACP-specific functions `getAgentAccess`, `getGroupAccess`, `getPublicAccess` and `getAuthenticatedAgent` access, to be later used by the more generic access API.

# Checklist

- [x] New functions/types have been exported in `index.ts`, if applicable. N/A
- [x] New modules (i.e. new `.ts` files) are listed in the `exports` field in `package.json`, if applicable. N/A
- [x] New modules (i.e. new `.ts` files) are listed in the `typedocOptions.entryPoints` field in `tsconfig.json`, if applicable. N/A

### DIFF
--- a/src/access/acp.test.ts
+++ b/src/access/acp.test.ts
@@ -52,6 +52,10 @@ import { addIri, addUrl } from "../thing/add";
 import { setThing } from "../thing/thing";
 import {
   internal_getActorAccess,
+  internal_getAgentAccess,
+  internal_getAuthenticatedAccess,
+  internal_getGroupAccess,
+  internal_getPublicAccess,
   internal_hasInaccessiblePolicies,
 } from "./acp";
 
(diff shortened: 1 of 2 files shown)


NSeydoux (approved):
It all looks good. Would it make sense to add an additional test for each of these public functions to check that it indeed calls the underlying `internal_getActorAccess` ? I know that ideally we don't want to be testing for implementation and rather for behaviour, but in this instance, `internal_getActorAccess` has extensive tests for as many combinations of rules and policies as possible, combinations that technically should be tested for any of the public API function as well. Instead of massively duplicating the code, testing for implementation would make sure that the code covered in `internal_getActorAccess` applies to all the other functions. It's a bit nitpicky, so I'm also happy if we simply consider that superficial tests are enough because it's obvious that the implementation is based on the carefully tested function.

Vinnl:
I'll take a look tomorrow to see if that can easily be added; can't do too much harm.

Vinnl:
@NSeydoux Unfortunately it seems hard to spy on this function without mocking the entire module (which I don't want to do since we're _testing_ that module :) ). I did the following, but that says that it's getting called 0 times even though I can clearly see it calling the function - it's just that that function does not appear to be replaced by the spy:

```
    it("calls out to the well-tested function getActorAccess() to cover all edge cases", () => {
      const acpModule = jest.requireActual("./acp.ts");
      const getActorAccessSpy = jest.spyOn(acpModule as any, "internal_getActorAccess");

      const resourceWithAcr = mockResourceWithAcr("https://some.pod/resource");

      internal_getAgentAccess(resourceWithAcr, webId);

      expect(getActorAccessSpy).toHaveBeenCalledTimes(1);
      expect(getActorAccessSpy).toHaveBeenCalledWith(resourceWithAcr, acp.agent, webId);
      getActorAccessSpy.mockRestore();
    });
```

Since I've spent quite a bit of time on this already, I'm going to merge this without that, if you don't mind.

NSeydoux:
No problem, go ahead ! It was really minor, since our chances to mess up the implementation without failing the superficial tests are really low.

PR #774, opened 2021-01-27
Fetched dataset don't conflict when saved back

### LINKED ISSUE #736: solid-client doesn't PATCH unchanged SolidDatasets
### Search terms you've used
412 Precondition failed

### Bug description

solid-client doesn't initialise an empty ChangeLog for fetched SolidDatasets, leading to `saveSolidDatasetAt()` not recognising that it needs to send an empty PATCH request rather than a PUT when saving back that SolidDataset unchanged.

### To Reproduce
1. Fetch a dataset from an ESS Pod with `getSolidDataset`
2. Immediately save it to the same location (assuming you have permissions to) using `saveSolidDatasetAt`
3. Observe the 412 Precondition Failed

NSS doesn't appear to enforce the precondition and therefore does not run into this.

**Minimal reproduction**

This should be a reproduction, but I since I wasn't able to make it publicly writeable so far it's not: https://codesandbox.io/s/async-dream-2j4bp?file=/src/index.ts

### Expected result

A successful (albeit useless) request.

### Actual result

```
```
This was first encountered by @ABhat00, I'm not sure what his environment was.

NSeydoux:
Resolves #736

This ensures that datasets which have been fetched from a remote
resource are PATCHed when saved back to the same resource instead of
PUT, even if they haven't been changed locally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,9 +8,9 @@ The following changes have been implemented but not released yet:
 
 - Added documentation about the `contentType` option for file writing/saving functions.
 
-### Bugfix
+### Bugs fixed
 
-- Saving back a dataset you just fetched no longer ends with "412: Conflict".
+- Saving back a SolidDataset you just fetched used to result in a "412: Conflict" error.
 
 The following sections document changes that have been released already:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ The following changes have been implemented but not released yet:
 
 - Added documentation about the `contentType` option for file writing/saving functions.
 
+### Bugfix
+
+- Saving back a dataset you just fetched no longer ends with "412: Conflict".
+
 The following sections document changes that have been released already:
 
 ## [1.3.0] - 2021-01-07
--- a/src/resource/solidDataset.ts
+++ b/src/resource/solidDataset.ts
@@ -55,7 +55,12 @@ import {
   internal_isUnsuccessfulResponse,
   internal_parseResourceInfo,
 } from "./resource.internal";
-import { thingAsMarkdown, getThingAll, getThing } from "../thing/thing";
+import {
+  thingAsMarkdown,
+  getThingAll,
+  getThing,
+  withChangeLog,
+} from "../thing/thing";
 import {
   internal_getReadableValue,
   internal_toNode,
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ The following changes have been implemented but not released yet:
 
 - Added documentation about the `contentType` option for file writing/saving functions.
 
+### Bugfix
+
+- Saving back a dataset you just fetched no longer ends with "412: Conflict".
```
Vinnl, line 13: Just to remain consistent with the rest of the changelog:

```suggestion
### Bugs fixed

- Saving back a SolidDataset you just fetched used to result in a "412: Conflict" error.
```

src/thing/thing.ts
```diff
@@ -228,7 +228,7 @@ export function removeThing<Dataset extends SolidDataset>(
   return newSolidDataset;
 }
 
-function withChangeLog<Dataset extends SolidDataset>(
+export function withChangeLog<Dataset extends SolidDataset>(
```
Vinnl, line 231: If you're exporting this, make sure to rename it to `internal_withChangelog` and move it to `thing.internal.ts`, to avoid people reaching in and using this without API stability guarantees.

src/resource/solidDataset.ts
```diff
@@ -217,18 +222,21 @@ export async function saveSolidDatasetAt<Dataset extends SolidDataset>(
     ...options,
   };
 
-  const requestInit = isUpdate(solidDataset, url)
-    ? await prepareSolidDatasetUpdate(solidDataset)
-    : await prepareSolidDatasetCreation(solidDataset);
+  // const datasetWithChangelog = withChangeLog(solidDataset);
```
Vinnl, line 225:
```suggestion
```

Vinnl (commented):
The bugfix I had in mind was to make sure the return value of `getSolidDataset` included a changelog. I think this should work too, but I'm a bit worried that this will also result in the saving of a new SolidDataset (i.e. `saveSolidDatasetAt("https://somewhere", createSolidDataset())`) also becoming a PATCH request. I'm not sure if that's the case, but could you verify that to be sure? (And maybe _if_ we don't have a test case covering that, add one? So basically the complement of the test you wrote: "it tries to create a new Resource if the change log contains no change but the Resource was newly initialised".)

NSeydoux:
Saving a Dataset is done in a PATCH request under two conditions: 
1- The dataset has a changelog (that was the condition that was biting us here)
2- The dataset has a source IRI, and the target IRI when saving matches this source IRI.

I think the problem you describe is prevented by (2), but there's no harm in adding a test to be sure.

src/resource/solidDataset.ts
NSeydoux: Oops, I commented that out to make sure my new test was failing without my fix, and then re-implemented my fix on the next line, because why not ^^.
5ad3ec10eda88b7fa7a735c71cb127706140e7bb

src/thing/thing.ts
NSeydoux: 5ad3ec10eda88b7fa7a735c71cb127706140e7bb

Vinnl: That didn't move it to `thing.internal.ts` :)

Vinnl (approved):
Ah great. New test looks good, so as far as I'm concerned, after moving the internal function to `thing.internal.ts` this is good to go :)

(Otherwise [it's also mentioned in the docs](https://lit-pod-r1shixk7h.vercel.app/modules/thing_thing.html#internal_withchangelog) and available to import even in Node 12+.)

PR #775, opened 2021-01-27
Fix updating of Things containing Blank Nodes

Vinnl:
This bug was caused by a combination of two factors:

1. When you fetch a SolidDataset twice, the two instances of the
   same Blank Node inside those two different SolidDatasets are not
   considered equal.
2. When a developer calls setThing(), we first delete all that
   Thing's existing Quads from the SolidDataset, and then re-add
   the ones of the given Thing. To prevent the same Quad from being
   both deleted and added in the same request, after re-adding a
   Quad we check if that Quad was already listed in the deletions
   of the ChangeLog, and if it is, rather than adding it to the
   ChangeLog's additions, we remove it from the deletions.

Now, checking if a Quad was listed in the deletions would fail for
Quads containing Blank Nodes if this was a different instance of
the same SolidDataset, because those Blank Nodes are considered
different.

To get around this, I've changed the code such that, when looking
whether a Quad was already listed in the deletions, I now consider
any two Blank Node objects equivalent. This should be safe, because
when a Quad with a Blank Node is listed in the additions, that can
only happen because it was already present in the first place.
After all, solid-client does not support adding Blank Nodes. (Well,
technically you can manually pass it one, but since that Blank Node
is not used as a Subject anywhere, that would not do anything.)


src/e2e-node/e2e.test.ts
```diff
@@ -540,6 +542,54 @@ describe.each(serversUnderTest)(
...
+      });
+      const updatedDataset = setThing(refetchedDataset, updatedThing);
+      await expect(
+        saveSolidDatasetAt(datasetUrl, updatedDataset, {
+          fetch: session.fetch,
+        })
+      ).resolves.not.toThrow();
+
+      // Clean up after ourselves
+      await deleteSolidDataset(datasetUrl, { fetch: session.fetch });
```
pmcb55, line 590: In general, shouldn't this clean-up be in a `finally` block (in case the test fails after creating this resource on the remote Pod)? But even that still wouldn't guarantee we **_always_** clean up properly (as the connection could drop before the DELETE arrives), so I'd always create a fixed-named resource (specifically named just for this test), and just attempt to delete that resource (whether it exists or not) as the first step of the test. (I assume this is why you use `${session.info.sessionId}` in the resource name - but I don't think that's necessary really.)

Vinnl: Ah yes, good point, I'll change that. I specifically avoided the fixed-named resource for two reasons:

- Multiple end-to-end tests may be running at the same time (e.g. in CI, in different versions of Node or different operating systems), so I don't want those to read and write to the same resource.
- I'd prefer not to risk have a failing test influence future tests, or a failing cleanup at the start of the test to influence its behaviour.

Given that an empty, unused resource is cheap, that felt like a fair trade-off.

src/thing/thing.ts
```diff
@@ -187,6 +190,32 @@ export function setThing<Dataset extends SolidDataset>(
   return newDataset;
 }
 
+/**
+ * Compare two Quads but, if both Quads have objects that are Blank Nodes and are otherwise equal, treat them as equal.
+ *
+ * The reason we do this is because you cannot write Blank Nodes using solid-client.
```
pmcb55, line 196: Maybe expand this - seems weird on first reading. We do support BNodes as rdf:object (limited), but not (yet) as rdf:subject.

Vinnl: 11b928a2

NSeydoux (approved):

pmcb55 (approved):
Approving this as it fixes the known bug - but I'm sure BNodes are something we'll have to revisit...

PR #786, opened 2021-02-01
Feat/1669 get access wac

NSeydoux:
This adds the getAgentAccess function to list permissions an Agent is
given for a Resource. This function fecthes the metadata, so it deals
with the notion of fallback ACL internally. Note that in order to
accomodate for the universal API, this adapts the notion of Access as
initially defined in the ACL logic to align with the fact that in ACP, there
is a distinction between an access which is denied or simply not granted. 

# Checklist

### DIFF
--- a/src/access/wac.test.ts
+++ b/src/access/wac.test.ts
@@ -88,8 +88,8 @@ describe("getAgentAccess", () => {
           status: 404,
           url: "https://some.pod/resource.acl",
         })
-        // Link to the fallback ACL...
       )
+      // Link to the fallback ACL...
       .mockResolvedValueOnce(
         mockResponse("", {
           status: 200,
@@ -98,8 +98,8 @@ describe("getAgentAccess", () => {
             Link: '<.acl>; rel="acl"',
           },
         })
-        // Get the fallback ACL
       )
+      // Get the fallback ACL
       .mockResolvedValueOnce(
         mockResponse("", {
           status: 404,
@@ -160,6 +160,10 @@ describe("getAgentAccess", () => {
 
     await expect(result).resolves.toStrictEqual({
       read: true,
+      append: undefined,
+      write: undefined,
+      controlRead: undefined,
+      controlWrite: undefined,
     });
   });
 
@@ -180,8 +184,8 @@ describe("getAgentAccess", () => {
           status: 404,
           url: "https://some.pod/resource.acl",
         })
-        // Link to the fallback ACL...
       )
+      // Link to the fallback ACL...
       .mockResolvedValueOnce(
         mockResponse("", {
           status: 200,
@@ -190,8 +194,8 @@ describe("getAgentAccess", () => {
             Link: '<.acl>; rel="acl"',
           },
         })
-        // Get the fallback ACL
       )
+      // Get the fallback ACL
       .mockResolvedValueOnce(
         mockResponse(await triplesToTurtle(Array.from(aclResource)), {
           status: 200,
@@ -208,6 +212,10 @@ describe("getAgentAccess", () => {
     });
     await expect(result).resolves.toStrictEqual({
       read: true,
+      append: undefined,
+      write: undefined,
+      controlRead: undefined,
+      controlWrite: undefined,
     });
   });
 
@@ -216,7 +224,7 @@ describe("getAgentAccess", () => {
       getMockDataset("https://some.pod/.acl"),
       "https://some.pod/profile#agent",
       "https://some.pod/",
-      { read: true, append: false, write: false, control: false },
+      { read: false, append: true, write: false, control: false },
       "default"
     );
 
@@ -224,7 +232,7 @@ describe("getAgentAccess", () => {
       getMockDataset("https://some.pod/resource.acl"),
       "https://some.pod/profile#agent",
       "https://some.pod/resource",
-      { read: false, append: true, write: false, control: false },
+      { read: true, append: false, write: false, control: false },
       "resource"
     );
 
@@ -236,8 +244,8 @@ describe("getAgentAccess", () => {
           status: 200,
           url: "https://some.pod/resource.acl",
         })
-        // Link to the fallback ACL...
       )
+      // Link to the fallback ACL...
       .mockResolvedValueOnce(
         mockResponse("", {
           status: 200,
@@ -246,8 +254,8 @@ describe("getAgentAccess", () => {
             Link: '<.acl>; rel="acl"',
           },
         })
-        // Get the fallback ACL
       )
+      // Get the fallback ACL
       .mockResolvedValueOnce(
         mockResponse(await triplesToTurtle(Array.from(fallbackAclResource)), {
           status: 200,
@@ -263,11 +271,114 @@ describe("getAgentAccess", () => {
       fetch: mockFetch,
     });
     await expect(result).resolves.toStrictEqual({
-      append: true,
+      append: undefined,
+      read: true,
+      write: undefined,
+      controlRead: undefined,
+      controlWrite: undefined,
     });
   });
 
-  it("returns an empty object if the actor isn't present", async () => {
+  it("returns true for both controlRead and controlWrite if the Agent has control access", async () => {
+    const aclResource = mock_addAclRuleQuads(
+      getMockDataset("https://some.pod/resource.acl"),
+      "https://some.pod/profile#agent",
+      "https://some.pod/resource",
+      { read: false, append: false, write: false, control: true },
+      "resource"
+    );
+
+    const mockFetch = jest.fn(window.fetch).mockResolvedValue(
+      mockResponse(await triplesToTurtle(Array.from(aclResource)), {
+        status: 200,
+        url: "https://some.pod/resource.acl",
+      })
+    );
+
+    const resource = getMockDataset(
+      "https://some.pod/resource",
+      "https://some.pod/resource.acl"
+    );
+    const result = getAgentAccess(resource, "https://some.pod/profile#agent", {
+      fetch: mockFetch,
+    });
+
+    await expect(result).resolves.toStrictEqual({
+      read: undefined,
+      append: undefined,
+      write: undefined,
+      controlRead: true,
+      controlWrite: true,
+    });
+  });
+
+  it("correctly reads the Agent append access", async () => {
+    const aclResource = mock_addAclRuleQuads(
+      getMockDataset("https://some.pod/resource.acl"),
+      "https://some.pod/profile#agent",
+      "https://some.pod/resource",
+      { read: false, append: true, write: false, control: false },
+      "resource"
+    );
+
+    const mockFetch = jest.fn(window.fetch).mockResolvedValue(
+      mockResponse(await triplesToTurtle(Array.from(aclResource)), {
+        status: 200,
+        url: "https://some.pod/resource.acl",
+      })
+    );
+
+    const resource = getMockDataset(
+      "https://some.pod/resource",
+      "https://some.pod/resource.acl"
+    );
+    const result = getAgentAccess(resource, "https://some.pod/profile#agent", {
+      fetch: mockFetch,
+    });
+
+    await expect(result).resolves.toStrictEqual({
+      read: undefined,
+      append: true,
+      write: undefined,
+      controlRead: undefined,
+      controlWrite: undefined,
+    });
+  });
+
+  it("correctly reads the Agent write access, which implies append", async () => {
+    const aclResource = mock_addAclRuleQuads(
+      getMockDataset("https://some.pod/resource.acl"),
+      "https://some.pod/profile#agent",
+      "https://some.pod/resource",
+      { read: false, append: false, write: true, control: false },
+      "resource"
+    );
+
+    const mockFetch = jest.fn(window.fetch).mockResolvedValue(
+      mockResponse(await triplesToTurtle(Array.from(aclResource)), {
+        status: 200,
+        url: "https://some.pod/resource.acl",
+      })
+    );
+
+    const resource = getMockDataset(
+      "https://some.pod/resource",
+      "https://some.pod/resource.acl"
+    );
+    const result = getAgentAccess(resource, "https://some.pod/profile#agent", {
+      fetch: mockFetch,
+    });
+
+    await expect(result).resolves.toStrictEqual({
+      read: undefined,
+      append: true,
+      write: true,
+      controlRead: undefined,
+      controlWrite: undefined,
+    });
+  });
+
+  it("returns undefined for all modes the Agent isn't present", async () => {
     const aclResource = mock_addAclRuleQuads(
       getMockDataset("https://some.pod/resource.acl"),
       "https://some.pod/profile#another-agent",
@@ -291,7 +402,13 @@ describe("getAgentAccess", () => {
       fetch: mockFetch,
     });
 
-    await expect(result).resolves.toStrictEqual({});
+    await expect(result).resolves.toStrictEqual({
+      read: undefined,
+      append: undefined,
+      write: undefined,
+      controlRead: undefined,
+      controlWrite: undefined,
+    });
   });
 
   it("does not return access for groups", async () => {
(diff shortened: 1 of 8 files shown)


src/access/wac.ts
```diff
@@ -0,0 +1,81 @@
...
+function getDefaultAccess(): Access {
+  return {} as Access;
+}
+
+function aclAccessToUniversal(access: AclAccess): Access {
+  const universalAccess = getDefaultAccess();
+  // In ACL, denying access to an actor is a notion that doesn't exist, so an
+  // access is either granted or not for a given mode.
+  // This creates a misalignment with the ACP notion of an access being granted,
+  // denied, or simply not mentionned. Here, we convert the boolean vision of
```
pmcb55, line 38:
```suggestion
  // denied, or simply not mentioned. Here, we convert the boolean vision of
```

src/acl/mock.internal.ts
```diff
@@ -0,0 +1,111 @@
...
+  agent: IriString,
+  resource: IriString,
+  access: Access,
+  type: "resource" | "default" | "legacyDefault",
+  ruleIri?: IriString,
+  targetType:
+    | "http://www.w3.org/ns/auth/acl#agent"
+    | "http://www.w3.org/ns/auth/acl#agentGroup"
+    | "http://www.w3.org/ns/auth/acl#agentClass"
+    | "http://www.w3.org/ns/auth/acl#origin" = "http://www.w3.org/ns/auth/acl#agent"
```
pmcb55, line 37: Can you not move all these strings into the `Constants.ts` file, and then also make them IRI's (so you don't have to wrap in `DataFactory.namedNode()` calls throughout this file...? In other words, it seems we have some terms in the Constant file, but not all...?

src/constants.ts
```diff
@@ -25,6 +25,7 @@
 /** @hidden */
 export const acl = {
```
pmcb55, line 26: We could discuss this, but the current naming convention we have for generated vocabs is to upper-case them (i.e., so this would be ACL).

pmcb55 (approved):
I went through this pretty quick - but looks good.

src/acl/mock.internal.ts
NSeydoux: I can, but this goes back to the whole DRY/DAMP question: should be the constants used in code and the constants used in the tests reference the same object ? Because what we eventually might want to do is to no longer use the `constants.ts` file, and have all the terms defined in a library we depend on. At this point, isn't it better to have the constants local to the test files for a sanity check, in case the library ever introduces a breaking change that we might want to handle in `solid-client` rather than directly cascading it, possibly missing the effects on our own dependants ?

src/constants.ts
NSeydoux: @Vinnl I think there was a reason for this, but I'm not sure what it was.

src/access/wac.ts
```diff
@@ -0,0 +1,81 @@
...
+  // ACL into the boolean or undefined vision of ACP.
+  for (const [mode, accessMode] of Object.entries(access)) {
+    if (accessMode) {
+      // There is no convenient way to type `mode` as the keys to the Access interface
+      // @ts-ignore
+      universalAccess[mode] = accessMode;
+    }
+  }
+  return universalAccess;
+}
```
Vinnl: I think I'd prefer to just explicitly set the relevant modes, so as to ensure that future changes in either interface will be caught here.

And actually, maybe we want to define an interface for the WAC-specific subsets of the possible Access Modes that would also be used as the return value for `getAgentAccess`, e.g. something like this:

```typescript
type WacAccess = Required<Access> & (
  { controlRead: true, controlWrite: true } |
  { controlRead: false, controlWrite: false }
);
```

(`Required<>` indicates that `undefined` is not an option, and the disjunction that `controlRead` and `controlWrite` will always have the same value.)

And so this would be something like:

```suggestion
function aclAccessToUniversal(access: AclAccess): WacAccess {
  return {
    read: access.read === true,
    append: access.append === true,
    write: access.write === true,
    controlRead: access.control === true,
    controlWrite: access.control === true,
  } as WacAccess;
}
```

This would also remove the need for the `getDefaultAccess` function.

src/access/wac.ts
```diff
@@ -0,0 +1,81 @@
...
+ * Note that this only lists permissions granted to the given Agent individually,
+ * and will not exhaustively list modes the given Agent may have access to because
+ * they apply to everyone, or because they apply to the Agent through a group for
+ * instance.
+ *
+ * @param resource The URL of the Resource for which we want to list Access
+ * @param agent The Agent for which the Access is granted
+ * @param options Optional parameter `options.fetch`: An alternative `fetch` function to make the HTTP request, compatible with the browser-native [fetch API](https://developer.mozilla.org/docs/Web/API/WindowOrWorkerGlobalScope/fetch#parameters).
+ * @returns True for Access modes granted to the Agent, False for Access modes
+ * denied to the Agent, and undefined otherwise.
```
Vinnl, line 63: `undefined` is denied in WAC :)

```suggestion
 * denied to the Agent.
```

src/access/wac.ts
Vinnl: Oh hmm, in fact... Maybe `WacAccess` should have each mode be `undefined | true` instead? Since access that is not specified can still be granted via e.g. a group, so in that sense it's different from an actual Deny.

src/access/wac.test.ts
```diff
@@ -0,0 +1,383 @@
...
+  return Object.assign(dataset(), {
+    internal_resourceInfo: {
+      sourceIri: sourceIri,
+      isRawData: false,
+      linkedResources: {},
+      aclUrl: aclIri,
+    },
+  });
+}
+
```
Vinnl, line 54: It's a bit annoying given that you've already done all the work of setting this up this way, but... Since this really builds on top of the existing WAC functions, which are independently tested already, I'm wondering if we should not just re-use that API in our tests, and also the mock functions defined in `acl/mock.ts`? As in, there's no need for us to verify here that it really does construct the right type and structure of the datasets, so testing that here means more code to keep up-to-date if that internal implementation ever changes, without it helping us catch other regressions that we wouldn't have caught otherwise.

src/access/wac.test.ts
```diff
@@ -0,0 +1,383 @@
...
+  it("returns null if no ACL is accessible", async () => {
+    const mockFetch = jest
+      .fn(window.fetch)
+      // No resource ACL available...
+      .mockResolvedValueOnce(
+        mockResponse("", {
+          status: 404,
+          url: "https://some.pod/resource.acl",
+        })
+        // Link to the fallback ACL...
```
Vinnl, line 91: I'm guessing this might be due to Prettier, but I think this comment and the one below it got with the code following it, i.e. it should be one line down?

src/access/wac.test.ts
```diff
@@ -0,0 +1,383 @@
...
+    );
+    const result = getAgentAccess(resource, "https://some.pod/profile#agent", {
+      fetch: mockFetch,
+    });
+    await expect(result).resolves.toStrictEqual({
+      append: true,
+    });
+  });
+
+  it("returns an empty object if the actor isn't present", async () => {
```
Vinnl, line 270:
```suggestion
  it("returns an empty object if the Agent isn't present", async () => {
```

src/access/wac.ts
Vinnl: Ah, and then this should also remove the mention of "deny" rather than of undefined, I suppose.

src/constants.ts
Vinnl: Don't think we explicitly discussed this, but they're not generated and there's no need for them to STAND OUT SCREAMING OVER THE OTHER CODE IMHO :)

pmcb55: Being generated or not isn't the issue (and we are generating constants for all these values already, and publishing them to npmjs, but just not depending on those npmjs modules at the moment).
Anyway, upper-casing classes of constants has always been standard practice in Java, and has always seemed to make total sense there (i.e., I've never heard anyone complain about the screams :) !).
We could trivially switch to lower-case for JavaScript, but does it really have a best practice for 'classes containing constants'...??!? These class names will nearly always be the prefixes recommended by the vocabularies themselves, so they'll necessarily be short, very often acronyms, and not the most descriptive in themselves (e.g., FOAF, DCTerms, LDP, RDF, RDFS, ACL, WAC, etc.), so that seems (to me at least) to be further justification for leaving them as they are in uppercase (i.e., the screaming is actually justified).

src/acl/mock.internal.ts
pmcb55: No, I've never in my life seen where it would ever make sense to have test-constants and code-constants separated as a sanity check. The idea is that these values are defined by vocabulary owners outside our control - so yeah, we need a discussion about versioning vocabs, but that's a completely different topic. So yeah, I think having both our code and tests depending on external libraries is clearly where we want to go here (and not 'eventually' either, by 'soon' :) !).

src/access/wac.ts
NSeydoux: Exactly, that's the intent behind this: in the universal API, access modes are aligned with the ACP logic, namely true if explicitly granted, false if explicitly denied, and `undefined` if neither. That said, creating an explicit `WacAccess` type and only setting the corresponding types in `aclAccessToUniversal` makes perfect sense.

src/access/wac.ts
NSeydoux: Right, but since there's no notion of "deny" in ACL, either this becomes "True for Access modes granted to the Agent, and undefined otherwise.", or remains as it is (the idea being that this comment ends up applying at the public universal API level, rather than at the WAC level)

src/access/wac.ts
Vinnl: Right, since `WacAccess` is still just a subset of `Access`, i.e. `false` is not actually a valid value.

src/access/wac.ts
Vinnl: Yes, the universal API would get its own TSDoc, so I'd make sure that the doc here applies to what this specific function actually does (i.e. the former option).

src/access/wac.test.ts
NSeydoux: If I understand your suggestion correctly, that would mean doing things like 
```
it("fetches the resource ACL if available", async () => {
      
    let aclResource: AclDataset = Object.assign(getMockDataset("https://some.pod/resource.acl"), {
      internal_accessTo: "https://some.pod/resource"
    })
    aclResource = setAgentResourceAccess(aclResource, "https://some.pod/profile#agent", {
      read: true, append: false, write: false, control: false 
    });

    const mockFetch = jest.fn(window.fetch).mockResolvedValue(
      mockResponse(await triplesToTurtle(Array.from(aclResource)), {
        status: 200,
        url: "https://some.pod/resource.acl",
      })
    );

    const resource = getMockDataset(
      "https://some.pod/resource",
      "https://some.pod/resource.acl"
    );
    const result = getAgentAccess(resource, "https://some.pod/profile#agent", {
      fetch: mockFetch
    });

    await expect(result).resolves.toStrictEqual({
      read: true,
      append: undefined,
      write: undefined,
      controlRead: undefined,
      controlWrite: undefined,
    });
  });
```
If so, this was my initial approach, but `setAgentResourceAccess` and the like create blank nodes, which are usually resolved when saved in a remote dataset. However in this case since they are not (since we don't actually go and save them), and being blank nodes they aren't matched by `getAgentAccess` from `acl/agent`, which is the function we should use. Mocking the actual content of the ACL using the pre-existing mock function was easier.

However, I did replace the implementation of `getMockDataset ` with our own internal mocks, because I ironically had forgotten about them ^^

NSeydoux: See d7ae3fd3ff7fbd5782f5de373924b8db552c700b

src/access/wac.test.ts
NSeydoux: 280cf4d09004624662e0a440c419db280a687d7b

src/access/wac.test.ts
NSeydoux: 280cf4d09004624662e0a440c419db280a687d7b

src/access/wac.ts
NSeydoux: 280cf4d09004624662e0a440c419db280a687d7b

src/access/wac.ts
NSeydoux: 2c0f5a827dfe3efa5f09caa9263a3fd9adee67b0

src/acl/mock.ts
```diff
@@ -104,7 +104,7 @@ export function addMockFallbackAclTo<T extends WithServerResourceInfo>(
   return resourceWithFallbackAcl;
 }
 
-function setMockAclUrl<T extends WithServerResourceInfo>(
+export function setMockAclUrl<T extends WithServerResourceInfo>(
```
Vinnl, line 107: By exporting this function is becomes available to users of the library - I'm not sure if that's what we want to do? If not, it should be moved to `mock.internal.ts` (and prefixed with `internal_`).

src/access/wac.ts
```diff
@@ -0,0 +1,86 @@
...
+import { internal_fetchAcl } from "../acl/acl.internal";
+import { getAgentAccess as getAgentAccessWac } from "../acl/agent";
+import { WebId, WithServerResourceInfo } from "../interfaces";
+import { internal_defaultFetchOptions } from "../resource/resource";
+import { Access } from "./universal";
+import { Access as AclAccess } from "../acl/acl";
+
+// Setting WacAccess = Required<Access> enforces that any change on Access will
+// reflect on WacAccess, but WacAccess has additional restrictions.
+type WacAccess = Required<Access> &
```
Vinnl, line 31: `Required<>` is the opposite of `Partial<>`, i.e. it makes all properties _not_ be `undefined`. Which made sense when I thought they had to be `true` or `false` for WacAccess, but if they're `true` or `undefined`, less so :)

(In fact, I'm wondering whether the union type with Access is actually needed/might even override the other type (in the sense that setting a property of a `WacAccess` to `false` would still be allowed). I think as long as our universal API, which has just `Access` as a return type`, directly passes on the return value of the WAC-specific API, we're still safe in terms of getting a warning when either of those types changes.)

src/access/wac.ts
```diff
@@ -0,0 +1,86 @@
...
+  (
+    | { controlRead: true; controlWrite: true }
+    | ({ controlRead: undefined; controlWrite: undefined } & {
+        read: true | undefined;
+        append: true | undefined;
+        write: true | undefined;
+      })
```
Vinnl, line 38: I think the brackets here might be the wrong way around? I'd think it'd be something like `(<control true> | <control undefined>) & <read append write>` (but I might be getting my De Morgan's mixed up)

(Yes I looked "De Morgan's laws" up to sound clever.)

Vinnl (approved):
I'm not really sure about the `mock_` prefix (i.e. is that something that needs to stand out to the developer, and if so, shouldn't we need that for our other mock functions as well?), but overall, pending some changes to the types and the marking of an internal function as such, this looks good.

src/access/wac.ts
NSeydoux: Well it worked, because I looked it up as well, and also the brackets were messed up. 90ec90d4c1de96c4ef16c609e6047a7c951a8cb6

src/access/wac.ts
NSeydoux: Removed the union with Access: 90ec90d4c1de96c4ef16c609e6047a7c951a8cb6

src/acl/mock.ts
NSeydoux: Right, I have to be more careful about this. And correct me if I'm wrong, but once (/if eventually) eveybody switches to ESM, this will no longer be an issue, because all the imports go through the import map, right ?

NSeydoux: 889bdc0d5de67aa02fe3729ecda64032076d50a3

NSeydoux:
I moved the `mock_` prefix into a `Mock` component in the function name in 889bdc0d5de67aa02fe3729ecda64032076d50a3, which actually aligns with our practices for "shared" mocks. I'm not sure we enforce anything specific on locally defined mocks, but that sounds alright.

src/acl/mock.ts
Vinnl: Yes and no. Yes, we should no longer need the `internal_` prefix because developers can no longer just import functions from files not added to the exports map, but no, because this specific file _is_ added to our exports map, so any export in there will still be public.

PR #792, opened 2021-02-02
insert a couple of line breaks at 80 chars

pmcb55:
Just tweaking README slightly to break up long lines (to make it easier to review on a phone!).


NSeydoux (approved):

Vinnl:
As a note, Nicolas M aligned all the READMEs a while ago, so you might want to apply these same changes to our other projects.

PR #795, opened 2021-02-03
 Add getAgentAccessAll for WAC Resources

NSeydoux:
This adds `getAgentAccessAll`, which returns a list of all the
individual agents being granted access to a Resource which access is
controlled by a WAC server.

Note that since the WAC code is largely based on pre-existing functions,
the tests check for implementation rather than behaviour by verifying
that the underlying function is called, in order to reduce the number of
tests.

# Checklist

### DIFF
--- a/src/access/wac.ts
+++ b/src/access/wac.ts
@@ -20,7 +20,11 @@
  */
 
 import { internal_fetchAcl, internal_setAcl } from "../acl/acl.internal";
-import { getAgentAccess as getAgentAccessWac } from "../acl/agent";
+import {
+  AgentAccess,
+  getAgentAccess as getAgentAccessWac,
+  getAgentAccessAll as getAgentAccessAllWac,
+} from "../acl/agent";
 import { getGroupAccess as getGroupAccessWac } from "../acl/group";
 import { getPublicAccess as getPublicAccessWac } from "../acl/class";
 import {
@@ -44,6 +48,8 @@ type WacAccess = (
   append: true | undefined;
   write: true | undefined;
 };
+
+type AgentWacAccess = Record<WebId, WacAccess>;
 
 function aclAccessToUniversal(access: AclAccess): WacAccess {
   // In ACL, denying access to an actor is a notion that doesn't exist, so an
@@ -90,6 +96,25 @@ async function getActorClassAccess(
     return null;
   }
   return aclAccessToUniversal(wacAccess);
+}
+
+async function getActorAccessAll(
+  resource: WithServerResourceInfo,
+  accessEvaluationCallback: typeof getAgentAccessAllWac,
+  options: Partial<typeof internal_defaultFetchOptions>
+): Promise<AgentWacAccess | null> {
+  const resourceAcl = await internal_fetchAcl(resource, options);
+  const wacAgentAccess = accessEvaluationCallback(
+    internal_setAcl(resource, resourceAcl)
+  );
+  if (wacAgentAccess === null) {
+    return null;
+  }
+  const result: AgentWacAccess = {};
+  for (const [webId, wacAccess] of Object.entries(wacAgentAccess)) {
+    result[webId] = aclAccessToUniversal(wacAccess);
+  }
+  return result;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


NSeydoux:
@Vinnl, I realize only with this PR that testing for the underlying function to be called was what you suggested earlier, which I initially misunderstood. Is that correct ?

src/access/wac.ts
```diff
@@ -92,6 +98,25 @@ async function getActorClassAccess(
...
+): Promise<AgentWacAccess | null> {
+  const resourceAcl = await internal_fetchAcl(resource, options);
+  const wacAgentAccess = accessEvaluationCallback(
+    Object.assign(resource, { internal_acl: resourceAcl })
+  );
+  if (wacAgentAccess === null) {
+    return null;
+  }
+  const result: AgentWacAccess = {};
+  for (const [webId, wacAccess] of Object.entries(wacAgentAccess)) {
```
Vinnl, line 114: `Object.entries` [is not supported by IE](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries#browser_compatibility), so maybe you could mention this in the README among the list of polyfills as well?

src/access/wac.ts
```diff
@@ -162,3 +187,26 @@ export async function getPublicAccess(
...
+ * and will not exhaustively list modes any Agent may have access to because
+ * they apply to everyone, or because they apply to an Agent through a group for
+ * instance.
+ *
+ * @param resource The URL of the Resource for which we want to list Agents Access
+ * @param options Optional parameter `options.fetch`: An alternative `fetch` function to make the HTTP request, compatible with the browser-native [fetch API](https://developer.mozilla.org/docs/Web/API/WindowOrWorkerGlobalScope/fetch#parameters).
+ * @returns A map of Agents WebIDs associated to a list of modes: True for Access modes
+ * granted to the Agent, False for Access modes denied to the Agent, and undefined otherwise.
+ */
+export async function getAgentAccessAll(
```
Vinnl, line 205: Since we're not awaiting things. (This might also be relevant to the functions you added earlier; I didn't check.)

```suggestion
export function getAgentAccessAll(
```

src/access/wac.ts
```diff
@@ -162,3 +187,26 @@ export async function getPublicAccess(
...
+ *
+ * Note that this only lists permissions granted to each Agent individually,
+ * and will not exhaustively list modes any Agent may have access to because
+ * they apply to everyone, or because they apply to an Agent through a group for
+ * instance.
+ *
+ * @param resource The URL of the Resource for which we want to list Agents Access
+ * @param options Optional parameter `options.fetch`: An alternative `fetch` function to make the HTTP request, compatible with the browser-native [fetch API](https://developer.mozilla.org/docs/Web/API/WindowOrWorkerGlobalScope/fetch#parameters).
+ * @returns A map of Agents WebIDs associated to a list of modes: True for Access modes
+ * granted to the Agent, False for Access modes denied to the Agent, and undefined otherwise.
```
Vinnl, line 203:
```suggestion
 * granted to the Agent, and undefined otherwise.
```

Vinnl (approved):
@NSeydoux I think so, yes :) (Although this still seems to test on implementation quite a bit too?)

src/access/wac.ts
```diff
@@ -92,6 +98,25 @@ async function getActorClassAccess(
...
 }
 
+async function getActorAccessAll(
+  resource: WithServerResourceInfo,
+  accessEvaluationCallback: typeof getAgentAccessAllWac,
+  options: Partial<typeof internal_defaultFetchOptions>
+): Promise<AgentWacAccess | null> {
+  const resourceAcl = await internal_fetchAcl(resource, options);
+  const wacAgentAccess = accessEvaluationCallback(
+    Object.assign(resource, { internal_acl: resourceAcl })
```
Vinnl, line 108: Ah also, maybe first merge #797 and then update this :)

src/access/wac.ts
NSeydoux: 4245918a0f69233fb750a25120c2a5a8317d2a3f

src/access/wac.ts
NSeydoux: Thanks for that, I didn't realize that could be an issue : see https://twitter.com/_jayphelps/status/1324565522755788803

NSeydoux: 975cb80ffc8c80844480922d3dcf4919bad72ade

src/access/wac.ts
NSeydoux: 975cb80ffc8c80844480922d3dcf4919bad72ade updates several occurences

PR #796, opened 2021-02-03
Fix misdetecting of ACRs that could not be fetched

Vinnl:
Detecting whether an ACR was present was done by checking if it's type was `object`, as expected. But of course, `null` is an object too :sob: 

(I also fixed the name of a deprecated function, just in case.)

### DIFF
--- a/src/acp/control.test.ts
+++ b/src/acp/control.test.ts
@@ -22,6 +22,7 @@
 import { describe, it, expect } from "@jest/globals";
 
 import {
+  AccessControlResource,
   acrAsMarkdown,
   addAcrPolicyUrl,
   addMemberAcrPolicyUrl,
@@ -244,6 +245,17 @@ describe("getControlAll", () => {
 
   it("throws an error if the given Resource does not have an Access Control Resource", () => {
     const withoutAcr = mockSolidDatasetFrom("https://some.pod/resource");
+
+    expect(() => internal_getControlAll(withoutAcr as any)).toThrow(
+      "Cannot work with Access Controls on a Resource (https://some.pod/resource) that does not have an Access Control Resource."
+    );
+  });
+
+  it("throws an error if the given Resource's ACR could not be fetched", () => {
+    const withoutAcr = addMockAcrTo(
+      mockSolidDatasetFrom("https://some.pod/resource"),
+      (null as unknown) as AccessControlResource
+    );
 
     expect(() => internal_getControlAll(withoutAcr as any)).toThrow(
       "Cannot work with Access Controls on a Resource (https://some.pod/resource) that does not have an Access Control Resource."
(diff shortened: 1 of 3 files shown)


NSeydoux (approved):
What a magnificent language feature

src/acp/control.test.ts
```diff
@@ -249,6 +250,17 @@ describe("getControlAll", () => {
...
   });
+
+  it("throws an error if the given Resource's ACR could not be fetched", () => {
+    const withoutAcr = addMockAcrTo(
+      mockSolidDatasetFrom("https://some.pod/resource"),
+      (null as unknown) as AccessControlResource
+    );
+
+    expect(() => internal_getControlAll(withoutAcr as any)).toThrow(
+      "Cannot work with Access Controls on a Resource (https://some.pod/resource) that does not have an Access Control Resource."
```
pmcb55, line 261:
```suggestion
      "Cannot work with Access Controls on a Resource [https://some.pod/resource] that does not have an Access Control Resource."
```

Vinnl: This sentence is present in all of the existing tests, so I can fix this in a separate PR. But before I do that: does this make sense here? Since the URL in this case is not the output or e.g. a variable, but rather a reference to the preconditions we defined in our test?

pmcb55: Not sure I follow the question, but the exception string is thrown from within the code under test, right? And I assume that that exception message **_does_** contain a variable, i.e., it's something like:
```
throw new Error(`Cannot work with Access Controls on a Resource [${resourceIRI}] that does not have an Access Control Resource.`);
```
I don't think it matters where the variable comes from - the code-under-test uses a variable, and so we delimit them in messages with '[' and ']'. Unless I'm missing something silly...?

Vinnl: Oh heh sorry, not it was me that was reading this wrong - I thought this was the message that was thrown if the `expect` didn't match.

Either way, the sentence still needs the brackets, so does that mean it would become

    Cannot work with Access Controls on a Resource ([https://some.pod/resource]) that does not have an Access Control Resource.

?

pmcb55: :) no worries. I'd just adjust the English then, to something like:
```
Cannot retrieve the Access Controls for the resource [https://some.pod/resource], as it does not have an Access Control Resource (ACR).
```
I'm not sure if that's giving away 'sensitive' information about that resource - i.e., the fact that it doesn't have an ACR, as opposed to the fact that maybe I'm just not authorized to see that ACR. You'll know the context better, but regardless the English would be something along the lines of: "Failed to retrieve/update/delete Access Data for resource [X] - reasons could be: A, B or C".

Vinnl: https://github.com/inrupt/solid-client-js/pull/811

PR #797, opened 2021-02-03
Centralize ACP assignment

NSeydoux:
This adds an internal function that sets the `internal_acl` property of
a given Dataset, which is our internal ACP representation. This is a
simple refactoring to avoid explicitly using the property name all over
the code, but to isolate it in one function and the associated type
annotations.

### DIFF
--- a/src/acl/acl.internal.ts
+++ b/src/acl/acl.internal.ts
@@ -40,6 +40,8 @@ import {
   hasAccessibleAcl,
   WithAccessibleAcl,
   WithAcl,
+  WithFallbackAcl,
+  WithResourceAcl,
 } from "./acl";
 
 /**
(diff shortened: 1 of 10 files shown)


src/acl/acl.internal.ts
```diff
@@ -463,3 +465,24 @@ export function internal_duplicateAclRule(sourceRule: AclRule): AclRule {
...
+>(
+  resource: ResourceExt,
+  acl: WithResourceAcl["internal_acl"]
+): ResourceExt & WithResourceAcl;
+export function internal_setResourceAcl<
+  ResourceExt extends WithServerResourceInfo
+>(
+  resource: ResourceExt,
+  acl: WithFallbackAcl["internal_acl"]
+): ResourceExt & WithFallbackAcl;
```
Vinnl, line 480: Given that we've so far been referring to a Resource's parent's ACL as "fallback ACL" and a Resource's own ACL as a "Resource ACL", and given that this function accepts both, should we just call it `internal_setAcl`?

Vinnl (approved):
Good improvement! Of course I had to make a bit of a fuzz over the name, but a clear improvement nonetheless :)

src/acl/acl.internal.ts
NSeydoux: I wasn't really happy with the previous name for that reason, so very good suggestion 03346bc8be43d54a698f05321c203c4160d73b6f

PR #799, opened 2021-02-04
Empty Rules no longer prevent a Policy from applying

Vinnl:
Apparently if a Policy points to an empty/non-existent Rule, that Policy still applies (unless of course prevented by other Rules). This is now reflected in our code too. (Also, Policies that do not specify access modes are properly ignored now too - this will be relevant for `setActorAccess`.)

Additionally, I've extracted all references to actorRelations into a single array listing all of them (just two at the time), to make sure that, when we need to take into account additional potential actorRelations, we update all relevant code. This will be especially relevant with `setActorAccess`, which will make copies of Policies and Rules, and therefore needs to make sure it copies over all relevant relations.

### DIFF
--- a/src/access/acp.test.ts
+++ b/src/access/acp.test.ts
@@ -21,8 +21,16 @@
 
 import { describe, it } from "@jest/globals";
 import { WithAccessibleAcr } from "../acp/acp";
-import { AccessControlResource, addPolicyUrl } from "../acp/control";
-import { internal_createControl } from "../acp/control.internal";
+import {
+  AccessControlResource,
+  addPolicyUrl,
+  getAcrPolicyUrlAll,
+  getPolicyUrlAll,
+} from "../acp/control";
+import {
+  internal_createControl,
+  internal_getAcr,
+} from "../acp/control.internal";
 import { addMockAcrTo } from "../acp/mock";
 import {
   createPolicy,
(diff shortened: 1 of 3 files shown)


src/access/acp.test.ts
```diff
@@ -1032,6 +1056,73 @@ describe("getActorAccess", () => {
...
+        memberAcrPolicies: {},
+      }
+    );
+
+    const access = internal_getActorAccess(resourceWithAcr, acp.agent, webId);
+
+    expect(access).toStrictEqual({
+      read: true,
+    });
+  });
```
NSeydoux, line 1083: I'm not sure that's what I understand from the latest discussions : 
> upon evaluation a policy will only be considered if it has at least one valid rule. And a rule is only considered valid if it has at least one valid triple aside from its type triple.

So in this case, shouldn't the previous behaviour prevail, and the policy not apply for its lack of valid rules ?

Vinnl: I added this when Emmet said that the behaviour implemented here is what should be needed. After that the discussion was re-opened and hasn't yet reached a conclusion, so when it does we might indeed need to revisit this. Unfortunately the discussions were in a private chat channel (because they're still preparing it for the official proposal) so I can't link them here, but I'll send the links to you.

Vinnl:
@NSeydoux Are you OK with merging this while we await a definitive decision on the desired behaviour? Since #800 builds on this.

Vinnl:
This was merged as part of #800.

PR #807, opened 2021-02-08
Add setAgentResourceAccess for a WAC resource

NSeydoux:
This adds setAgentResourceAccess to the WAC API. This function sets access for a
given agent and a given resource if the resource is controlled using
WAC. This only applies to Resource ACLs (as opposed to fallback ACLs),
updating it

tested for in that section of the code base,  and duplicating all these

NOTE: `setGroupAccess`, `setPublicAccess` etc will be added in upcoming PRs, with some code factored out of `setAgentAccess` into `setActorAccess`.

# Checklist

### DIFF
--- a/src/access/wac.test.ts
+++ b/src/access/wac.test.ts
@@ -2202,11 +2202,17 @@ describe("setAgentAccess", () => {
 
     const wacModule = jest.requireActual("../acl/agent") as {
       setAgentResourceAccess: () => Promise<AgentAccess>;
+    };
+    const aclModule = jest.requireActual("../acl/acl") as {
       saveAclFor: () => Promise<AclDataset>;
     };
     const setAgentResourceAccessWac = jest.spyOn(
       wacModule,
       "setAgentResourceAccess"
+    );
+    const saveAclForWac = jest.spyOn(
+      aclModule,
+      "saveAclFor"
     );
 
     const resource = getMockDataset(
@@ -2228,5 +2234,6 @@ describe("setAgentAccess", () => {
       }
     );
     expect(setAgentResourceAccessWac).toHaveBeenCalled();
+    expect(saveAclForWac).toHaveBeenCalled();
   });
 });
--- a/src/access/wac.test.ts
+++ b/src/access/wac.test.ts
@@ -29,6 +29,7 @@ import {
   getGroupAccessAll,
   getPublicAccess,
   setAgentResourceAccess,
+  WacAccess,
 } from "./wac";
 import { Response } from "cross-fetch";
 import { triplesToTurtle } from "../formats/turtle";
@@ -2119,6 +2120,21 @@ describe("setAgentAccess", () => {
       write: false,
       control: false,
     });
+  });
+
+  it("throws if the access being set can't be expressed in WAC", async () => {
+    const resource = getMockDataset(
+      "https://some.pod/resource",
+      "https://some.pod/resource.acl"
+    );
+    await expect(
+      setAgentResourceAccess(resource, "https://some.pod/profile#agent", ({
+        controlRead: true,
+        controlWrite: undefined,
+      } as unknown) as WacAccess)
+    ).rejects.toThrow(
+      "For WAC resources, controlRead and controlWrite must be equal."
+    );
   });
 
   it("calls the underlying setAgentAccess function", async () => {
--- a/src/access/wac.test.ts
+++ b/src/access/wac.test.ts
@@ -2210,10 +2210,7 @@ describe("setAgentAccess", () => {
       wacModule,
       "setAgentResourceAccess"
     );
-    const saveAclForWac = jest.spyOn(
-      aclModule,
-      "saveAclFor"
-    );
+    const saveAclForWac = jest.spyOn(aclModule, "saveAclFor");
 
     const resource = getMockDataset(
       "https://some.pod/resource",
--- a/src/access/wac.test.ts
+++ b/src/access/wac.test.ts
@@ -50,7 +50,7 @@ jest.mock("../fetcher.ts", () => ({
 import { mockSolidDatasetFrom } from "../resource/mock";
 import { AgentAccess } from "../acl/agent";
 import { internal_getResourceAcl } from "../acl/acl.internal";
-import { Access } from "../acl/acl";
+import { Access, AclDataset } from "../acl/acl";
 
 function getMockDataset(
   sourceIri: IriString,
@@ -2009,6 +2009,13 @@ describe("setAgentAccess", () => {
           status: 200,
           url: "https://some.pod/.acl",
         })
+      )
+      // Save the ACL
+      .mockResolvedValueOnce(
+        mockResponse("", {
+          status: 201,
+          url: "https://some.pod/.acl",
+        })
       );
 
     const resource = getMockDataset(
@@ -2137,7 +2144,55 @@ describe("setAgentAccess", () => {
     );
   });
 
-  it("calls the underlying setAgentAccess function", async () => {
+  it("returns null if the ACL cannot be written back", async () => {
+    const aclResource = addMockAclRuleQuads(
+      getMockDataset("https://some.pod/resource.acl"),
+      "https://some.pod/profile#agent",
+      "https://some.pod/resource",
+      { read: false, append: false, write: false, control: false },
+      "resource"
+    );
+
+    const mockFetch = jest
+      .fn(window.fetch)
+      // The resource ACL is available...
+      .mockResolvedValueOnce(
+        mockResponse(await triplesToTurtle(Array.from(aclResource)), {
+          status: 200,
+          url: "https://some.pod/resource.acl",
+        })
+      )
+      // Cannot save the ACL
+      .mockResolvedValueOnce(
+        mockResponse("Not authorized", {
+          status: 403,
+          url: "https://some.pod/.acl",
+        })
+      );
+
+    const resource = getMockDataset(
+      "https://some.pod/resource",
+      "https://some.pod/resource.acl"
+    );
+
+    const result = await setAgentResourceAccess(
+      resource,
+      "https://some.pod/profile#agent",
+      {
+        read: true,
+        append: undefined,
+        write: undefined,
+        controlRead: undefined,
+        controlWrite: undefined,
+      },
+      {
+        fetch: mockFetch,
+      }
+    );
+    expect(result).toBeNull();
+  });
+
+  it("calls the underlying WAC API functions", async () => {
     const mockFetch = jest.fn(window.fetch).mockResolvedValue(
       mockResponse("", {
         status: 200,
@@ -2147,6 +2202,7 @@ describe("setAgentAccess", () => {
 
     const wacModule = jest.requireActual("../acl/agent") as {
       setAgentResourceAccess: () => Promise<AgentAccess>;
+      saveAclFor: () => Promise<AclDataset>;
     };
     const setAgentResourceAccessWac = jest.spyOn(
       wacModule,
@@ -2171,7 +2227,6 @@ describe("setAgentAccess", () => {
         fetch: mockFetch,
       }
     );
-
     expect(setAgentResourceAccessWac).toHaveBeenCalled();
   });
 });
--- a/src/access/wac.test.ts
+++ b/src/access/wac.test.ts
@@ -21,12 +21,14 @@
 
 import { jest, describe, it } from "@jest/globals";
 import { IriString, SolidDataset, WithServerResourceInfo } from "../interfaces";
+import { getAgentResourceAccess } from "../acl/agent";
 import {
   getAgentAccess,
   getAgentAccessAll,
   getGroupAccess,
   getGroupAccessAll,
   getPublicAccess,
+  setAgentResourceAccess,
 } from "./wac";
 import { Response } from "cross-fetch";
 import { triplesToTurtle } from "../formats/turtle";
@@ -46,6 +48,8 @@ jest.mock("../fetcher.ts", () => ({
 
 import { mockSolidDatasetFrom } from "../resource/mock";
 import { AgentAccess } from "../acl/agent";
+import { internal_getResourceAcl } from "../acl/acl.internal";
+import { Access } from "../acl/acl";
 
 function getMockDataset(
   sourceIri: IriString,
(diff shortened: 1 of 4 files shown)


src/acl/acl.internal.ts
```diff
@@ -491,3 +496,19 @@ export function internal_setAcl<ResourceExt extends WithServerResourceInfo>(
 ): ResourceExt & WithAcl {
   return Object.assign(resource, { internal_acl: acl });
 }
+
+export function internal_setResourceAcl<
+  T extends WithServerResourceInfo & WithAcl
+>(resource: T, acl: AclDataset): T & WithResourceAcl {
+  const newAcl: WithResourceAcl["internal_acl"] = {
+    resourceAcl: acl,
+    fallbackAcl: null,
```
Vinnl, line 505: Just wondering if maybe we should check whether a fallback ACL is available for `resource` and, if so, preserve that?

Vinnl (changes requested):
If I understood it correctly I think this doesn't actually save the changed ACL, which is kinda essential :)

(And I think we should warn as soon as we get different `control` values, rather than pretending that everything worked.)

src/access/wac.ts
```diff
@@ -68,6 +83,20 @@ function aclAccessToUniversal(access: AclAccess): WacAccess {
   } as WacAccess;
 }
 
+function universalAccessToAcl(access: Partial<WacAccess>): AclAccess {
```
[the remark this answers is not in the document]
NSeydoux: That makes it clearer: I wasn't sure there was a difference between `{someProperty: undefined}` and `{someproperty?: undefined}`, but now I get it. Fixed in 57fd4406e6fb9ba005495ea006d630a373d887b5

src/access/wac.ts
```diff
@@ -68,6 +83,20 @@ function aclAccessToUniversal(access: AclAccess): WacAccess {
...
+  // Universal access is aligned on ACP, which means there is a distinction between
+  // controlRead and controlWrite. This split doesn't exist in WAC, which is why
+  // the type for the input variable of this function is a restriction on the
+  // universal Access type. Also, in WAC, an undefined and false Access modes are
+  // equivalent.
+  return {
+    read: access.read ? true : false,
+    append: access.append ? true : false,
+    write: access.write ? true : false,
+    control: access.controlRead ? true : false,
```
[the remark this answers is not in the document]
NSeydoux: The type definition enforced that, but it only applies to users using TS, so it's not a bad thing to add a defensive check here. Fixed in 57fd4406e6fb9ba005495ea006d630a373d887b5.

src/access/wac.ts
```diff
@@ -235,3 +264,52 @@ export function getGroupAccessAll(
...
   return getActorAccessAll(resource, getGroupAccessAllWac, options);
 }
+
+/**
+ * Set the Access modes for a given Agent to a given Resource.
+ *
+ * Important note: if the target resource did not have a Resource ACL, and its
+ * Access was regulated by its Fallback ACL, said Fallback ACL is copied to create
+ * a new Resource ACL. This has the side effect that the next time the Fallback
+ * ACL is updated, the changes WILL NOT IMPACT the target resource.
```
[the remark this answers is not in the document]
NSeydoux: Fixed in c6427b709730448fd8b03b519fb41b19400a5c11

src/access/wac.ts
```diff
@@ -235,3 +264,52 @@ export function getGroupAccessAll(
...
+ *
```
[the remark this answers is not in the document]
NSeydoux: The `@returns` comment somewhat mentioned it, but the documentation is more detailed with 67d47f10559f69874ce57cf3a42a028c35f93fc0

src/access/wac.ts
```diff
@@ -235,3 +264,52 @@ export function getGroupAccessAll(
...
+    // This probably isn't possible, because otherwise hasAccessibleAcl is false
+    // and the function already returned, so probably exclude from coverage.
+    return null;
+  }
+  const updatedResourceAcl = setAgentResourceAccessWac(
+    resourceAcl,
+    agent,
+    universalAccessToAcl(access)
+  );
+  return internal_setResourceAcl(resourceWithAcl, updatedResourceAcl);
```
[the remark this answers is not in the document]
NSeydoux: For some reason, I assumed the ACL would be saved when saving the resource, which is absolutely not the case. I fixed it with ab7ef22a653da203fac5beb3ff279bce166e99ac.

However, this got me thinking: would it be possible to only fetch the ACL if the provided resource didn't have one attached locally, and to only save the ACL when the resource is saved ? I guess it would add risk for conflict, or prevent one from realizing the user doesn't have access to the ACL, so I guess the current approach of always fetching and writing the ACL back is safer.

src/acl/acl.internal.ts
NSeydoux: That was my initial though too, but the type definition actually prevents it: 
```
export type WithAcl = {
  internal_acl:
    | {
        resourceAcl: AclDataset;
        fallbackAcl: null;
      }
    | {
        resourceAcl: null;
        fallbackAcl: AclDataset | null;
      };
};
```

Which makes sense: as soon as a Resource has a Resource ACL, its Fallback ACL becomes irrelevant, so I guess overriding it isn't too much of an issue. Otherwise, we can update the type definition to allow both to coexist.

src/access/wac.ts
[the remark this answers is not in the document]
Vinnl: Hmm, only fetching it if it didn't have one sounds like a good idea and doable - we can just do a `hasResourceAcl` and `hasFallbackAcl` before fetching it, I'd presume?

Saving it is more problematic as long as we don't have atomic updates (which I'm starting to think we might have optimised for too early in our API design, given that it doesn't seem like that's actually coming any time soon, and that it makes things like this a bit awkward), because it's not clear what to do if one of the two updates failed.

src/acl/acl.internal.ts
Vinnl: Oh that's right, we changed it to only fetch the fallback ACL if the Resource ACL was not available, as indeed it's not relevant if there is a Resource ACL. Never mind then!

src/access/wac.ts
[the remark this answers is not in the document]
Vinnl: Haha, I saw the commit message - it would actually be hilarious if screen readers would actually SHOUT THIS at people. But I'm not sure if the actual behaviour if any better: it thinks it's an acronym, and will spell it out letter by letter :P

src/access/wac.ts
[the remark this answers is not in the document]
Vinnl: Note that that was my interpretation of what happened - I'm not sure if it's actually correct :)

src/access/wac.test.ts
```diff
@@ -1670,3 +1675,558 @@ describe("getGroupAccessAll", () => {
...
+        append: true,
+        write: undefined,
+        controlRead: undefined,
+        controlWrite: undefined,
+      },
+      {
+        fetch: mockFetch,
+      }
+    );
+    expect(setAgentResourceAccessWac).toHaveBeenCalled();
```
Vinnl, line 2230: A check for `saveAclFor` would also be appropriate here, I think?

Vinnl (approved):

src/access/wac.test.ts
NSeydoux: Oops, I initially half-added this test, then added another test to check that if the save failed we returned `null` and decided the former was no longer useful, but did not remove it properly ^^. But yeah, only testing the failure case isn't good either. Fixed in 59e2bf3513e8fe318de5c284bbd870dfae3e86fa.

src/access/wac.ts
```diff
@@ -68,6 +84,25 @@ function aclAccessToUniversal(access: AclAccess): WacAccess {
...
+  if (access.controlRead !== access.controlWrite) {
+    throw new Error(
+      "For WAC resources, controlRead and controlWrite must be equal."
+    );
+  }
+  return {
+    read: access.read ? true : false,
+    append: access.append ? true : false,
+    write: access.write ? true : false,
+    control: access.controlRead ? true : false,
```
Vinnl, line 102: Actually, I just realised something: when you pass `false` for a property to `setAgentResourceAccessWac`, it does remove existing statements that give the user access. So I think what we should be doing here is allow for both `undefined` and a `boolean`, and have `undefined` result in whatever the existing value is. "Denying" access has somewhat different semantics than it has with ACPs, but it's still distinct from leaving things unchanged.

(And actually, that makes me realise that the ACP API might be a bit troublesome either: although you can e.g. give someone read access, you can't _remove_ that read access anymore - you can only _deny_ read access. This might be more than you want. So I'm wondering if maybe we should align the ACP API with the WAC one, and have a value of `false` only remove existing `acp:allow` values, rather than adding `acp:deny` ones?

Vinnl (changes requested):
Sorry, just realised something potentially problematic, that might have repercussions for the ACP API as well :/

src/access/wac.ts
NSeydoux: So that means the API could have a `setAuthorizedAgentAccess` and `setDeniedAgentAccess`, each accepting modes with `true` to set it, `false` to remove it, and `undefined` to leave it as it is. An alternative would be to change `Access` to become something like
```
type Access = {
  read: "add" | "remove" | "deny" | undefined
  write: ...
}
```
which would result in not splitting the access-related functions between authorize and deny in the API, but looks less convenient to use.

Vinnl: Yeah that would be cumbersome. The "problem" with adding separate allow and deny functions, is that the latter doesn't work for WAC at all, so it's not really universal. So maybe we should just change `setAgentAccess`'s behaviour to what you called `setAuthorizedAgentAccess`? (Which is kinda confusing given that we also have `setAuthorizedAccess` :) )

NSeydoux: Oh dear I didn't think about `setAuthorizedAccess`.

And actually even for ACP only having `setAgentAccess` is limiting: if I `setAgentAccess(resource, agent, {read:false})`, do I intend to deny access to the agent, or to remove read access granted individually to the agent if it exists, but they could still have read access through a Group (which isn't the case if the access is denied I think) ? While if we have `setAgentAuthorizedAccess` and `setAgentDeniedAccess` (name WIP), the latter approach is applied more explicitly ? And when talking to an ACL backend, `setAgentDeniedAccess` would simply apply a reverse logic compared to `setAgentAuthorizedAccess`: remove access modes marked `true`, add triples marked `false`; and leave the undefined ones.

Vinnl: > `setAgentDeniedAccess` would simply apply a reverse logic compared to `setAgentAuthorizedAccess`: remove access modes marked true, add triples marked false; and leave the undefined ones.

What do you mean by "add triples marked false"? That sounds like specifying deny, which isn't possible in WAC?

I'd think you could simply pass e.g. `read: true` to `setAgentAccess` which _adds_ an allow statement, `read: false` to remove any existing ones, and `undefined` to leave existing ones be. That is also the current behaviour of the WAC-specific API, IIRC.

NSeydoux: > >  if I setAgentAccess(resource, agent, {read:false}), do I intend to deny access to the agent, or to remove read access granted individually to the agent if it exists, but they could still have read access through a Group (which isn't the case if the access is denied I think) ?

> It would remove read access granted explicitly to the given agent, and they could indeed still have read access through a group. We explicitly make specify the caveat that we cannot guarantee effective access, because there is no way for us to do that :)

If I follow, `setAgentAccess` (from the universal API) would add specified access on `true`, remove specified access on `false`, and leave the given access untouched on `undefined`. That means that effectively, we don't support denying access through the universal API. 

> > setAgentDeniedAccess would simply apply a reverse logic compared to setAgentAuthorizedAccess: remove access modes marked true, add triples marked false; and leave the undefined ones.

> What do you mean by "add triples marked false"? That sounds like specifying deny, which isn't possible in WAC?
I'd think you could simply pass e.g. read: true to setAgentAccess which adds an allow statement, read: false to remove any existing ones, and undefined to leave existing ones be. That is also the current behaviour of the WAC-specific API, IIRC.

What I meant is in WAC, basically removing a granted access and denying access are two equivalent things. Therefore, if I `setAgentDeniedAccess(resource, agent, {read: true})`, in the case of an ACL backend that would simply remove any triples giving Read access to the Resource for the Agent, being strictly equivalent to `setAgentAuthorizedAccess(resource, agent, {read: false})`. In the case of an ACP backend though, these two statements would not have a similar effect on the effective access the Agent gets to the Resource.

I'm happy to have `setAgentAccess` of the first proposal behave the same as `setAgentAuthorizedAccess` of the second proposal, but then I'm not sure how `deny` access could ever be set.

Vinnl: Right, so indeed my proposal would be to not allow setting `deny` access using the universal API, because that functionality is not common to all access systems we support - just like we don't support setting Creator access using this API.

So we basically have two options:
- Only support allow (`true`), unallow (`false`) and leave the same (`undefined`) - i.e. no deny.
- Have allow- and deny-specific functions, where the former is as above, and the latter supports deny (`true`), undeny (`false`) and leave the same (`undefined`) in the case of ACP, and unallow (`true`), allow (`false`) and leave the same (`undefined`) in the case of WAC.

As mentioned I'm inclined to the former, but not wedded to it and see the point of the latter. Should be bring this up in the team?

NSeydoux: Whichever approach we chose, 7afa637110544b89b9f12c33f4cff175a7f863e2 fixes the current behaviour and accepts `boolean | undefined` for each mode.

NSeydoux: For the record, after some discussion, we'll be going with the 'no deny' approach to keep the universal API as universal as possible, and limit the addition of features that are access control system-specific.

src/access/wac.ts
```diff
@@ -38,20 +45,97 @@ import {
...
+): AclAccess;
+function universalAccessToAcl(
+  newAccess: WacAccess,
+  previousAccess?: AclAccess
+): AclAccess {
+  // Universal access is aligned on ACP, which means there is a distinction between
+  // controlRead and controlWrite. This split doesn't exist in WAC, which is why
+  // the type for the input variable of this function is a restriction on the
+  // universal Access type. Also, in WAC, an undefined and false Access modes are
+  // equivalent.
```
Vinnl, line 111: This last sentence is no longer true, is it? `undefined` leaves things as-is, `false` removes existing access?

src/access/wac.ts
```diff
@@ -38,20 +45,97 @@ import {
...
+  if (newAccess.controlRead !== newAccess.controlWrite) {
+    throw new Error(
+      "For WAC resources, controlRead and controlWrite must be equal."
```
Vinnl, line 114: Since the access system is not specific to the one Resource :) (And to spell out the term people need to google for :P)

```suggestion
      "For Pods using Web Access Control, controlRead and controlWrite must be equal."
```

src/access/wac.ts
```diff
@@ -235,3 +319,92 @@ export function getGroupAccessAll(
 ): Promise<AgentWacAccess | null> {
   return getActorAccessAll(resource, getGroupAccessAllWac, options);
 }
+
+async function setActorResourceAccess<T extends WithServerResourceInfo>(
```
Vinnl, line 323: Kinda nitpick since this is an internal API anyway, but since we're not really distinguishing between Resource and fallback ACLs, I'd prefer to keep these function names aligned with the ACP ones, i.e. just `setActorAccess`.

src/access/wac.ts
```diff
@@ -38,20 +45,97 @@ import {
...
+        control: newAccess.controlRead ?? previousAccess!.control,
+      };
+}
+
+const NO_ACCESS: AclAccess = {
+  read: false,
+  write: false,
+  append: false,
+  control: false,
+};
```
Vinnl, line 138: Unused variable.

```suggestion
```

src/access/wac.ts
```diff
@@ -38,20 +45,97 @@ import {
...
+  );
+}
+
+function universalAccessToAcl(
+  newAccess: WacAccess,
+  previousAccess: AclAccess
+): AclAccess;
+function universalAccessToAcl(
+  newAccess: NoUndefinedWacAccess,
+  previousAccess: undefined
```
Vinnl, line 101: Rou're still calling it with an explicit `undefined` value below, which seems kinda unnecessary.

```suggestion
```

src/access/wac.ts
```diff
@@ -235,3 +319,92 @@ export function getGroupAccessAll(
...
+  }
+
+  const resourceWithOldAcl = internal_setResourceAcl(
+    resourceWithAcl,
+    resourceAcl
+  );
+  let wacAccess: AclAccess | undefined = undefined;
+  if (hasNoUndefinedAccessModes(access)) {
+    // If all the access modes in `access` are set, no need for the previous access.
+    wacAccess = universalAccessToAcl(access, undefined);
```
Vinnl, line 355: You shouldn't need this if you update the overloads as above. That said, might it not be easier/more readable to just remove `hasNoUndefinedAccessModes` and its related types altogether, and just always run `readAccess` anyway?

```suggestion
    wacAccess = universalAccessToAcl(access);
```

src/access/wac.test.ts
```diff
@@ -1670,3 +1675,703 @@ describe("getGroupAccessAll", () => {
...
+
+    expect(newAccess).toStrictEqual({
+      read: true,
+      append: true,
+      write: false,
+      control: false,
+    });
+  });
+
+  it("removes access previoulsy granted if the new access is set to false", async () => {
```
Vinnl, line 2009:
```suggestion
  it("removes access previously granted if the new access is set to false", async () => {
```

Vinnl (approved):
Added some teeny tiny refactoring suggestions, but otherwise this appears to do what I'd expect :tada:

src/access/wac.ts
```diff
@@ -235,3 +319,92 @@ export function getGroupAccessAll(
...
+ * fallback ACL if it does not have a Resource ACL, then `null` is returned.
+ *
+ * @param resource The Resource for which Access is being set
+ * @param agent The Agent for whom Access is being set
+ * @param access The Access being set
+ * @param options Optional parameter `options.fetch`: An alternative `fetch` function to make the HTTP request, compatible with the browser-native [fetch API](https://developer.mozilla.org/docs/Web/API/WindowOrWorkerGlobalScope/fetch#parameters).
+ * @returns The Resource, with its ACL updated, or null if the new Access could not
+ * be set.
+ */
+export async function setAgentResourceAccess<T extends WithServerResourceInfo>(
```
Vinnl, line 394: Ah and so likewise I'd call this

```suggestion
export async function setAgentAccess<T extends WithServerResourceInfo>(
```

NSeydoux: I agree for `setActorAccess`, which is internal and does handle both resource and default access, but I was thinking of adding https://github.com/inrupt/solid-client-js/pull/823 to the WAC API too

NSeydoux: Actually, that raises the question of adding `get*DefaultAccess`, which isn't part of the API yet.

Vinnl: Hmm I see. I suppose we should then also consider whether we should be adding something similar to the high-level ACP API, although then the challenge is that they both have somewhat different semantics viz a viz applying to children (i.e. in WAC, it only applies to children without their own ACL, whereas in ACPs it's retroactively applied to all children).

src/access/wac.ts
NSeydoux: cd04f8cfa1c7e0e33e292cd7c1f84b336b12c92f

src/access/wac.ts
NSeydoux: 6156b10f217e8dd7a323ebb0c72662878756f28c

src/access/wac.ts
NSeydoux: dbaafa53cb91f1e4d0b853f2bcbd36859739cda6

PR #822, opened 2021-02-15
Add setGroupResourceAccess

NSeydoux:
This adds setGroupResourceAccess to the WAC API. This function sets access for a
given group and a given resource if the resource is controlled using
WAC. This only applies to Resource ACLs (as opposed to fallback ACLs),
updating it

ested for in that section of the code base,  and duplicating all these

# Checklist

### DIFF
--- a/src/access/wac.test.ts
+++ b/src/access/wac.test.ts
@@ -29,6 +29,7 @@ import {
   getGroupAccessAll,
   getPublicAccess,
   setAgentResourceAccess,
+  setGroupResourceAccess,
   WacAccess,
 } from "./wac";
 import { Response } from "cross-fetch";
@@ -51,6 +52,7 @@ import { mockSolidDatasetFrom } from "../resource/mock";
 import { AgentAccess } from "../acl/agent";
 import { internal_getResourceAcl } from "../acl/acl.internal";
 import { Access, AclDataset } from "../acl/acl";
+import { getGroupResourceAccess } from "../acl/group";
 
 function getMockDataset(
   sourceIri: IriString,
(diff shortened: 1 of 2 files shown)


src/access/wac.test.ts
```diff
@@ -2375,3 +2377,714 @@ describe("setAgentAccess", () => {
...
+
+    expect(newAccess).toStrictEqual({
+      read: true,
+      append: true,
+      write: false,
+      control: false,
+    });
+  });
+
+  it("removes access previoulsy granted to the group if the new access is set to false", async () => {
```
Vinnl, line 2709:
```suggestion
  it("removes access previously granted to the group if the new access is set to false", async () => {
```

src/access/wac.ts
```diff
@@ -408,3 +409,40 @@ export async function setAgentResourceAccess<T extends WithServerResourceInfo>(
...
+ * Access was regulated by its Fallback ACL, said Fallback ACL is copied to create
+ * a new Resource ACL. This has the side effect that the next time the Fallback
+ * ACL is updated, the changes _will not impact_ the target resource.
+ *
+ * If the target Resource's Access mode cannot be determined, e.g. the user does
+ * not have Read and Write access to the target Resource's ACL, or to its
+ * fallback ACL if it does not have a Resource ACL, then `null` is returned.
+ *
+ * @param resource The Resource for which Access is being set
+ * @param agent The Group for whom Access is being set
```
Vinnl, line 426:
```suggestion
 * @param agent The Group for which Access is being set
```

src/access/wac.ts
```diff
@@ -408,3 +409,40 @@ export async function setAgentResourceAccess<T extends WithServerResourceInfo>(
...
+ * @param access The Access being set
+ * @param options Optional parameter `options.fetch`: An alternative `fetch` function to make the HTTP request, compatible with the browser-native [fetch API](https://developer.mozilla.org/docs/Web/API/WindowOrWorkerGlobalScope/fetch#parameters).
+ * @returns The Resource, with its ACL updated, or null if the new Access could not
+ * be set.
+ */
+export async function setGroupResourceAccess<T extends WithServerResourceInfo>(
```
Vinnl, line 432: As in the other review I'd call this

```suggestion
export async function setGroupAccess<T extends WithServerResourceInfo>(
```

Vinnl (approved):

src/access/wac.ts
NSeydoux: Same as for the previous PR #807, I think this should remain `setGroupResourceAccess`, and we can see how to design the default access in the universal API

PR #823, opened 2021-02-15
Add functions to set default access

NSeydoux:
In addition to setting Resource access, one also need to be able to set
default access, which applies to the children of the target Resource if
they don't have their own Resource ACL.

# Checklist

### DIFF
--- a/src/access/wac.ts
+++ b/src/access/wac.ts
@@ -29,15 +29,18 @@ import {
   getAgentAccess as getAgentAccessWac,
   getAgentAccessAll as getAgentAccessAllWac,
   setAgentResourceAccess as setAgentResourceAccessWac,
+  setAgentDefaultAccess as setAgentDefaultAccessWac,
 } from "../acl/agent";
 import {
   getGroupAccess as getGroupAccessWac,
   getGroupAccessAll as getGroupAccessAllWac,
   setGroupResourceAccess as setGroupResourceAccessWac,
+  setGroupDefaultAccess as setGroupDefaultAccessWac,
 } from "../acl/group";
 import {
   getPublicAccess as getPublicAccessWac,
   setPublicResourceAccess as setPublicResourceAccessWac,
+  setPublicDefaultAccess as setPublicDefaultAccessWac,
 } from "../acl/class";
 import {
   IriString,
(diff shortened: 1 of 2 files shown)


src/access/wac.ts
```diff
@@ -412,6 +415,44 @@ export async function setAgentResourceAccess<T extends WithServerResourceInfo>(
...
+  access: WacAccess,
+  options: Partial<
+    typeof internal_defaultFetchOptions
+  > = internal_defaultFetchOptions
+): Promise<(T & WithResourceAcl) | null> {
+  return await setActorAccess(
+    resource,
+    agent,
+    access,
+    getAgentAccessWac,
```
Vinnl, line 450: Should this not use `getAgentDefaultAccess`? (And likewise for group and public?)

Vinnl (commented):
My main concern here is whether this does the right thing, i.e. if an agent currently has read access to a Resource, it looks like that will then also be set for its children afterwards if not set using `setAgentDefautlAccess` (and likewise for the other functions).

Other than that just a variable name and some comments that haven't been updated.

NSeydoux:
After some though, I think you're right: this doesn't do the right thing. Actually, I think we need something like the current `getAgentAccess` from the WAC low-level API, but for default access: 
```
export function getAgentDefaultAccess(
  resourceInfo: WithAcl & WithServerResourceInfo,
  agent: WebId
): Access | null {
  if (hasResourceAcl(resourceInfo)) {
    return getAgentDefaultAccess(resourceInfo.internal_acl.resourceAcl, agent);
  }
  if (hasFallbackAcl(resourceInfo)) {
    return getAgentDefaultAccess(resourceInfo.internal_acl.fallbackAcl, agent);
  }
  return null;
}
```
This way, we can copy over the actual **default** access for the given resource. Which means we have to create its resource ACL, so we have to set its actual resource access as well, otherwise it would become unreachable. 

Anyways, I don't think this is high priority right now, so I'll park it for the time being.

Vinnl:
OK, works for me. As a note: `createResourceAclFromFallbackAcl` already only copies over the default access of the fallback ACL, and applies that as both resource and default access rules. (Because, as a default access rule of the parent, it already applied both to the resource itself and its children.) So if we use that, we don't have to explicitly do anything to set its resource access.

Vinnl:
Closing since the codebase has since evolved quite a bit, so once we've figured out what the right thing to do is, we'll probably want to recreate this.

PR #824, opened 2021-02-15
Add setPublicResourceAccess

NSeydoux:
This adds setPublicResourceAccess to the WAC API. This function sets access for everyone to a given resource if the resource is controlled using WAC. This only applies to Resource ACLs (as opposed to fallback ACLs),
updating it

ested for in that section of the code base,  and duplicating all these

# New feature description

# Checklist

### DIFF
--- a/src/access/wac.ts
+++ b/src/access/wac.ts
@@ -55,10 +55,12 @@ import {
   Access as AclAccess,
   AclDataset,
   createAclFromFallbackAcl,
+  getResourceAcl,
   hasAccessibleAcl,
   hasFallbackAcl,
   hasResourceAcl,
   saveAclFor,
+  WithAccessibleAcl,
   WithAcl,
   WithResourceAcl,
 } from "../acl/acl";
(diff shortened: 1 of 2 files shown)


src/access/wac.ts
```diff
@@ -292,6 +297,24 @@ async function setActorAccess<T extends WithServerResourceInfo>(
...
   options: Partial<typeof internal_defaultFetchOptions>
+): Promise<(T & WithResourceAcl) | null>;
+async function setActorAccess<T extends WithServerResourceInfo>(
+  resource: T,
+  actor: undefined,
+  access: WacAccess,
+  readAccess: typeof getPublicAccessWac,
+  writeAccess: typeof setPublicResourceAccessWac,
+  options: Partial<typeof internal_defaultFetchOptions>
+): Promise<(T & WithResourceAcl) | null>;
```
Vinnl, line 308: I don't think the overloads actually add anything here? (Though might be missing something?)

src/access/wac.ts
```diff
@@ -314,10 +337,26 @@ async function setActorAccess<T extends WithServerResourceInfo>(
...
   );
-  const currentAccess = readAccess(resourceWithOldAcl, actor) as AclAccess;
-  const wacAccess = universalAccessToAcl(access, currentAccess);
 
-  const updatedResourceAcl = writeAccess(resourceAcl, actor, wacAccess);
+  let currentAccess: AclAccess;
+  let updatedResourceAcl: (AclDataset & WithChangeLog) | undefined = undefined;
+  let wacAccess: AclAccess | undefined = undefined;
+
+  if (actor !== undefined) {
```
Vinnl, line 345: Since there are three different places this function is called form, and we're special casing to distinguish between one and the two others, it doesn't feel that generic anymore, so... Maybe this isn't the right abstraction?

I wonder if maybe it would be better to just have the first part of the function be a standalone function, and to have everything from this `if` statement on just happen in the calling functions. (Or potentially, the final `try..catch` block in a separate function as well.) What do you think?

(Also, I just realised that `readAccess` can be a bit confusing given that "read" is also an access mode :) If we do keep it, maybe it should be `getAccess` or something like that?)

src/access/wac.ts
```diff
@@ -314,10 +337,26 @@ async function setActorAccess<T extends WithServerResourceInfo>(
...
+    currentAccess = readAccess(resourceWithOldAcl, actor) as AclAccess;
+    wacAccess = universalAccessToAcl(access, currentAccess);
+    // @ts-ignore
+    updatedResourceAcl = writeAccess(resourceAcl, actor, wacAccess);
+  } else {
+    // In this case, no actor means we're using the Agent Class read/write Access,
+    // but TS still expects the Agent/Group read/write, with a slightly different
+    // signature.
+    // @ts-ignore
```
Vinnl, line 354: I'd also generally prefer to avoid `@ts-ignore` and use the appropriate type assertion instead, to make sure we're not suppressing unrelated errors. So I think you'd cast `readAccess` to `typeof getPublicAccessWac` here. Though again, special-casing for one or two calling functions I think is a code smell.

src/access/wac.ts
NSeydoux: It's intended to ensure that either the function is called with an `undefined` actor, and with a `readAccess` function that matches `typeof getPublicAccessWac`, or with a proper actor and a function that matches `typeof getActorAccessWac`, which expects a non-undefined actor. However, I think you're right, it looks like it's getting to a point where that's not the right abstraction, and that having this function be unique instead of two functions, one for actors and one for actor classes would make more sense.

NSeydoux: `setActorAccess` is now split in two functions, one for actors and one for actor classes, with the common code factored in its own functions. See c685ad7893dec1789efbe2d4b5f8e2858ae723d2

src/access/wac.ts
NSeydoux: c685ad7893dec1789efbe2d4b5f8e2858ae723d2

Vinnl (approved):

src/access/wac.ts
Vinnl: Ah of course, thanks for the explanation still :)

PR #830, opened 2021-02-16
Clarify how to run end-to-end tests

Vinnl:
I added instructions on how to add and obtain your credentials and
what they are used for, and aligned the filename of the local .env
file across both browser-based and Node-based end-to-end tests.


Vinnl:
Added b3d855b2 to clarify the reason Pat's local running of the end-to-end tests had a different result than on CI.

package.json
```diff
@@ -6,7 +6,7 @@
   "scripts": {
     "test": "eslint --config .eslintrc.js \"src/\" --ext .js,.jsx,.ts,.tsx && jest",
     "e2e-test-node": "jest --config=jest.e2e.config.js",
-    "e2e-test-browser": "testcafe",
+    "e2e-test-browser": "NODE_ENV=test testcafe",
```
pmcb55, line 9: We could do this, but in my PR I did it in the code instead, which I think is a big cleaner (i.e., from the command line you can just run `testcafe` without having to think about the NODE_ENV value at all, but you can still override the 'test' value with an env var if you want):
```javascript
// Load environment variables from .env.test.local if available:
config({
  node_env: process.env.NODE_ENV || "test",
  path: __dirname,
  // In CI, actual environment variables will overwrite values from .env files.
  // We don't need warning messages in the logs for that:
  silent: process.env.CI === "true",
});
```

src/e2e-browser/.env.defaults
```diff
@@ -1,10 +1,14 @@
-# Save a copy of this file as .env.local to prevent your credentials from being committed
+# Save a copy of this file as .env.local to prevent your credentials from being
```
pmcb55, line 1: My PR also added explanatory text around all this - should probably just merge the text of both I think, but this reference to `.env.local` is still wrong though:
```suggestion
# Save a copy of this file as .env.test.local to prevent your credentials from being
```

src/e2e-node/.env.defaults
```diff
@@ -1,12 +1,19 @@
-# Save a copy of this file as .env.local to prevent your credentials from being committed
+# Save a copy of this file as .env.local to prevent your credentials from being
```
pmcb55, line 1:
```suggestion
# Save a copy of this file as .env.test.local to prevent your credentials from being
```

src/e2e-node/.env.defaults
```diff
@@ -1,12 +1,19 @@
...
+# committed.
+# These credentials determine what Pod the end-to-end tests will write data to.
+# Generally, the tests should clean up after themselves, but it is advisable to
+# use a Pod created specifically for running the end-to-end tests.
+#
+# To obtain a refresh token, client ID and client secret for your account, run
+#     npx @inrupt/generate-oidc-token
```
pmcb55, line 8: I discussed mentioning this explicitly in here with NicAS already, and we agreed not to. Could add a reference back to README instead (i.e., "For guidance on how to generate these values, see the README.md...") - just so the information is only maintained in one place instead of multiple.

package.json
Vinnl: That's better indeed. Turns out it even has a specific option for that: https://www.npmjs.com/package/dotenv-flow#optionsdefault_node_env

17e1776a

src/e2e-browser/.env.defaults
Vinnl: Ah found them. I merged them into here, although I removed the remark about the resource - the tests do assume them to be the Pod root. 7d5bcfd6

src/e2e-node/.env.defaults
Vinnl: 7d5bcfd6

src/e2e-node/.env.defaults
Vinnl: Good point. I merged in your README and removed this from `.env.defaults`: 7d5bcfd6

src/e2e-node/README.md
```diff
@@ -2,8 +2,10 @@
...
 ## Running the tests
 
-To run the tests, run `npm install` at the root. Then, copy
-`/src/e2e-node/.env.defaults` into `/src/e2e-node/.env.test.local`,
-and insert the credentials of a Solid Pod to use for testing.
+To run the tests locally:
 
-You can now run `npm run e2e-tests-node` at the root.
+1. At the root, run `npm install`.
+2. Copy `.env.default` to `.env.test.local`.
```
pmcb55, line 8: I the last commit to my PR, I renamed `env.default(s)` to be `.env.example` (#9a0091e) - the commit message was:

Better reflects the purpose of these files, and avoids potential
confusion over using as fallback config: https://www.npmjs.com/package/dotenv-flow#alternative-defaults-envdefaults

Vinnl: I thought it made sense to have it automatically fall back to values for e.g. the server URL, but no strong opinion there: 26fccf7e

src/e2e-node/.env.defaults
Vinnl: Ah and 10bab4a4 of course.

src/e2e-node/README.md
Vinnl: Ah and 10bab4a4 of course.

NSeydoux (approved):

PR #831, opened 2021-02-16
Add wrapper API's that switch between WAC and ACP

Vinnl:
# New feature description

This adds functions to read and set access for agents, groups and the general public, automatically switching to WAC or ACP as necessary. Authenticated access is not included because it's unclear whether that will remain in WAC, and Creator access isn't included because it's not included in WAC.

New functions haven't been exported yet, because I haven't been able to verify yet whether the code actually _works_ and doesn't result in privilege escalation. For that, we'll first have to write end-to-end and property-based tests, respectively.

# Checklist

### DIFF
--- a/src/access/universal.test.ts
+++ b/src/access/universal.test.ts
@@ -1,0 +1,1026 @@
+/**
+ * Copyright 2020 Inrupt Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to use,
+ * copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the
+ * Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,
+ * INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A
+ * PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+ * HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+ * OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+ * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+import { jest, describe, it, expect } from "@jest/globals";
+import { addMockAcrTo, mockAcrFor } from "../acp/mock";
+import { mockSolidDatasetFrom } from "../resource/mock";
+import {
+  getAgentAccess,
+  getGroupAccess,
+  getPublicAccess,
+  setAgentAccess,
+  setGroupAccess,
+  setPublicAccess,
+} from "./universal";
+import * as acpLowLevel from "../acp/acp";
+import * as acpModule from "./acp";
+import * as wacModule from "./wac";
+import { addMockResourceAclTo } from "../acl/mock";
+
+describe("getAgentAccess", () => {
+  it("calls out to the well-tested ACP API for Resources with an ACR", async () => {
+    const getResourceInfoWithAcr = jest.spyOn(
+      acpLowLevel,
+      "getResourceInfoWithAcr"
+    );
+    const mockedResource = mockSolidDatasetFrom(
+      "https://arbitrary.pod/resource"
+    );
+    const mockedAcr = mockAcrFor("https://arbitrary.pod/resource");
+    const mockedResourceWithAcr = addMockAcrTo(mockedResource, mockedAcr);
+    getResourceInfoWithAcr.mockResolvedValueOnce(mockedResourceWithAcr);
+    jest.spyOn(acpModule, "internal_getAgentAccess");
+
+    await getAgentAccess(
+      "https://arbitrary.pod/resource",
+      "https://arbitrary.pod/profile#me"
+    );
+
+    expect(acpModule.internal_getAgentAccess).toHaveBeenCalledTimes(1);
+  });
+
+  it("calls out to the well-tested WAC API for Resources with an ACL", async () => {
+    const getResourceInfoWithAcr = jest.spyOn(
+      acpLowLevel,
+      "getResourceInfoWithAcr"
+    );
+    const mockedResource = mockSolidDatasetFrom(
+      "https://arbitrary.pod/resource"
+    );
+    const mockedResourceWithAcl = addMockResourceAclTo(mockedResource);
+    getResourceInfoWithAcr.mockResolvedValueOnce(mockedResourceWithAcl as any);
+    const wacGetAgentAccess = jest.spyOn(wacModule, "getAgentAccess");
+    wacGetAgentAccess.mockResolvedValue(null);
+
+    await getAgentAccess(
+      "https://arbitrary.pod/resource",
+      "https://arbitrary.pod/profile#me"
+    );
+
+    expect(wacGetAgentAccess).toHaveBeenCalledTimes(1);
+  });
+
+  it("uses the provided fetch function", () => {
+    const getResourceInfoWithAcr = jest.spyOn(
+      acpLowLevel,
+      "getResourceInfoWithAcr"
+    );
+    const options = { fetch: () => new Promise(jest.fn()) as any };
+
+    // eslint-disable-next-line @typescript-eslint/no-floating-promises
+    getAgentAccess(
+      "https://some.pod/resource",
+      "https://arbitrary.pod/profile#me",
+      options
+    );
+
+    expect(getResourceInfoWithAcr).toHaveBeenCalledWith(
+      "https://some.pod/resource",
+      options
+    );
+  });
+
+  it("passes the provided fetch function to the WAC module", async () => {
+    const getResourceInfoWithAcr = jest.spyOn(
+      acpLowLevel,
+      "getResourceInfoWithAcr"
+    );
+    const mockedResource = mockSolidDatasetFrom(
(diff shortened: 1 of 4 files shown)


src/access/universal.test.ts
```diff
@@ -0,0 +1,889 @@
...
+        "https://some.pod/resource",
+        "https://arbitrary.pod/profile#me",
+        { controlRead: true, controlWrite: undefined }
+      )
+    ).rejects.toThrow(
+      "When setting access for a Resource in a Pod implementing Web Access Control (i.e. [https://some.pod/resource]), `controlRead` and `controlWrite` should have the same value."
+    );
+  });
+
+  it("returns null when the ACP module could not update the ACR", async () => {
```
NSeydoux, line 244: Idem, should this also be tested for ACL ?

src/access/universal.test.ts
```diff
@@ -0,0 +1,889 @@
...
+    const getResourceInfoWithAcr = jest.spyOn(
+      acpLowLevel,
+      "getResourceInfoWithAcr"
+    );
+    const mockedResource = mockSolidDatasetFrom(
+      "https://arbitrary.pod/resource"
+    );
+    const mockedResourceWithAcl = addMockResourceAclTo(mockedResource);
+    getResourceInfoWithAcr.mockResolvedValueOnce(mockedResourceWithAcl as any);
+    const wacGetGroupAccess = jest.spyOn(wacModule, "getGroupAccess");
```
NSeydoux, line 368: Seeing them side-by-side, this reminds me that the WAC functions should be prefixed with `internal`, right ? It's not necessary to do it in this PR though.

src/access/universal.test.ts
```diff
@@ -0,0 +1,889 @@
...
+
+    await getGroupAccess(
+      "https://arbitrary.pod/resource",
+      "https://arbitrary.pod/groups#group"
+    );
+
+    expect(acpModule.internal_getGroupAccess).toHaveBeenCalledTimes(1);
+  });
+
+  it("calls out to the well-tested WAC API for Resources with an ACR", async () => {
```
NSeydoux, line 358:
```suggestion
  it("calls out to the well-tested WAC API for Resources with an ACL", async () => {
```

src/access/universal.test.ts
```diff
@@ -0,0 +1,889 @@
...
+    const returnedAccess = await setGroupAccess(
+      "https://arbitrary.pod/resource",
+      "https://arbitrary.pod/groups#group",
+      { read: true }
+    );
+
+    expect(returnedAccess).toStrictEqual({ append: true });
+  });
+
+  it("calls out to the well-tested WAC API for Resources with an ACR", async () => {
```
NSeydoux, line 498:
```suggestion
  it("calls out to the well-tested WAC API for Resources with an ACL", async () => {
```

src/access/universal.test.ts
```diff
@@ -0,0 +1,889 @@
...
+    const mockedResourceWithAcr = addMockAcrTo(mockedResource, mockedAcr);
+    getResourceInfoWithAcr.mockResolvedValueOnce(mockedResourceWithAcr);
+    jest.spyOn(acpModule, "internal_getPublicAccess");
+
+    await getPublicAccess("https://arbitrary.pod/resource");
+
+    expect(acpModule.internal_getPublicAccess).toHaveBeenCalledTimes(1);
+  });
+
+  it("calls out to the well-tested WAC API for Resources with an ACR", async () => {
```
NSeydoux, line 653:
```suggestion
  it("calls out to the well-tested WAC API for Resources with an ACL", async () => {
```

src/access/universal.test.ts
```diff
@@ -0,0 +1,889 @@
...
+
+    const returnedAccess = await setPublicAccess(
+      "https://arbitrary.pod/resource",
+      { read: true }
+    );
+
+    expect(returnedAccess).toStrictEqual({ append: true });
+  });
+
+  it("calls out to the well-tested WAC API for Resources with an ACR", async () => {
```
NSeydoux, line 770:
```suggestion
  it("calls out to the well-tested WAC API for Resources with an ACL", async () => {
```

src/access/universal.ts
```diff
@@ -32,3 +61,317 @@ export interface Access {
...
+ * Additionally, two caveats apply to users with a Pod server that uses WAC:
+ * - If the Resource did not have an ACL yet, a new one will be initialised.
+ *   This means that changes to the ACL of a parent Container can no longer
+ *   affect access people have to this Resource, although existing access will
+ *   be preserved.
+ * - Setting different values for `controlRead` and `controlWrite` is not
+ *   supported, and **will throw an error**. If you expect (some of) your users
+ *   to have Pods implementing WAC, be sure to pass the same value for both.
+ * - Denying access is not technically possible in WAC. Rather, what this does
+ *   is removing "allow" access.
```
NSeydoux, line 131: When we do the same for ACP (i.e. `unallow` instead of `deny`), is this description still going to be relevant ? If not, I think I'd add a note to the related ticket to update it. This remark applies to other comments in this file as well.

src/access/universal.ts
```diff
@@ -32,3 +61,317 @@ export interface Access {
...
+    return null;
+  }
+  if (hasAccessibleAcl(resourceInfo)) {
+    if (access.controlRead != access.controlWrite) {
+      throw new Error(
+        `When setting access for a Resource in a Pod implementing Web Access Control (i.e. [${getSourceIri(
+          resourceInfo
+        )}]), ` + "`controlRead` and `controlWrite` should have the same value."
+      );
+    }
```
NSeydoux, line 163: This is also tested for in the WAC part of the API. I guess testing it only once is sufficient, but I'm not sure which one is more relevant. I guess testing earlier is best, and we can update the underlying type definitions to reflect the fact that this test happened already.

NSeydoux (approved):
It is all coming together :). Some minor remarks, but overall looks good.

src/access/universal.ts
Vinnl: Yeah, I was wondering the same. I think there are two possible approaches:

- We change the WAC interface to not take `controlRead` and `controlWrite`, but just a single `control` property. After all, it doesn't support the difference anyway.
- We keep the WAC interface aligned with the universal one, with the requirement that `controlRead` and `controlWrite` are equal. In that case, the check in the WAC module is technically redundant, because TypeScript will enforce it, but I think it might be good to keep it there just in case we ever expose it to non-TS users, or if the TS interface ever gets problematic and we remove that check.

If we were starting from scratch, I think I'd lean to the former, but I'm not sure if I'm leaning towards that strong enough to outweigh the fact that the latter is already implemented. Either way changing it would be a non-breaking change, luckily :)

src/access/universal.test.ts
```diff
@@ -0,0 +1,889 @@
...
+    const options = { fetch: () => new Promise(jest.fn()) as any };
+
+    await getAgentAccess(
+      "https://arbitrary.pod/resource",
+      "https://some.pod/profile#me",
+      options
+    );
+
+    expect(wacGetAgentAccess).toHaveBeenCalledWith(
+      expect.anything(),
```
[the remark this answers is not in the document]
Vinnl: / cop-out :P

src/access/universal.test.ts
```diff
@@ -0,0 +1,889 @@
...
+
+    await getAgentAccess(
+      "https://arbitrary.pod/resource",
+      "https://arbitrary.pod/profile#me"
+    );
+
+    expect(acpModule.internal_getAgentAccess).toHaveBeenCalledTimes(1);
+  });
+
+  it("calls out to the well-tested WAC API for Resources with an ACR", async () => {
```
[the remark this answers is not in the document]
Vinnl: Whoops! 30f7c937

src/access/universal.test.ts
```diff
@@ -0,0 +1,889 @@
...
+    const returnedAccess = await setAgentAccess(
+      "https://arbitrary.pod/resource",
+      "https://arbitrary.pod/profile#me",
+      { read: true }
+    );
+
+    expect(returnedAccess).toStrictEqual({ append: true });
+  });
+
+  it("calls out to the well-tested WAC API for Resources with an ACR", async () => {
```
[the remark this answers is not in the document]
Vinnl: 30f7c937

src/access/universal.test.ts
Vinnl: 30f7c937

src/access/universal.test.ts
Vinnl: 30f7c937

src/access/universal.test.ts
Vinnl: 30f7c937

src/access/universal.test.ts
Vinnl: 30f7c937

(that you actually went through the effort of pointing them all out - I'd just have said "etc." :P)

src/access/universal.test.ts
```diff
@@ -0,0 +1,889 @@
...
+    await setAgentAccess(
+      "https://arbitrary.pod/resource",
+      "https://arbitrary.pod/profile#me",
+      { read: true }
+    );
+
+    expect(acpModule.internal_setAgentAccess).toHaveBeenCalledTimes(1);
+  });
+
+  it("returns the new effective access for ACP-controlled Resources", async () => {
```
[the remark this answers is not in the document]
Vinnl: Ehm yes, yes it should. I'll add that.

src/access/universal.test.ts
Vinnl: Yep, will do.

src/access/universal.test.ts
Vinnl: Technically, yeah. I'll try to think of it after merging this.

src/access/universal.ts
Vinnl: Yeah good point, I suppose it's technically most accurate to update that together with that ticket. I'll add a note.

src/access/universal.test.ts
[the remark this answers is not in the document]
Vinnl: d7ff2da0

src/access/universal.test.ts
Vinnl: b82640fe

src/access/universal.ts
NSeydoux: Yeah, if we had approached it in a different order, I'd have done option 1, but now that the checks are both there, it's more work to refactor things for little benefit I think. I implemented the WAC check in addition to the TS check in the spirit of defensive programming, so I guess it doesn't hurt if it stays there.

PR #836, opened 2021-02-18
Make the high-level ACP API unallow, not deny

Vinnl:
This changes the high-level ACP API to, when the value `false` is
passed for an access mode, remove allow statements for that access
mode, rather than adding a deny statement.

There are two reasons for this:
1. This aligns it with the WAC behaviour.
2. This allows the developer to actually undo previously set
   access without unintended consequences. For example, removing
   access for an agent shouldn't prevent their access if a group
   they're in still has access.

I'm a bit worried that it was a little bit _too_ straightforward to get all the tests succeeding, and the change itself was smaller than I expected as well, but since we can't test it against an actual server implementation yet, I can only cross my fingers and hope that it works as intended :)

### DIFF
--- a/src/access/acp.test.ts
+++ b/src/access/acp.test.ts
@@ -467,7 +467,7 @@ describe("hasInaccessiblePolicies", () => {
 describe("getActorAccess", () => {
   const webId = "https://some.pod/profile#me";
 
-  it("returns undefined for all access if no access was granted to the given actor", () => {
+  it("returns false for all access if no access was granted to the given actor", () => {
     const resourceWithAcr = mockResourceWithAcr(
       "https://arbitrary.pod/resource",
       "https://arbitrary.pod/resource?ext=acr",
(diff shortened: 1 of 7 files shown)


Vinnl:
Ah, forgot that the WAC API also needed changes to ensure that the `getAccess` functions did not return undefined for access modes: dcd41d26 (I'll squash that into the other commit before merging.)

NSeydoux (approved):
As part of the PR, I think all the `@returns` annotations in the WAC high-level API could be removed, provided the description of the `Access` type is part of the API documentation. Right now, the `@returns` still mention `true`, `false` or `undefined` for returned access modes, and there are no `@returns` annotations on the ACP high-level functions. Avoiding to have duplicate information between the `@returns` and the `Access` description seems to be the best approach.

Vinnl:
Ah nice catch. I updated the descriptions to state _what_ it returns (i.e. access for the given actor), but not how that's structured. That also avoids duplicate information :) And I added `@returns` annotations to the ACP API while I was at it: 26825f8e

PR #838, opened 2021-02-18
Add property-based test for the high-level ACP API

Vinnl:
The test is still disabled, because it finds examples that fail. I
have added a couple of those as (also disabled) individual tests.
After we fix those, we can re-enable the property-based tests and
see what else arises.

### DIFF
--- a/src/access/acp.test.ts
+++ b/src/access/acp.test.ts
@@ -20,6 +20,7 @@
  */
 
 import { describe, it } from "@jest/globals";
+import * as fc from "fast-check";
 import { WithAccessibleAcr } from "../acp/acp";
 import {
   AccessControlResource,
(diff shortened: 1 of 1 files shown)


src/access/acp.test.ts
```diff
@@ -4612,6 +4613,273 @@ describe("setActorAccess", () => {
...
+            );
+            const updatedResource = internal_setActorAccess(
+              resourceWithAcr,
+              actorRelation,
+              actorUrl,
+              accessToSet
+            );
+            expect(
+              internal_getActorAccess(updatedResource!, actorRelation, actorUrl)
+            ).toStrictEqual(accessToSet);
```
Vinnl, line 4725: This is the actual interesting part; So given a randomly generated pre-existing ACR configuration, and a randomly generated object of access to set, it verifies that the return value of `getActorAccess` actually was what was passed to `setActorAccess`.

As can be seen in the other skipped unit tests below, there are some ACR configurations in which that is not the case, unfortunately.

NSeydoux (approved):

PR #839, opened 2021-02-18
add clone test for objs with no ctor

pmcb55:
This PR fixes #<issue ID>.
We now allow cloning objects that contain objects with no constructor field (which can happen if an object is created with `Object.create(null);`.


Vinnl:
Already addressed by #837.

PR #840, opened 2021-02-19
Chore/move datafactory to rdfjs

pmcb55:
Internal refactor - just moved all DataFactory imports to pull from `rdfjs.js` rather than from N3 directly. The Turtle parsing code still relies explicitly on N3, but that's fine (test needed to be updated to support quad-equality between RDF/JS implementations, but that's a test improvement).
This change makes it really easy to flip our DataFactory implementation dependency now - and so I did try that out by flipping it from N3 to @rdfjs/data-model (that flip is *not* included in this PR). Only 4 tests failed, which is pretty good - with the reasons documented in `rdfjs.ts`. Those reasons need to be verified, but it would be nice to resolve them (not urgent though - we're still relying on the RDF spec compliance of N3, as we did before).

### DIFF
--- a/src/acl/acl.test.ts
+++ b/src/acl/acl.test.ts
@@ -31,9 +31,8 @@ jest.mock("../fetcher.ts", () => ({
 }));
 
 import { Response } from "cross-fetch";
-import { dataset } from "../rdfjs";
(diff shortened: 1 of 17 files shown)


ajacksified:
Merged in #837

PR #843, opened 2021-02-22
[WIP] use skolem LocalNode impl instead of BNode+obj

pmcb55:
[WIP] - **This is just for discussion**
Replacing LocaNode implementation using `BNode+attached object` with `NamedNode` skolem. Results in no assumptions on the internal structure of Quads (e.g., Graphy reconstructs Quads as opposed to returning them 'as-received' with potentially attached data structures).
DIdn't update the entire codebase, just enough to get all tests in `thing.test.ts` passing...


Vinnl:
Superseded by #1036.

PR #846, opened 2021-02-22
Add getAccessFor, wrapper for libraries

NSeydoux:
This adds `getAccessFor`, a wrapper around `getAgentAccess`,
`getGroupAccess` and `getPublicAccess` that is convenient to call from a
library, because it's only one function, and the appropriate underlying
function is determined based on the parameters.

It is implemented in a separate file from the functions it calls to
mainly to make it easier to test: the whole "universal" module can be
mocked. If `getAccessFor` is implemented alongside e.g.
`getAgentAccess`, jest cannot mock the implementation of the latter when
testing the former.

# Checklist

### DIFF
--- a/src/access/accessFor.ts
+++ b/src/access/accessFor.ts
@@ -1,0 +1,84 @@
+/**
+ * Copyright 2020 Inrupt Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to use,
+ * copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the
+ * Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,
+ * INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A
+ * PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+ * HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+ * OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+ * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+import { UrlString, WebId } from "../interfaces";
+import { internal_defaultFetchOptions } from "../resource/resource";
+import {
+  Access,
+  getAgentAccess,
+  getGroupAccess,
+  getPublicAccess,
+} from "./universal";
+
+export type Actor = "agent" | "group" | "public";
+
+/**
+ * Get an overview of what access is defined for a given actor (Agent, Group or everyone).
+ *
+ * This function works with Solid Pods that implement either the Web Access
+ * Control spec or the Access Control Policies proposal, with some caveats:
+ *
+ * - If access to the given Resource has been set using anything other than the
+ *   functions in this module, it is possible that it has been set in a way that
+ *   prevents this function from reliably reading access, in which case it will
+ *   resolve to `null`.
+ * - It will only return access specified explicitly for the given Agent. If
+ *   additional restrictions are set up to apply to the given Agent in a
+ *   particular situation, those will not be reflected in the return value of
+ *   this function.
+ * - It will only return access specified explicitly for the given Resource.
+ *   In other words, if the Resource is a Container, the returned Access may not
+ *   apply to contained Resources.
+ * - If the current user does not have permission to view access for the given
+ *   Resource, this function will resolve to `null`.
+ *
+ * @param resourceUrl URL of the Resource you want to read the access for.
+ * @param actorType type of actor whose access is being read.
+ * @param actor Identifier of the actor whose access being read for an Agent or a Group, undefined for everyone.
+ * @returns What access the given actor has.
+ */
+export async function getAccessFor(
+  resourceUrl: UrlString,
+  actorType: Actor,
+  actor?: WebId | UrlString,
+  options = internal_defaultFetchOptions
+): Promise<Access | null> {
+  if (actorType === "agent") {
+    if (actor === undefined) {
+      throw new Error(
+        "When reading Agent-specific access, the given agent cannot be left undefined."
+      );
+    }
+    return await getAgentAccess(resourceUrl, actor, options);
+  }
+  if (actorType === "group") {
+    if (actor === undefined) {
+      throw new Error(
+        "When reading Group-specific access, the given group cannot be left undefined."
+      );
+    }
+    return await getGroupAccess(resourceUrl, actor, options);
+  }
+  if (actorType === "public") {
+    return await getPublicAccess(resourceUrl, options);
+  }
+  return null;
+}
--- a/src/access/accessFor.ts
+++ b/src/access/accessFor.ts
@@ -30,6 +30,10 @@ import {
 
 export type Actor = "agent" | "group" | "public";
 
+export type GetAccessForOptions = typeof internal_defaultFetchOptions & {
+  actor?: UrlString | WebId;
+};
+
 /**
  * Get an overview of what access is defined for a given actor (Agent, Group or everyone).
  *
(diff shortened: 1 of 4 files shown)


src/access/accessFor.ts
```diff
@@ -0,0 +1,84 @@
...
+ * @param actorType type of actor whose access is being read.
+ * @param actor Identifier of the actor whose access being read for an Agent or a Group, undefined for everyone.
+ * @returns What access the given actor has.
+ */
+export async function getAccessFor(
+  resourceUrl: UrlString,
+  actorType: Actor,
+  actor?: WebId | UrlString,
+  options = internal_defaultFetchOptions
+): Promise<Access | null> {
```
Vinnl, line 63: Might be nice to have some overloads here to narrow down which combinations of arguments are required:

```suggestion
export async function getAccessFor(
  resourceUrl: UrlString,
  actorType: "agent" | "group",
  actor: WebId | UrlString,
  options = internal_defaultFetchOptions
): Promise<Access | null>;
export async function getAccessFor(
  resourceUrl: UrlString,
  actorType: "public",
  options = internal_defaultFetchOptions
): Promise<Access | null>;
export async function getAccessFor(
  resourceUrl: UrlString,
  actorType: Actor,
  actor?: WebId | UrlString,
  options = internal_defaultFetchOptions
): Promise<Access | null>;
export async function getAccessFor(
  resourceUrl: UrlString,
  actorType: Actor,
  actor?: WebId | UrlString,
  options = internal_defaultFetchOptions
): Promise<Access | null> {
```

Actually... That brings to light that it doesn't work that way - right now, if you call it with `public`, you have to explicitly pass in `undefined` for `actor`. Maybe we could instead use `internal_defaultFetchOptions` as the default for `actor` as well, and then determine which to read depending on what argument is passed?

src/access/universal.test.ts
```diff
@@ -1084,3 +1086,96 @@ describe("setPublicAccess", () => {
...
+      options
+    );
+    expect(universalModule.getAgentAccess).toHaveBeenCalledWith(
+      "https://some.resource",
+      "https://some.pod/profile#webid",
+      options
+    );
+  });
+
+  it("throws if the agent has been ommited", async () => {
```
Vinnl, line 1114:
```suggestion
  it("throws if the agent has been omitted", async () => {
```

src/access/universal.test.ts
```diff
@@ -1084,3 +1086,96 @@ describe("setPublicAccess", () => {
...
+      options
+    );
+    expect(universalModule.getGroupAccess).toHaveBeenCalledWith(
+      "https://some.resource",
+      "https://some.pod/groups#group",
+      options
+    );
+  });
+
+  it("throws if the group has been ommited", async () => {
```
Vinnl, line 1146:
```suggestion
  it("throws if the group has been omitted", async () => {
```

Vinnl (changes requested):
I understand the concern about mocking, but the module name will be part of the API (since people will be importing directly from there) - and I'm not sure whether `accessFor` is the clearest name for that module.

I'm not opposed to keeping it a separate module, but I hope we can come up with a clearer module name. I can't think of one right now, but I'll give that some more thought.

Alternatively, we can move it into the universal module and test the behaviour more - e.g. by calling `getAgentAccess` after setting an Agent, `getGroupAccess` after setting a Group, etc.

Other than that, I think we should also give some thought to how to avoid having to pass `undefined` to the function in some cases, which I don't think is desirable. Here, too, are two things I think we can do:

- Figure out a way to have the third argument act as either the actor or the options object.
- Just not support public access using this API? I can imagine that that would be acceptable, but I'm not sure.
 
Alternatively, maybe we could do something with named arguments instead? Though I suppose the interaction with the `options` object could be messy there too. Hmm... :/

src/access/accessFor.ts
NSeydoux: I wasn't sure how to go with this: either the `actor` argument is part of the optional `options`, but it feels a bit different from the other functions in the API, or it is an argument on its own, but it must be assigned to `undefined` for `public` access. I went with the latter, but without a strong opinion, so I'll change this to the former since it looks like you prefer it from an API design point of view.

Vinnl: I'm not sure if I prefer the former - I think `options` generally should catch optional arguments (hence the name :) ). However, if `actorType` is `public`, then what we call `actor` would ideally become the `options` object, if that makes sense?

NSeydoux: Fixed in b1a682995618590230c48aeb72a0e9ab9a71152f. The overload really makes it nicer, I had to add type assertions to pass in wrong parameters. With this, it actually makes more sense to have `actor` be part of `options`.

Also, I learned that default identifiers can't be used in overloads, only in the actual implementation (since it's technically not a type annotation)

src/access/universal.test.ts
NSeydoux: Dangit, another false friend ! To this day, I can't write professional in English properly on the first try because it has a double n in French.
6a18b8484ceb95bf1a0c72ad9e8f455c48a422be

src/access/universal.test.ts
NSeydoux: 6a18b8484ceb95bf1a0c72ad9e8f455c48a422be

NSeydoux:
I don't think having the third argument act as  either the actor or the options object would work, because we'll need both sometimes. I think the type overload makes it acceptable to add an `actor` entry to the `options` object, because the type system notifies you when it's missing. 

As for the module names, I initially called it `universal-convenience` (hence the name of the branch), but I think it's pretty awful, so I tried `accessFor`, since this module will basically contain `getAccessFor`, `getAccessForAll` and `setAccessFor`. Since it's an "alternative" approach to our universal API, maybe `universal-alt` ?

NSeydoux:
Small update, I tried to move things back to one module, but it would make it really intricate to test: 
```
    const getResourceInfoWithAcr = jest.spyOn(
      acpLowLevel,
      "getResourceInfoWithAcr"
    );
    const mockedResource = mockSolidDatasetFrom(
      "https://arbitrary.pod/resource"
    );
    const mockedAcr = mockAcrFor("https://arbitrary.pod/resource");
    const mockedResourceWithAcr = addMockAcrTo(mockedResource, mockedAcr);
    getResourceInfoWithAcr.mockResolvedValueOnce(mockedResourceWithAcr);
    const universalModule = jest.requireActual("./universal") as {
      getAgentAccess: () => Promise<Access | null>;
    };
    const spiedAgentAccess = jest.spyOn(universalModule, "getAgentAccess");
    await getAccessFor("https://some.resource", "agent", {
      actor: "https://some.pod/profile#webid",
    });
    expect(spiedAgentAccess).toHaveBeenCalledWith(
      "https://some.resource",
      "https://some.pod/profile#webid",
    );
```
does not work either, because (if I understand it correctly) the function reference isn't the same internally to the module and exported, so in this case we spy on the exported version but `getAccessFor` uses the internal reference. So either we mock the underlying ACP/WAC modules, which is possible but cumbersome, or we move this in a separate module. I favor the latter, because I suspect one unsing `getAccessFor` would not use `setAgentAccess`, so they would be "two" APIs in a sense. `versatile` could be an alternative to `universal`.

Vinnl:
@NSeydoux This is what I meant by having the third argument as options, which I think should work in all cases: 7ebd625d

As for the module name: `versatile` sounds better. Another idea, given your reasoning w.r.t. calling it `accessFor`: how about calling the module `"for"`? As in, the actual module name would then be `@inrupt/solid-client/access/for`, and it'd be listed in the docs as `access/for` next to `access/universal`.

src/access/for.ts
```diff
@@ -0,0 +1,112 @@
...
+ * Control spec or the Access Control Policies proposal, with some caveats:
+ *
+ * - If access to the given Resource has been set using anything other than the
+ *   functions in this module, it is possible that it has been set in a way that
+ *   prevents this function from reliably reading access, in which case it will
+ *   resolve to `null`.
+ * - It will only return access specified explicitly for the given Agent. If
+ *   additional restrictions are set up to apply to the given Agent in a
+ *   particular situation, those will not be reflected in the return value of
+ *   this function.
```
Vinnl, line 49: I _think_ that every overload can take its own docblock. So maybe what we should do here is repeat this doc block for the first two overloads (so that the `@param`s are accurate), and then make sure that this refers to either "Agent/Group" or "everyone", as appropriate.

Vinnl: (For extra bonus points we could even split up the `agent` and `group` overloads so that one takes a `WebId` and the other a `UrlString`, but since that's a lot of copy-pasting and both resolve to a string anyway, I'd be totally happy not to do that.)

src/access/universal.test.ts
```diff
@@ -1268,3 +1271,113 @@ describe("getGroupAccessAll", () => {
...
+      options
+    );
+    expect(universalModule.getAgentAccess).toHaveBeenCalledWith(
+      "https://some.resource",
+      "https://some.pod/profile#webid",
+      expect.objectContaining(options)
+    );
+  });
+
+  it("throws if the agent has been omited", async () => {
```
Vinnl, line 1299: A double T, too :smile:  (Likewise for the other one.)

```suggestion
  it("throws if the agent has been omitted", async () => {
```

src/access/universal.test.ts
```diff
@@ -1268,3 +1271,113 @@ describe("getGroupAccessAll", () => {
...
+    await getAccessFor(
+      "https://some.resource",
+      "agent",
+      "https://some.pod/profile#webid",
+      options
+    );
+    expect(universalModule.getAgentAccess).toHaveBeenCalledWith(
+      "https://some.resource",
+      "https://some.pod/profile#webid",
+      expect.objectContaining(options)
```
Vinnl, line 1295: I think this can now just be `options` again (just like in the other tests), right?

```suggestion
      options
```

src/access/universal.test.ts
```diff
@@ -1268,3 +1271,113 @@ describe("getGroupAccessAll", () => {
     expect(access).toBeNull();
   });
 });
+
+describe("getAccessFor", () => {
+  it("calls to getAgentAccess with the appropriate parameters", async () => {
+    const universalModule = jest.requireActual("./universal") as {
+      getAgentAccess: () => Promise<Access | null>;
+    };
+    universalModule.getAgentAccess = jest.fn();
```
Vinnl, line 1280: Hmm, so I take it `spyOn` still doesn't work here? Either way, if this approach does work, do we still need the separate module?

One other question: how does this overwriting of a property on `universalModule` affect other tests? Specifically, if you were to add another test that also calls code that calls out to `getAgentAccess`, will that hit this mock function or the actual one? If the former, I think that's something we'll want to prevent now, rather than hitting something really hard to debug later. (Which I suppose we could do by storing the old value and restoring it at the end of the `it`.)

Vinnl (approved):
Small suggestions. Only worry I have is whether these tests are still self-contained (i.e. do not influence (potential) other tests) - if not, I think it's best to fix that now.

src/access/universal.test.ts
NSeydoux: Next time, I'll write "is missing" :p.

I had a French teacher who always said "If you don't know how to write bicycle, just write bike", and I felle I should have taken this piece of advice more seriously.

Vinnl: Your French teacher was very wise :smile:

src/access/universal.test.ts
NSeydoux: Indeed 3118821075055cc5f9eafccbb13fb32d6046736c

src/access/universal.test.ts
NSeydoux: I don't think this impacts other tests, because what gets mocked here is the exported function, and not its reference that gets used internally in the module (which is the reason why this doesn't work if `getAccessFor` is in the same module as `getAgentAccess`. And regardless, in an upcoming PR (https://github.com/inrupt/solid-client-js/commit/0f8b532f6ba867a0ebd5b84d8fb8e5b830985a01) I already moved the tests mocking the `universal` module in a different file, because it no longer made sense to test `getAccessFor` in the same test file as `getAgentAccess` and the like, so this won't be a problem anyway very soon.

Vinnl: Ah perfect! And then maybe even the `spyOn`s will work again? :)

PR #856, opened 2021-02-23
Feat/1635/convenience get all access

NSeydoux:
`getAccessForAll` accepts getting access for groups or agents, and returns
all access set for a given resource, grouped by group or agent.

This adds a function to the `access/for` module, which is a convenience layer on top of `access/universal` This is why tests here are purely implementation-oriented: we only verify that the underlying functions have been called, and don't test for additional behaviour. 

# Checklist

- [X] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/). (to some extent, the WIP one is a tad ugly to be honest, should have squashed it)


src/access/universal.test.ts
```diff
@@ -1273,110 +1272,6 @@ describe("getGroupAccessAll", () => {
...
-    await expect(
-      getAccessFor(
-        "https://some.resource",
-        ("unknown-actor" as unknown) as "public"
-      )
-    ).resolves.toBeNull();
-  });
-
   it("re-exports getAccessFrom", () => {
     expect(reexport_getAccessFor).toBeDefined();
```
Vinnl, line 1276: I suppose `getAccessForAll` should also be re-exported?

src/access/for.ts
```diff
@@ -136,3 +138,42 @@ export async function getAccessFor(
...
+ *   of this function.
+ * - It will only return access specified explicitly for the given Resource.
+ *   In other words, if the Resource is a Container, the returned Access may not
+ *   apply to contained Resources.
+ * - If the current user does not have permission to view access for the given
+ *   Resource, this function will resolve to `null`.
+ *
+ * @param resourceUrl URL of the Resource you want to read the access for.
+ * @param actorType type of actor whose access is being read.
+ * @returns What access are set for the given resource, grouped by resp. Agent or Group.
```
Vinnl, line 165:
```suggestion
 * @returns What access is set for the given resource, grouped by resp. Agent or Group.
```

src/access/for.ts
```diff
@@ -136,3 +138,42 @@ export async function getAccessFor(
...
+  actorType: Exclude<Actor, "public">,
+  options = internal_defaultFetchOptions
+): Promise<Record<UrlString, Access> | null> {
+  if (actorType === "agent") {
+    return await getAgentAccessAll(resourceUrl, options);
+  }
+  if (actorType === "group") {
+    return await getGroupAccessAll(resourceUrl, options);
+  }
+  return null;
```
Vinnl, line 178: Given that the actor-specific function throws an error when an unknown actor type is passed, should this do so too?

If we keep it this when, then I think we should change this to:

```suggestion
  return null as never;
```

If I understand it correctly, that should mean that as soon as a new value (e.g. `authenticatedAgent`) is added to `Actor`, TypeScript will warn us that we haven't dealt with that in this function.

src/access/for.test.ts
```diff
@@ -0,0 +1,168 @@
...
+      options
+    );
+    expect(universalModule.getAgentAccess).toHaveBeenCalledWith(
+      "https://some.resource",
+      "https://some.pod/profile#webid",
+      options
+    );
+  });
+
+  it("throws if the agent has been omited", async () => {
```
Vinnl, line 51:
```suggestion
  it("throws if the agent has been omitted", async () => {
```

... :sweat_smile:

src/access/for.test.ts
```diff
@@ -0,0 +1,168 @@
...
+      options
+    );
+    expect(universalModule.getGroupAccess).toHaveBeenCalledWith(
+      "https://some.resource",
+      "https://some.pod/groups#group",
+      options
+    );
+  });
+
+  it("throws if the group has been omited", async () => {
```
Vinnl, line 86:
```suggestion
  it("throws if the group has been omitted", async () => {
```

src/access/for.ts
Vinnl: Ah, the actor-specific function doesn't throw either. Then I'd just add the `as never` :)

Vinnl (approved):

src/access/universal.test.ts
NSeydoux: eb26cabf165a204d7d57588ac2b8b4d9a1df8059

src/access/for.ts
NSeydoux: be2c4c98a0482558260202e988fb83a00cb8da11

PR #857, opened 2021-02-23
Add setAccessFor

NSeydoux:
`setAccessFor` is part of the `access/for` module. It adds a convenience
layer on top of actor type-specific functions from the
`access/universal` module to set access to a given resource for eiter
Agents, Groups, or everyone.

# Checklist

### DIFF
--- a/src/access/for.test.ts
+++ b/src/access/for.test.ts
@@ -21,7 +21,7 @@
 
 import { jest, describe, it, expect } from "@jest/globals";
 import { Access } from "./universal";
-import { getAccessFor, getAccessForAll } from "./for";
+import { getAccessFor, getAccessForAll, setAccessFor } from "./for";
 
 jest.mock("./universal");
 
@@ -31,9 +31,7 @@ describe("getAccessFor", () => {
       getAgentAccess: () => Promise<Access | null>;
     };
     const options = {
-      fetch: jest.fn().mockResolvedValue({
-        ok: false,
-      } as never) as typeof fetch,
+      fetch: jest.fn() as typeof fetch,
     };
     await getAccessFor(
       "https://some.resource",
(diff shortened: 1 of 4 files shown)


src/access/for.test.ts
```diff
@@ -166,3 +164,129 @@ describe("getAccessForAll", () => {
...
+      "https://some.resource",
+      "https://some.pod/groups#group",
+      {
+        read: true,
+      },
+      options
+    );
+  });
+
+  it("throws if the group is missing", async () => {
```
NSeydoux, line 231: Fool me once, shame on me. Fool me twice, MEGA SHAME ON ME

src/access/for.ts
```diff
@@ -177,3 +180,131 @@ export async function getAccessForAll(
...
+  options?: typeof internal_defaultFetchOptions
+): Promise<Access | null>;
+export async function setAccessFor(
+  resourceUrl: UrlString,
+  actorType: Actor,
+  access: Partial<Access>,
+  actor?: WebId | UrlString | typeof internal_defaultFetchOptions,
+  options?: typeof internal_defaultFetchOptions
+): Promise<Access | null>;
+export async function setAccessFor(
```
Vinnl, line 275: When an overload is specified, the function's own type is hidden. Since some libraries might not yet know the type of actor (i.e. they're just passing along an arbitrary string passed to them), it can be annoying if only the specific overloads are available. To avoid annoying warnings there, we can add the generic type as a fallback overload:

```suggestion
export async function setAccessFor(
  resourceUrl: UrlString,
  actorType: Actor,
  access: Partial<Access>,
  actor?:
    | WebId
    | UrlString
    | typeof internal_defaultFetchOptions,
  options?: typeof internal_defaultFetchOptions
): Promise<Access | null>;
export async function setAccessFor(
```

Vinnl (approved):

src/access/for.ts
NSeydoux: I'm a bit confused: isn't this already the case ? If I understand this correctly, there are three overloads currently specified, the two first (with the docblocks) being specific, the third one being generic, and then the actual function is declared ? Anyways, now I understand better why the last overload matches the function signature ^^.

Vinnl: Ehm, well, you know, that might be the case if you actually look properly, but if, like me, you just don't look properly, then you can easily pretend that the overload does not exist.

Jokes aside, you're right, I didn't look properly. Carry on!

PR #861, opened 2021-02-24
Export Access API from index

NSeydoux:
This exposes our Access API from the index entry point. This way, applications which aren't using a latest Node version supporting ES modules can also use this API.

# Checklist


src/access/v1.ts
```diff
@@ -0,0 +1,29 @@
+/**
```
Vinnl, line 1: Do we really need this file? We've given a bit more thought to this API and are still marketing this as a preview release (i.e. the big 1.1 is later in the year), so there's less worry here about breaking backwards compatibility.

(Plus, the "v1" isn't even exposed to developers, so it doesn't actually do anything to help ease breaking changes :) )

src/index.ts
```diff
@@ -226,3 +226,18 @@ export { acp_v1 } from "./acp/v1";
...
+ * discovers metadata associated to the Resource itself, and calls the appropriate
+ * underlying API to deal with the Access Control in place for the target Resource.
+ *
+ * Being still developped, the following export is *only* intended for experimentation
+ * by early adopters, and is not recommended yet for production applications. Because
+ * of this, all of the Access-related API's are exported on a single object, which does
+ * not facilitate tree-shaking: if you use one ACP-related API, all of them will be
+ * included in your bundle.
+ */
+export { access_api } from "./access/v1";
```
Vinnl, line 243: Since everything here is an API, this could be the start of [smurf naming](https://devcards.io/smurf-naming-convention) :) Maybe something like `resource_access` (since it just supports setting access for individual Resources) or possibly just `access` would be more appropriate?

src/index.ts
```diff
@@ -226,3 +226,18 @@ export { acp_v1 } from "./acp/v1";
...
 export { acp_v2 } from "./acp/v2";
+
+/**
+ * This API is still experimental, and subject to change. It builds on top of both
+ * ACP and ACL, aiming at being adaptable to any Access Control system that may be
+ * implemented in Solid. That is why it is purely Resource-centric: the library
+ * discovers metadata associated to the Resource itself, and calls the appropriate
+ * underlying API to deal with the Access Control in place for the target Resource.
+ *
+ * Being still developped, the following export is *only* intended for experimentation
```
Vinnl, line 237: In the case of this API, we're only including this object because not all bundlers support exports maps yet. So it's probably a good idea to add a note here that we recommend importing directly from `@inrupt/solid-client/access/universal` if you're using Node 12+, Webpack 5+, or anything else that supports exports maps.

CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ The following changes have been implemented but not released yet:
 
 ## [Unreleased]
 
+### New features
+
+- `access_api`: A set of APIs to manage Access which are agnostic to the Access
```
Vinnl, line 11: Sorry, the ticket was unclear here, but since this object is a workaround for people who aren't able to use exports maps, as part of this PR we should also make the universal access module available through the exports map (i.e. using the "New feature" PR checklist).

And since that's the recommended way to import it (if your tooling supports it, which most tooling will as time goes by), we should just mention that module here and only mention this object in the docs with a note on compatibility.

And maybe explicitly mention that the Access Control mechanism currently supported are the only two mechanisms for Solid currently known, i.e. Web Access Control and Access Control Policies. Or would that be too detailed and something better suited for a note somewhere in the docs?

Vinnl (changes requested):
Mea culpa, the ticket was a bit unclear, but this is the PR in which we should also start exposing the `access/universal` module proper :)

src/access/v1.ts
NSeydoux: True, initially I wanted to isolate the export of all the functions, and named it `v1` even if the name wasn't exposed to developers to be consistent with the other files exporting groups of functions in `acp` in particular, but you're right actually this doesn't add any value in this context :)

NSeydoux: b7725797b5be63c137b0ec84258765b2c756e662

src/index.ts
NSeydoux: I went for just `access`, mostly because it's shorter, and a bit because I'm not sure adding `resource_` makes it much clearer for users. b7725797b5be63c137b0ec84258765b2c756e662

src/index.ts
NSeydoux: 2a7b6b9c2ebc2d1e971724c29cb4989774f74224

CHANGELOG.md
NSeydoux: Fixed respectively in de729cf17da372ad5b87e27a1b4be238ab95e6c5 and 35a8b79d33792e4e0b2565da636a76cff5aa0b29. Thanks for the detailed explanations !

CHANGELOG.md
```diff
@@ -6,6 +6,14 @@ The following changes have been implemented but not released yet:
 
 ## [Unreleased]
 
+### New features
+
+- `@inrupt/solid-client/access/universal` and `@inrupt/solid-client/access/for`:
```
Vinnl, line 11:
```suggestion
- `@inrupt/solid-client/access/universal`:
```

The former re-exports the latter, so there's no need to expose the latter :)

CHANGELOG.md
```diff
@@ -6,6 +6,14 @@ The following changes have been implemented but not released yet:
...
+  A set of APIs to manage Access which are agnostic to the Access Control mechanism
+  actually implemented (from the user's point of view). Currently, these modules
+  support the two know access controls for Solid, i.e. Web Access Control and
```
Vinnl, line 14:
```suggestion
  support the two known access controls for Solid, i.e. Web Access Control and
```

package.json
```diff
@@ -48,7 +48,9 @@
     "./acl/agent": "./dist/acl/agent.mjs",
     "./acl/group": "./dist/acl/group.mjs",
     "./acl/class": "./dist/acl/class.mjs",
-    "./acl/mock": "./dist/acl/mock.mjs"
+    "./acl/mock": "./dist/acl/mock.mjs", 
+    "./access/universal": "./dist/access/universal.mjs",
+    "./access/for": "./dist/access/for.mjs"
```
Vinnl, line 53: Likewise:

```suggestion
    "./access/universal": "./dist/access/universal.mjs"
```

src/index.ts
```diff
@@ -226,3 +226,23 @@ export { acp_v1 } from "./acp/v1";
...
```
Vinnl, line 237:
```suggestion
 * Being still developed, the following export is *only* intended for experimentation
```

though I think the tense is off. Unfortunately I'm not enough of an English scholar to know when terms like "present participle" and all that apply, but I think it should be something like the following :P 

```suggestion
 * As it is still under development, the following export is *only* intended for experimentation
```

src/index.ts
```diff
@@ -226,3 +226,23 @@ export { acp_v1 } from "./acp/v1";
...
+ * by early adopters, and is not recommended yet for production applications. Because
+ * of this, all of the Access-related API's are exported on a single object, which does
+ * not facilitate tree-shaking: if you use one ACP-related API, all of them will be
+ * included in your bundle.
+ *
+ * Note that the following object is exposed to be available for environments not
+ * supporting export maps. For developers using Node 12+, Webpack 5+, or any tool
+ * or environment with support for export maps, we recommand you import these
```
Vinnl, line 245:
```suggestion
 * or environment with support for export maps, we recommend you import these
```

tsconfig.json
```diff
@@ -84,6 +84,8 @@
       "src/acl/group.ts",
       "src/acl/class.ts",
       "src/acl/mock.ts",
+      "src/access/universal.ts",
+      "src/access/for.ts"
```
Vinnl, line 88:
```suggestion
      "src/access/universal.ts"
```

CHANGELOG.md
```diff
@@ -6,6 +6,14 @@ The following changes have been implemented but not released yet:
 
 ## [Unreleased]
 
+### New features
+
+- `@inrupt/solid-client/access/universal` and `@inrupt/solid-client/access/for`:
+  A set of APIs to manage Access which are agnostic to the Access Control mechanism
```
Vinnl, line 12: Also, maybe we should mention here too that this is a preview release?

```suggestion
- `@inrupt/solid-client/access/universal`: an early preview of
  a set of APIs to manage Access which are agnostic to the Access Control mechanism
```

Possibly even noting the fact that there are some bugs that are yet to be fixed, but that they only occur in a limited number of edge cases involving existing Access Policies.

Vinnl (approved):
There's no need to export the `for` module since it's re-exported from `universal`, other than that a couple of typo fixes and maybe a note in the changelog and we're good to go :)

CHANGELOG.md
NSeydoux: If you only used `getAccessFor`, and directly imported `@inrupt/solid-client/access/for`, wouldn't you be able to tree-shake `@inrupt/solid-client/access/universal` ? It's not like it was a massive gain, but that was my reasoning initially ^^

Vinnl: Most bundlers are luckily smarter than that and can statically analyse code on the function level, i.e. if you only import a single function from a module, and that doesn't reference other functions in that module and doesn't have side effects, it will drop all other code from that module. (This only works with ES modules though.)

PR #869, opened 2021-03-02
Docs: include interfaces in api docs

kay-kim:
Tweaks to landing page and makefile to include the interfaces api docs.

https://lit-pod-hznaeypwy-inrupt.vercel.app/interfaces/access_universal.access.html

Note: I put the interfaces after the Modules, but can switch order if that's preferable.

We can also maybe think about tweaking the left-hand nav to have top level pages so that we can collapse them. 

```
├──Modules
|   └──Module:blah 
|   └──Module:etc
    ...
├── Interfaces
|  └──Interface: blah 
├── Classes
   └──Class: fooError
   ... 
```

 If so, give a holler and I can do that in a separate pull request later this week.

### DIFF
--- a/docs/api/Makefile
+++ b/docs/api/Makefile
@@ -38,6 +38,7 @@ migrate: clean
 	cp -R $(SOURCEDIR)/index.rst $(SOURCECOPYDIR)
 	cp -R $(SOURCEDIR)/api/modules/ $(SOURCECOPYDIR)/modules/
 	cp -R $(SOURCEDIR)/api/classes/ $(SOURCECOPYDIR)/classes/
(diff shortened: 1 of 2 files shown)


Vinnl (approved):
Huh, weird that the `interfaces` module lists its interfaces inline, but the `access/universal` does not.

I'm going to project my experience over all developers', but I'd prefer not to have the nav to be collapsed. Usually when I look at reference documentation I know what I'm looking for, and if I don't have the relevant page in my browser history yet, being able to Ctrl+F to find it is a boon (and better than having to use the site's own search field, since Ctrl+F is consistent across sites).

One thing we could look at is removing/inlining the class pages. Most of the properties/methods are native `Error` ones and thus not that interesting. But since the modules are listed first, that's still not a major issue I think.

PR #874, opened 2021-03-03
Fix self-contradicting Policies being ignored

Vinnl:
This appears to be the last bug discovered by the proptests, so I've enabled them now (but only in CI, since they take a while). I think the ACP API is in a good state now!


NSeydoux (approved):
Well done, the property testing really has gotten us ahead of bugs that would have been a pain to debug !

Vinnl:
Yeap, super happy with it, gives me a lot of confidence that it's working as intended :)

PR #877, opened 2021-03-04
Fix removal of access without replacing it

Vinnl:
Another bug discovered by fast-check, when running the CI job for #876. Apparently the first couple of thousand times we ran the proptests weren't enough to surface a situation with this bug.

If:
- An access mode was requested to be set to false,
- A Policy exists that denies that access,
- Another Policy exists that allows that access, also applies to
  another actor, and allows other desired access

then that latter Policy would be unapplied (because it allowed
access we did not want to apply), but no new Rule would be added to
apply the other desired access (because the existing combination of
Rules already resulted in that access).

Now, we first simulate removal of the access, then check whether
the desired access would still apply, and if not, then we *do* add
a new Rule applying the desired access.

### DIFF
--- a/src/access/acp.test.ts
+++ b/src/access/acp.test.ts
@@ -4823,6 +4823,62 @@ describe("setActorAccess", () => {
         internal_getActorAccess(updatedResource!, actorRelation, actorUrl)
       ).toStrictEqual(accessToSet);
     });
+
+    it("adds a new Policy with the desired access if a Policy that already applies the access gets removed", () => {
+      const acrConfig = {
+        acrPolicies: {
+          "https://some.pod/resource?ext=acl#policy1": {
+            allow: { read: true, write: true },
+            anyOf: {
+              "https://some.pod/resource?ext=acl#rule2": {
+                "http://www.w3.org/ns/solid/acp#agent": [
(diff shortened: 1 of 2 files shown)


NSeydoux:
Funny how it took multiple thousands permutations to find this out, but I'm sure it would have surfaced three days after people start building apps with this ^^

Vinnl:
Haha well I hope not, since it is the result of existing Policies. But surely as soon as people start defining their own Policies (possibly with PodBrowser), yes, probably.

Another "fun" fact: setting the number of iterations to 20.000 locally I got another counter example...

src/access/acp.test.ts
```diff
@@ -4823,6 +4823,62 @@ describe("setActorAccess", () => {
         internal_getActorAccess(updatedResource!, actorRelation, actorUrl)
       ).toStrictEqual(accessToSet);
     });
+
+    it("sets adds a new Policy with the desired access if a Policy that already applies the access gets removed", () => {
```
NSeydoux, line 4827:
```suggestion
    it("adds a new Policy with the desired access if a Policy that already applies the access gets removed", () => {
```

NSeydoux (approved):

PR #878, opened 2021-03-04
Revert CodeSandbox back to Parcel

Vinnl:
This partially reverts 4027d2a.

Apparently CodeSandbox interprets the app as a "Node app" when not
using Parcel, which (currently) prevents people from forking it
when not signed in, and presumably will cause it to make
assumptions that do not hold. Since Parcel only caused problems in
setups we're not using, I've reverted the Webpack migration for
now. We can re-enable it if the problems re-appear and CodeSandbox
support is more stable.


Vinnl:
Proptests failing, presumably due to #877.

ajacksified (changes requested):
> Since Parcel only caused problems in setups we're not using

Parcel caused problems as soon as we directly imported n3 - I'm not comfortable moving forward with this until we're confident we can make minor changes, like importing a package, without it breaking.

Additionally, Webpack and Rollup are the only two currently approved packaging systems as per our JS guidelines.

Vinnl:
Right, so as far as I can see, the following are currently true:

- The current setup breaks a QA mechanism we're actively using (that's why I discovered i was broken)
- We are not currently directly importing n3.js
- There are three projects (n3.js, Parcel, CodeSandbox) each of which can independently fix both our issues
- We have the fix for direct imports right here, ready to re-apply if we do need to directly import from N3

So I would think that that means that we can fix our breakage right now, and if we do need to import directly from n3.js, either enough time has passed that one of those three projects has issued a fix, or we can re-apply this. From that, this PR seems like a no-brainer to me, but maybe I'm missing something?

That said, now that you mention direct imports from n3.js: the reason we had to do that is because of the way it's setup breaking either Node+ESM or [Parcel and Webpack <=4]. So if you encountered the problems only when importing directly from n3, it sounds like maybe that was what you were running into, rather than a problem with Parcel? (You could verify this by trying to run the sandbox using Webpack 4 - if that breaks too, the problem is not with Parcel.) In that case, we _shouldn't_ be importing from n3 directly, as that would break projects that are using Webpack 4, like PodBrowser.

I don't think the last comment applies, as we're not using it as our packaging system, but as a devServer, and emulating user setups. And most importantly, we're not using it because we want to, because it's the only choice we have here, unfortunately.

ajacksified (approved):
As discussed - we can move forward with this for now, with the understanding that n3 causes issues if directly imported.

Vinnl:
See also https://github.com/inrupt/solid-client-js/blob/4a9aced72b006d41bae570153476705ff3f39dc0/src/formats/turtle.ts#L84-L98

That said, it does appear that either TestCafe or Parcel is throwing a fit now, so I'll see if I have a flash of insight on how to easily solve that hopefully soon :P

PR #884, opened 2021-03-05
Added `internal_getClientAccess`

NSeydoux:
This adds a function to read which access are granted to a given Client
App for a given resource.

Note that currently, this function isn't available as part of the public
API, because it is only implemented for resources which access control
is based on ACP.

The current suggestion is to not expose this feature as part of the universal API until client access control is clarified for WAC. In order to make this function part of a public API, an upcoming PR should change the current `src/access/acp.ts` into `src/access/acp.internal.ts`, move functions we want to add to the public API to `src/access/acp.ts`, and add the latter to the export map. 

# Checklist

- [ ] The changelog has been updated, if applicable. N/A, since nothing changes from an external point of view.

### DIFF
--- a/src/acp/rule.test.ts
+++ b/src/acp/rule.test.ts
@@ -65,6 +65,8 @@ import {
   setCreator,
   ruleAsMarkdown,
   removeRule,
+  getClientAll,
+  hasPublicClient,
 } from "./rule";
 
 import { Policy } from "./policy";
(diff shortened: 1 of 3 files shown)


Vinnl (changes requested):
I think it's a bit early to add this to the high-level ACP API, both because it's not actually accessible here yet, but also because it's not clear how this would interact with Agent or Group access. (So for example, if I set both Agent and Group access, then someone needs to be either that Agent or in that Group for either to apply. However, if I set Client access, does that mean that anyone using that Client can change the data? Or should it be combined with an Agent or Group Rule, and if so how?)

And of course, the low-level API, i.e. the one that should make all functionality available, still doesn't support with these changes.

So I think the implementation we want to go for here is not in `access/acp`, but rather in `acp/rule`. What do you think? (I hope you haven't spent too much effort on this... Sorry :disappointed:  )

NSeydoux:
No worries, your comments are valid, and to be honest it wasn't that much work.

Vinnl (approved):
As discussed, let's rename references to "PublicClient" to "AnyClient". Other than that just a small polishing up of the test and we're good to go :)

src/acp/rule.test.ts
```diff
@@ -1282,6 +1322,50 @@ describe("setCreator", () => {
...
+    const rule = mockRule(MOCKED_RULE_IRI, {
+      publicClient: true,
+    });
+    expect(hasAnyClient(rule)).toEqual(true);
+  });
+  it("returns false if the rule only applies to individual clients", () => {
+    const rule = mockRule(MOCKED_RULE_IRI, {
+      clients: [MOCK_CLIENT_WEBID_1],
+    });
+    expect(hasAnyClient(rule)).toEqual(false);
```
Vinnl, line 1365: Oh btw, booleans are atomic values too, so `.toBe` works for those :)

PR #888, opened 2021-03-08
Fix existing access not being taken into account

Vinnl:
The fix in a13dee6f5fde5ecde0015c6516e2da8047fc4db3 contained a
mistake where it should look at remaining access to determine
whether an existing Policy needed replacing, but it should compare
it to the Access that was defined before.


NSeydoux (approved):

Vinnl:
Just ran the proptests for 20 000 runs and it detected no errors, so I think we're finally good now :)

PR #913, opened 2021-03-16
Change class modifiers to setters/removers

Vinnl:
# New feature description

This splits `setPublic`, `setAuthenticated`, `setCreator` and `setAnyClient` into `set` and `remove` functions, rather than taking a boolean as the second argument. It also preserves the old version in the v2 API, and adds a warning if the developer is still calling the old signature somewhere.

It's not yet public, since there's some other work up next that also breaks the API, so I'd prefer to package them both up as v3 (should be the final one), rather than creating a v3 now and a v4 tomorrow.

I also fixed two bugs I happened to encounter: `setAgent` overwrote Creators (it's so weird that these are just "special" Agents...), and clients weren't listed in a Rule's Markdown representation.

And finally, I made it explicit what is part of `acp_v1` and `acp_v2` so future changes are not blindly added to them.

# Checklist

- [ ] Relevant documentation, if any, has been written/updated. - I'll send a PR to the docs when it's made public.
- [x] The changelog has been updated, if applicable. N/A - not publicly supported yet
- [x] New functions/types have been exported in `index.ts`, if applicable. N/A
- [x] New modules (i.e. new `.ts` files) are listed in the `exports` field in `package.json`, if applicable. N/A
- [x] New modules (i.e. new `.ts` files) are listed in the `typedocOptions.entryPoints` field in `tsconfig.json`, if applicable. N/A

### DIFF
--- a/src/acp/rule.test.ts
+++ b/src/acp/rule.test.ts
@@ -71,6 +71,10 @@ import {
   removeClient,
   hasAnyClient,
   setAnyClient,
+  removePublic,
+  removeAuthenticated,
+  removeCreator,
+  removeAnyClient,
 } from "./rule";
 
 import { Policy } from "./policy";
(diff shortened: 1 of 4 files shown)


src/acp/rule.test.ts
```diff
@@ -895,10 +899,11 @@ describe("setAgent", () => {
     ).toBe(true);
   });
 
-  it("does not overwrite public and authenticated agents", () => {
+  it("does not overwrite public, authenticated and creator agents", () => {
```
NSeydoux, line 902: Good catch !

NSeydoux (approved):

src/acp/rule.test.ts
Vinnl: Yeah, I'm a bit worried that it needed catching, and that people will be using outdated version of the library as more "special" Agents are added to the spec, but it is what it is :/

PR #940, opened 2021-03-25
Regularly run end-to-end tests

Vinnl:
Runs the end-to-end tests every five minutes, and sends a notice to a
Slack channel when they fail.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,11 @@
 name: CI
 
-on: [push]
+on:
+  push:
+  schedule:
+  # Run every fifth minute.
+  # This is to verify that ESS is running properly.
+  - cron: '*/5 * * * *'
 
 env:
   CI: true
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -1,6 +1,11 @@
 name: CI
 
-on: [push]
+on:
+  push:
+  schedule:
+  # Run every fifth minute.
```
NSeydoux, line 6: Is there a rate limit to how much we can run GH actions, even on public repos ?

NSeydoux (approved):

.github/workflows/ci.yml
Vinnl: [Doesn't look like it](https://docs.github.com/en/github/setting-up-and-managing-billing-and-payments-on-github/managing-your-spending-limit-for-github-actions), although 5 minutes is [the shortest available interval](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#schedule). If there is, hopefully six days will be enough to hit it, so we can fix it before next month :)

PR #942, opened 2021-03-26
Prevent overriding resource if IRI ends with #

NSeydoux:
This should fix https://github.com/solid/solidproject.org/issues/546.

When fetching a resource, its source IRI is server-managed, which means
it does not include a hash fragment. Writing back to a resource that
has been fetched including the hash fragment, there was a mismatch
between the source IRI (no fragment) and the target IRI (with the
fragment), which lead the library to do a PUT instead of a PATCH.

IRIs are now normalized before being compared.

**Note to reviewers**: Initially, I thought about normalizing the response's `url` on fetch, and only normalizing the target IRI in `isUpdate`, but it had more side-effects, and it needed a lot of test updated because they do not set `url` on the `Response` when  fetching. I didn't find any additional places where the source IRI should be updated.

Additionally, do you think it would be wise to add to the API a `saveSolidDatasetBack` function, that does not take a target `url`, and only uses the `source` IRI from the provided `SolidDataset` ? That would enable being extra-explicit about the expected behaviour, and would prevent a `PUT` to be considered at all.

### DIFF
--- a/src/resource/iri.internal.ts
+++ b/src/resource/iri.internal.ts
@@ -1,0 +1,33 @@
+/**
+ * Copyright 2020 Inrupt Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to use,
+ * copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the
+ * Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,
+ * INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A
+ * PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+ * HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+ * OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+ * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+import { IriString } from "../interfaces";
+
+/**
+ * This function normalizes IRIs as managed by the server to ease accurate comparison.
+ * @param iri
+ * @hidden
+ */
+export function normalizeServerSideIri(iri: IriString): IriString {
+  const iriObj = new URL(iri);
+  iriObj.hash = "";
+  return iriObj.href;
+}
--- a/src/resource/solidDataset.test.ts
+++ b/src/resource/solidDataset.test.ts
@@ -370,6 +370,13 @@ describe("saveSolidDatasetAt", () => {
       >;
     };
 
+    mockedFetcher.fetch.mockResolvedValue(
+      mockResponse(null, {
+        headers: { Location: "/resource" },
+        url: "https://some.pod/resource",
+      })
+    );
+
     await saveSolidDatasetAt("https://some.pod/resource", dataset());
 
     expect(mockedFetcher.fetch.mock.calls).toHaveLength(1);
@@ -425,7 +432,9 @@ describe("saveSolidDatasetAt", () => {
     it("sets relative IRIs for LocalNodes", async () => {
       const mockFetch = jest
         .fn(window.fetch)
-        .mockReturnValue(Promise.resolve(new Response()));
+        .mockResolvedValue(
+          mockResponse(null, { url: "https://some.irrelevant/url" })
+        );
       const mockDataset = dataset();
       const subjectLocal: LocalNode = Object.assign(DataFactory.blankNode(), {
         internal_name: "some-subject-name",
@@ -454,7 +463,8 @@ describe("saveSolidDatasetAt", () => {
     it("resolves relative IRIs in the returned SolidDataset", async () => {
       const mockFetch = jest
         .fn(window.fetch)
-        .mockReturnValue(Promise.resolve(new Response()));
+        .mockResolvedValue(mockResponse(null, { url: "https://saved.at/url" }));
+
       const mockDataset = dataset();
       const subjectLocal: LocalNode = Object.assign(DataFactory.blankNode(), {
         internal_name: "some-subject-name",
@@ -481,9 +491,9 @@ describe("saveSolidDatasetAt", () => {
 
       expect(
         storedSolidDataset.match(
-          DataFactory.namedNode("https://some.pod/resource#some-subject-name"),
+          DataFactory.namedNode("https://saved.at/url#some-subject-name"),
           null,
-          DataFactory.namedNode("https://some.pod/resource#some-object-name")
+          DataFactory.namedNode("https://saved.at/url#some-object-name")
         ).size
       ).toBe(1);
     });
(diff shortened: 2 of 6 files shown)


src/resource/iri.internal.ts
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+import { IriString } from "../interfaces";
+
+/**
+ * This function normalizes IRIs as managed by the server to ease accurate comparison.
+ * @param iri
+ * @hidden
+ */
+export function normalizeServerSideIri(iri: IriString): IriString {
```
Vinnl, line 29: At the bottom of `acp/acp.ts`, this exact function also exists - might be a good moment to replace that with this one.

Also, would you mind adding this as an additional check to all the `save*` (and `overwrite*`) functions? I'm afraid similar bugs might be lurking in there - for example, line 244 of `solidDataset.ts` sets the URL of the returned SolidDataset to the URL passed to `saveSolidDatasetAt`. (Which I suppose would be better if it would take that from `response.url` anyway...)

src/resource/solidDataset.test.ts
```diff
@@ -788,6 +788,61 @@ describe("saveSolidDatasetAt", () => {
...
+        "https://some.pod/resource#something",
+        mockDataset,
+        {
+          fetch: mockFetch,
+        }
+      );
+
+      expect(mockFetch.mock.calls).toHaveLength(1);
+      expect(mockFetch.mock.calls[0][0]).toEqual(
+        "https://some.pod/resource#something"
```
Vinnl, line 832: So in line with my other suggestion, this would become

```suggestion
        "https://some.pod/resource"
```

Vinnl (approved):
Looks good; I think there are a couple more places where it'd be best to apply this too though :)

src/resource/iri.internal.ts
pmcb55, line 29: I don't really follow this comment, and it's not an accurate description of what the function does, right? i.e., shouldn't the function just be called `removeIriHashFragment(iri))`, and a detailed comment added to wherever this function is being called to explain why we need to call it in that context? (In other words, we _could_ consider adding a detailed comment to this function trying to explain use-cases for calling it, but that feels a bit like added use-case comments to a `String.trim()` function, so it doesn't feel right).

src/resource/solidDataset.ts
```diff
@@ -444,7 +445,8 @@ function isUpdate(
     hasChangelog(solidDataset) &&
     hasResourceInfo(solidDataset) &&
     typeof solidDataset.internal_resourceInfo.sourceIri === "string" &&
-    solidDataset.internal_resourceInfo.sourceIri === url
+    normalizeServerSideIri(solidDataset.internal_resourceInfo.sourceIri) ===
+      normalizeServerSideIri(url)
```
pmcb55, line 449: An alternative here, perhaps more readable, would be to provide an explicit comparison function that understands this particular context, e.g.:
```suggestion
    solidDatasetSourceEquals(solidDataset, url)
```

```javascript
function solidDatasetSourceEquals(solidDataset: SolidDataset, target: Iri): boolean {
  return solidDataset.internal_resourceInfo.sourceIri === target;
}
```

src/resource/solidDataset.test.ts
NSeydoux: Well in that case I think it's up to the `fetch` function to strip the hash fragment from the target IRI before making the request, as opposed to the client app tampering with the target IRI. However, in line with your previous suggestion, the library should use the URL of the response (here likely to be `https://some.pod/resource`) instead of the user-provided `https://some.pod/resource#something` target IRI. That's what's enforced in f4f2d817579809804f5c6db76ea63438550d7d4d, and tested in c020d3ed48efc11504b4e4ce6866b6fdfebb061c.

src/resource/iri.internal.ts
NSeydoux: Good catch, I didn't realize it was there :) 400531cb49fb424dbfaea936aa91a535dd6b0dd4.

As for other `save` or `overwrite` methods, I'm not sure if anything is required: the bug in this case is that we use the IRI comparison to check whether we want to `PUT` or `PATCH`. When saving in container or overwriting, the method to use is explicit, so no IRI comparison is used, which prevents the bug from manifesting I think.

NSeydoux: My thinking was the following: this function intends to take a user-supplied IRI, and normalize it to compare it to the server "view" of the same IRI. At the present, it only strips the hash fragment, but I don't exclude that we discover additional bugs like this one that require to do additional normalization on the IRI. 

In other words, the name of the function here describes the context in which it should be used, rather than exactly what it does.

src/resource/solidDataset.ts
NSeydoux: 709081950bf8327b2904c5101ddb687db76f58ea

NSeydoux:
> > Initially, I thought about normalizing the response's url on fetch, and only normalizing the target IRI in isUpdate

> Wasn't that already the case? That does indeed sound desirable, but maybe something to do in a separate PR if it had so much fallout.

Arguably, now that we are only using the server-managed IRIs (as opposed to user-provided ones), such normalization when fetching isn't as critical, but that still sounds like a reasonable thing to do.

src/resource/solidDataset.ts
```diff
@@ -518,8 +528,7 @@ export async function saveSolidDatasetInContainer(
     );
   }
 
-  const resourceIri = new URL(locationHeader, new URL(containerUrl).origin)
-    .href;
+  const resourceIri = new URL(locationHeader, response.url).href;
```
Vinnl, line 531: I think this would be a bug: the URL in `locationHeader` is relative to the origin of the server, rather than the Container's URL, IIRC (but be sure to check):

```suggestion
  const resourceIri = new URL(locationHeader, new URL(response.url).origin).href;
```

src/resource/solidDataset.ts
```diff
@@ -599,8 +608,7 @@ export async function createContainerInContainer(
...
```
Vinnl, line 611: Ditto:

```suggestion
  const resourceIri = new URL(locationHeader, new URL(response.url).origin).href;
```

src/resource/iri.internal.ts
Vinnl: Fair enough - I suppose it's only relevant in determining the source URL when returning it, fetching should be fine.

src/resource/solidDataset.test.ts
```diff
@@ -2168,13 +2323,37 @@ describe("createContainerInContainer", () => {
...
+  });
+
+  it("includes the final slug with the return value, relative to the target Container", async () => {
     const mockFetch = setMockOnFetch(
       jest.fn(window.fetch),
-      new Response("Arbitrary response", {
+      mockResponse("Arbitrary response", {
         headers: {
-          Location: "parent-container/child-container/",
+          Location: "child-container/",
```
Vinnl, line 2354: So this mock does not reflect actual server behaviour now, I think:

```suggestion
          Location: "parent-container/child-container/",
```

Vinnl: Oh wait, I just noticed that this is an extra test. Is the server/spec behaviour actually that it's always relative to the Container URL, but using a preceding slash to make it relevant to the origin if so desired? If so, nice fix - then we didn't process it properly before, I suppose.

Vinnl (commented):
I thought this introduced a bug, but now I'm not sure - it might actually be resolving another one. See my comments.

src/resource/solidDataset.ts
NSeydoux: I actually learnt on Friday that if the relative IRI starts with a `/`, it is relative to the origin of the target URL (that's the case in NSS), and if it starts directly with a path element, it is relative to the complete target URL. Both cases are handled by the `URL` constructor, as well as the case when `locationHeader` is an absolute IRI (as it is the case with ESS), when the `base` should be ignored.

So I think this resolves a potential bug actually :)

src/resource/solidDataset.test.ts
NSeydoux: Exactly :)

src/resource/solidDataset.ts
Vinnl: Weird that I had a test case without a preceding slash before then, but indeed:

> A relative (to the request URL) or absolute URL.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Location

Checking NSS, it doesn't indeed seem to include the leading slash, and ESS returns the entire URL.

PR #947, opened 2021-03-29
Strawman proposal for plain-object representation of SolidDatasets

Vinnl:
This PR allows for a preliminary review of the plain-object representation of SolidDatasets to identify major issues that need addressing now. The desired outcome is being ready to start implementing; not to have nailed down the perfect final data structure right away, not to have something production-ready right away (so this PR will not be merged), and not for it to be ready for re-use outside of solid-client.

To address the issues we identified, the design goals of this proposal are

- `JSON.parse(JSON.stringify(data))` is lossless
- Optimised for reading and copying in an expected shape in stateful (Solid) applications
- `fromRdfJsDataset(toRdfJsDataset(data))` is lossless

The tests verify the latter claim, but **the relevant parts to review are [the type definitions in `src/pojo-interfaces.ts`](https://github.com/inrupt/solid-client-js/pull/947/files#diff-e7aee4a28e22c3fcb899cccabf8a23e2925752948633a3a571f2b1857da003a1R28-R57)**.

### DIFF
--- a/src/pojo-interfaces.test.ts
+++ b/src/pojo-interfaces.test.ts
@@ -276,6 +276,42 @@ describe("fromRdfJsDataset", () => {
     const thereAndBackAgain = toRdfJsDataset(fromRdfJsDataset(rdfJsDataset));
     expect(thereAndBackAgain.size).toBe(rdfJsDataset.size);
   });
+
+  it("does not trip over circular blank nodes", () => {
+    const namedNode = DataFactory.namedNode("https://example.com/namedNode");
+    const blankNode1 = DataFactory.blankNode();
+    const blankNode2 = DataFactory.blankNode();
+    const blankNode3 = DataFactory.blankNode();
+    const predicate = DataFactory.namedNode("https://example.com/predicate");
+    const literalString = DataFactory.literal("Arbitrary literal string");
+    const quads = [
+      DataFactory.quad(namedNode, predicate, blankNode2),
+      DataFactory.quad(blankNode1, predicate, blankNode2),
+      DataFactory.quad(blankNode2, predicate, blankNode3),
+      DataFactory.quad(blankNode3, predicate, blankNode1),
+      DataFactory.quad(blankNode2, predicate, literalString),
+    ];
+    const rdfJsDataset = dataset(quads);
+    const thereAndBackAgain = toRdfJsDataset(fromRdfJsDataset(rdfJsDataset));
+    expect(thereAndBackAgain.size).toBe(rdfJsDataset.size);
+  });
+
+  it("does not trip over Datasets that only contain Blank Node Subjects", () => {
+    const blankNode1 = DataFactory.blankNode();
+    const blankNode2 = DataFactory.blankNode();
+    const blankNode3 = DataFactory.blankNode();
+    const predicate = DataFactory.namedNode("https://example.com/predicate");
+    const literalString = DataFactory.literal("Arbitrary literal string");
+    const quads = [
+      DataFactory.quad(blankNode1, predicate, blankNode2),
+      DataFactory.quad(blankNode2, predicate, blankNode3),
+      DataFactory.quad(blankNode3, predicate, blankNode1),
+      DataFactory.quad(blankNode2, predicate, literalString),
+    ];
+    const rdfJsDataset = dataset(quads);
+    const thereAndBackAgain = toRdfJsDataset(fromRdfJsDataset(rdfJsDataset));
+    expect(thereAndBackAgain.size).toBe(rdfJsDataset.size);
+  });
 });
 
 describe("toRdfJsDataset", () => {
--- a/src/pojo-interfaces.test.ts
+++ b/src/pojo-interfaces.test.ts
@@ -30,7 +30,11 @@ import {
   serializeInteger,
   xmlSchemaTypes,
 } from "./datatypes";
-import { fromRdfJsDataset, toRdfJsDataset } from "./pojo-interfaces";
+import {
+  fromRdfJsDataset,
+  ImmutableDataset,
+  toRdfJsDataset,
+} from "./pojo-interfaces";
 
 describe("fromRdfJsDataset", () => {
   const fcNamedNode = fc
@@ -328,6 +332,32 @@ describe("toRdfJsDataset", () => {
 
     expect(fcResult.counterexample).toBeNull();
     expect(fcResult.failed).toBe(false);
+  });
+
+  it("can represent dangling Blank Nodes", () => {
+    const datasetWithDanglingBlankNodes: ImmutableDataset = {
+      type: "Dataset",
+      graphs: {
+        default: {
+          "_:danglingSubjectBlankNode": {
+            type: "Subject",
+            url: "_:danglingSubjectBlankNode",
+            predicates: {
+              "http://www.w3.org/ns/auth/acl#origin": {
+                blankNodes: [{}],
+              },
+            },
+          },
+        },
+      },
+    };
+
+    const rdfJsDataset = toRdfJsDataset(datasetWithDanglingBlankNodes);
+    expect(rdfJsDataset.size).toBe(1);
+    const quad = Array.from(rdfJsDataset)[0];
+    expect(quad.subject.termType).toBe("BlankNode");
+    expect(quad.predicate.value).toBe("http://www.w3.org/ns/auth/acl#origin");
+    expect(quad.object.termType).toBe("BlankNode");
   });
 });
 
--- a/src/pojo-interfaces.test.ts
+++ b/src/pojo-interfaces.test.ts
@@ -255,6 +255,27 @@ describe("fromRdfJsDataset", () => {
       ).size
     ).toBe(1);
   });
+
+  it("does not lose any predicates", () => {
+    const blankNode1 = DataFactory.blankNode();
+    const blankNode2 = DataFactory.blankNode();
+    const blankNode3 = DataFactory.blankNode();
+    const blankNode4 = DataFactory.blankNode();
+    const predicate1 = DataFactory.namedNode("https://example.com/predicate1");
+    const predicate2 = DataFactory.namedNode("https://example.com/predicate2");
+    const predicate3 = DataFactory.namedNode("https://example.com/predicate3");
+    const acrGraph = DataFactory.namedNode("https://example.com/acrGraph");
+    const literalString = DataFactory.literal("Arbitrary literal string");
+    const quads = [
+      DataFactory.quad(blankNode1, predicate1, blankNode2, acrGraph),
+      DataFactory.quad(blankNode2, predicate2, blankNode3, acrGraph),
+      DataFactory.quad(blankNode3, predicate3, blankNode4, acrGraph),
+      DataFactory.quad(blankNode4, predicate2, literalString, acrGraph),
+    ];
+    const rdfJsDataset = dataset(quads);
+    const thereAndBackAgain = toRdfJsDataset(fromRdfJsDataset(rdfJsDataset));
+    expect(thereAndBackAgain.size).toBe(rdfJsDataset.size);
+  });
 });
 
 describe("toRdfJsDataset", () => {
--- a/src/pojo-interfaces.test.ts
+++ b/src/pojo-interfaces.test.ts
@@ -33,7 +33,9 @@ import {
 import { fromRdfJsDataset, toRdfJsDataset } from "./pojo-interfaces";
 
 describe("fromRdfJsDataset", () => {
-  const fcNamedNode = fc.webUrl().map((url) => DataFactory.namedNode(url));
+  const fcNamedNode = fc
+    .webUrl({ withFragments: true, withQueryParameters: true })
+    .map((url) => DataFactory.namedNode(url));
   const fcString = fc.string().map((value) => DataFactory.literal(value));
   const fcInteger = fc
     .integer()
@@ -74,7 +76,7 @@ describe("fromRdfJsDataset", () => {
     )
     .map(([value, lang]) => DataFactory.literal(value, lang));
   const fcArbitraryLiteral = fc
-    .tuple(fc.string(), fc.webUrl())
+    .tuple(fc.string(), fc.webUrl({ withFragments: true }))
     .map(([value, dataType]) =>
       DataFactory.literal(value, DataFactory.namedNode(dataType))
     );
@@ -221,7 +223,10 @@ describe("toRdfJsDataset", () => {
     return Object.keys(value).length > 0;
   };
   const fcLiterals = fc
-    .dictionary(fc.webUrl(), fc.set(fc.string(), { minLength: 1 }))
+    .dictionary(
+      fc.webUrl({ withFragments: true }),
+      fc.set(fc.string(), { minLength: 1 })
+    )
     .filter(isNotEmpty);
   const fcLangStrings = fc
     .dictionary(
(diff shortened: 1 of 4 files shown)


src/pojo-interfaces.ts
```diff
@@ -0,0 +1,418 @@
...
+      blankNodes: blankNodes,
+    });
+  }
+
+  throw new Error(
+    `Objects of type [${quad.object.termType}] are not supported.`
+  );
+}
+
+function addRdfJsBlankNodeObjectsToPredicates(
```
NSeydoux, line 246: The only concern I have with the current approach to blank nodes is that, if I understand correctly, blank nodes would be effectively duplicated each time they appear as an object. In some cases, that may result in fairly large objects being cloned and passed around. I suspect with native `Object` methods being used, the overhead will not be too consequent, so that may be fine. I don't really see a satisfactory way (as in, not too hacky) to give blank node an identifier that would be consistent from one serialization to the other to enable comparison anyway, so we can simply wait and see if that becomes an issue in the future.

NSeydoux (approved):
That looks like a very good draft ! It looks quite "uncontroversial" to me (famous last words), which is really a compliment: nothing jumps to me as being too much of a radical choice, or something that's really counter-intuitive in order to materialize an RDFJS dataset into an immutable JSON object with the properties being listed in the scope.

src/pojo-interfaces.test.ts
```diff
@@ -0,0 +1,332 @@
...
+    .map(([subject, predicate, object, graph]) =>
+      DataFactory.quad(subject, predicate, object, graph)
+    );
+  const fcTerm = fc.oneof(fcNamedNode, fcLiteral, fcBlankNode, fcQuad);
+  const fcDataset = fc.set(fcQuad).map((quads) => dataset(quads));
+
+  it("can represent all Quads", () => {
+    const blankNode1 = DataFactory.blankNode();
+    const blankNode2 = DataFactory.blankNode();
+    const subject1Iri = "https://some.pod/resource#subject1";
```
pmcb55, line 108: Since `subject1` must be an IRI (by virtue of being an RDF subject at all), I'd suggest naming this local var `subject1Str`, or `subject1AsString`:
```suggestion
    const subject1AsString = "https://some.pod/resource#subject1";
```

src/pojo-interfaces.ts
```diff
@@ -0,0 +1,418 @@
...
+      quad.object.value,
+    ]);
+    return freeze({
+      ...objects,
+      namedNodes: namedNodes,
+    });
+  }
+
+  if (quad.object.termType === "Literal") {
+    if (quad.object.datatype.value === xmlSchemaTypes.langString) {
```
pmcb55, line 195: This is just a small thing, but we really should fix this ASAP (as it's 'infecting' this code too now) - i.e., `langString` is not an XML Schema type, it's defined in the [RDF vocab](https://www.w3.org/1999/02/22-rdf-syntax-ns#).

src/pojo-interfaces.ts
```diff
@@ -0,0 +1,418 @@
...
+
+export function fromRdfJsDataset(rdfJsDataset: RdfJs.DatasetCore): Dataset {
+  const dataset: Dataset = {
+    graphs: {},
+    type: "Dataset",
+  };
+
+  const quads = Array.from(rdfJsDataset);
+  const quadsWithoutBlankNodes = quads.filter(
+    (quad) => !isBlankNode(quad.subject)
```
pmcb55, line 74: Are we deliberately only filtering on Subject BNodes? i.e., should we also be filtering on Object BNodes too...?
Probably warrants a comment regardless (i.e., why are we filtering out quads with BNodes at all?), and how do these quads get added back in (i.e., I assume they do get added back in!)...?

src/pojo-interfaces.test.ts
```diff
@@ -0,0 +1,332 @@
...
+  serializeBoolean,
+  serializeDatetime,
+  serializeDecimal,
+  serializeInteger,
+  xmlSchemaTypes,
+} from "./datatypes";
+import { fromRdfJsDataset, toRdfJsDataset } from "./pojo-interfaces";
+
+describe("fromRdfJsDataset", () => {
+  const fcNamedNode = fc.webUrl().map((url) => DataFactory.namedNode(url));
```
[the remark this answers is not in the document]
Vinnl: Oh good one, I didn't realise they weren't included by default. And I think query parameters should be enabled too, since e.g. `?ext=acl` is also a Resource's name. 28f34221

src/pojo-interfaces.test.ts
```diff
@@ -0,0 +1,332 @@
...
+  const fcGraph = fc.oneof(fcDefaultGraph, fcNamedNode);
+  const fcQuadSubject = fc.oneof(fcNamedNode, fcBlankNode);
+  const fcQuadPredicate = fcNamedNode;
+  const fcQuadObject = fc.oneof(fcNamedNode, fcLiteral, fcBlankNode);
+  const fcQuad = fc
+    .tuple(fcQuadSubject, fcQuadPredicate, fcQuadObject, fcGraph)
+    .map(([subject, predicate, object, graph]) =>
+      DataFactory.quad(subject, predicate, object, graph)
+    );
+  const fcTerm = fc.oneof(fcNamedNode, fcLiteral, fcBlankNode, fcQuad);
```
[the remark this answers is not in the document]
Vinnl: Yeah I figured this might be useful for that, but then I realised that comparing different instances of RDF/JS Datasets was a chore and did not implement that.

src/pojo-interfaces.ts
```diff
@@ -0,0 +1,418 @@
...
+import * as RdfJs from "rdf-js";
+import { IriString } from "./interfaces";
+import { xmlSchemaTypes } from "./datatypes";
+
+const localNodeSkolemPrefix = "https://inrupt.com/.well-known/sdk-local-node/" as const;
+type LocalNodeIri = `${typeof localNodeSkolemPrefix}${string}`;
+export type LocalNodeName = string;
+export type Objects = Readonly<
+  Partial<{
+    literals: Readonly<Record<IriString, readonly string[]>>;
```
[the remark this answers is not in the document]
Vinnl: I think so! e65b862bc30f4f4980d78ef5157e034f35f2a250

src/pojo-interfaces.ts
```diff
@@ -0,0 +1,418 @@
...
+ */
+const freeze: typeof Object.freeze = Object.freeze;
+
+export function fromRdfJsDataset(rdfJsDataset: RdfJs.DatasetCore): Dataset {
+  const dataset: Dataset = {
+    graphs: {},
+    type: "Dataset",
+  };
+
+  const quads = Array.from(rdfJsDataset);
```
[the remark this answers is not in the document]
Vinnl: According to the spec, yes :)

src/pojo-interfaces.ts
Vinnl: Yep, my guess here would be that it is uncommon to have many quads with the same blank node as the object in the use cases in which you'd use an immutable dataset (at that point it's easier, and I'd encourage, to reach for named nodes), so being able to deduplicate them across different instances and not having to use hacky workarounds outweighs that in my view.

To have a 1:1 correspondence the other way around (i.e. from an arbitrary RDF/JS Dataset to this one), we'd probably still at some point need arbitrary unstable names for blank nodes that only appear in either Subject or Object position but not both, or just reject them at point of import.

Either way there's nothing blocking any such approach in the future, so we can still change it if implementation experience teaches provides new insights.

src/pojo-interfaces.test.ts
Vinnl: e0cb294d49cf5a3f88f4ac8c0653d7ad057a04c9

src/pojo-interfaces.ts
Vinnl: I won't do it in this PR since it's not intended to be merged, but I wouldn't worry too much about it "infecting": it's internal-only and can be renamed to e.g. `knownDataTypeIriStringAll` (or whatever name you prefer) in one go in your IDE.

src/pojo-interfaces.ts
Vinnl: b7714c88f200cc1cf61659f185125a49fb4490ca

src/pojo-interfaces.test.ts
pmcb55: I added a comment on that commit (i.e., my suggestion was less work than the changes there :) !): https://github.com/inrupt/solid-client-js/commit/e0cb294d49cf5a3f88f4ac8c0653d7ad057a04c9#r48915352

src/pojo-interfaces.ts
pmcb55: Yeah, I wasn't trying to suggest fixing it as part of this PR, I just meant fixed in the main codebase. As you say, it's an easy fix - maybe we have ticket already, but I couldn't find one, so I created one (internal ID: SDK-1911).

src/pojo-interfaces.ts
```diff
@@ -0,0 +1,422 @@
...
+
+export type Subject = Readonly<{
+  type: "Subject";
+  url: LocalNodeIri | IriString;
+  predicates: Predicates;
+}>;
+
+export type Graph = Readonly<Record<IriString, Subject>>;
+
+export type Dataset = Readonly<{
```
pmcb55, line 56: So as discussed today, this seems like a straightforward suggestion (i.e., to reflect the overall intent of the PR):
```suggestion
export type ImmutableDataset = Readonly<{
```

src/pojo-interfaces.ts
```diff
@@ -0,0 +1,422 @@
...
+  type: "Dataset";
```
pmcb55, line 57: So as discussed today, this seems like a straightforward suggestion (i.e., to reflect the overall intent of the PR):
```suggestion
export type ImmutableDataset = Readonly<{
  type: "ImmutableDataset";
```

Vinnl: 6fc0730754b8c3efd3168a868298697e4052e1fd

src/pojo-interfaces.test.ts
Vinnl: Responded here: https://github.com/inrupt/solid-client-js/pull/947#commitcomment-48932899

The change: 002ec66

ajacksified (commented):
Generally looks good to me - we'll need to carefully profile the final implementation for large datasets (and as you noted, Object.freeze will likely cause performance issues).

I'd also be curious to see how you envision handling WithChangelog and WithResourceInfo. Would `getSolidDataset` return an ImmutableDataset with the metadata as graph nodes (the resource info, the changelog, and the actual dataset) or as properties on the object as they are now?

Vinnl:
> I'd also be curious to see how you envision handling WithChangelog and WithResourceInfo. Would getSolidDataset return an ImmutableDataset with the metadata as graph nodes (the resource info, the changelog, and the actual dataset) or as properties on the object as they are now?

We could definitely add them to a graph, although before that we'll probably want to add them to the graph in the `toRdfJsDataset` function, since that is then the actual external representation, making it available to others as RDF. But definitely room to switch those two around if we want that: they're independent pieces of work, so we can implement them whenever after we implement this.

src/pojo-interfaces.test.ts
```diff
@@ -0,0 +1,351 @@
...
+      DataFactory.quad(subject1, predicate1, literalLangString, defaultGraph),
+      DataFactory.quad(subject1, predicate1, literalInteger, defaultGraph),
+      DataFactory.quad(subject1, predicate2, subject2, defaultGraph),
+      DataFactory.quad(subject2, predicate1, blankNode1, acrGraph),
+      DataFactory.quad(subject2, predicate1, blankNode2, acrGraph),
+      DataFactory.quad(blankNode1, predicate1, literalString, acrGraph),
+      DataFactory.quad(blankNode2, predicate1, literalString, acrGraph),
+      DataFactory.quad(blankNode2, predicate1, literalInteger, acrGraph),
+      DataFactory.quad(blankNode2, predicate2, literalInteger, acrGraph),
+    ];
```
pmcb55, line 149: Seems to be a bug here - if I add the single line below here, the test still passes (i.e., as the code is filtering out quads with BNodes as Subject, and only adding them back in if that Subject BNode appears as the object of another quad, but shouldn't this quad appear in the output too...?):
```js
      DataFactory.quad(DataFactory.blankNode(), predicate2, literalInteger, acrGraph),
```

Vinnl: Yeah, I addressed that here: https://github.com/inrupt/solid-client-js/pull/947#discussion_r603932077

The tl;dr is: my judgment was that that does not appear in practice (i.e. when would a Dataset include the above Quad?), so I went with Turtle's approach, but full 1:1 correspondence can be achieved by also adding the ability to add labelled blank nodes for "dangling" blank nodes, i.e. expanding

```
    blankNodes: readonly Predicates[];
```

into

```
    blankNodes: readonly Predicates[] | DanglingBlankNode; // With DanglingBlankNode = `_${string}`;
```

pmcb55:
> Would getSolidDataset return an ImmutableDataset with the metadata as graph nodes
And don't forget `getSolidDatasetWithAcl()` and `getSolidDatasetWithAcr()` and `getSolidDatasetWithAccessDatasets()`.
And I'm not sure how adding the current `WithResourceInfo` or `ChangeLog` or `WithAcl` or `WithAcr`, etc. would be independent pieces of work - i.e., translating a current SolidDataset that has an attached `WithResourceInfo` into an RDF/JS dataset would necessarily lose all that `WithResourceInfo` data, right?
So there's currently no lossless round-tripping (i.e., translating from SolidDataset -> RDF/JS Dataset -> SolidDataset -> RDF/JS Dataset -> etc.) of `WithResourceInfo` datasets, or `WithAcl`, or `WithAcr`, etc. - is that right?

src/pojo-interfaces.ts
```diff
@@ -0,0 +1,424 @@
...
+      const predicates = graph[subjectIri].predicates;
+      const subjectNode = RdfJsDataFactory.namedNode(subjectIri);
+      quads.push(...subjecToRdfJsQuads(predicates, subjectNode, graphNode));
+    });
+  });
+
+  return quads;
+}
+
+function subjecToRdfJsQuads(
```
pmcb55, line 356:
```suggestion
function subjectToRdfJsQuads(
```

src/pojo-interfaces.ts
```diff
@@ -0,0 +1,424 @@
...
+ * Non-goals:
+ * - Optimising for streaming data
+ * - Optimising for unknown use cases
+ * - Be perfect from the get-go
+ * - Be immediately useful outside of solid-client
+ * @module
+ */
+
+import { dataset as rdfJsDataset } from "@rdfjs/dataset";
+import RdfJsDataFactory from "@rdfjs/data-model";
```
pmcb55, line 23: I think these 2 lines might be wrong, as both are specific implementations of the RDF/JS Dataset and Data Model respectively, and not the pure interfaces. For example, I think where the Data Model is being used should be replaced by just the pure Data Model interface (i.e., you're already correctly importing `rdf-js` on the next line), and the signatures of `toRdfJsQuads()` and `toRdfJsDataset()` need to be updated to take in as a parameter the specific DataFactory instance of whatever RDF/JS implementation the developer happens to want to use.

src/pojo-interfaces.ts
Vinnl: Heh: f2230f58

src/pojo-interfaces.ts
Vinnl: As mentioned in the PR description, this isn't meant to be a production implementation of translation functions, but rather just to get a feel for what working with the new types would look like, so that's why I didn't implement passing in your own DataFactory.

I did that now though: 1e2a4056

Note that the imports are still there, because of course we don't want to _require_ developers to always pass in their own; we use those implementations by default.

Vinnl:
> And I'm not sure how adding the current WithResourceInfo or ChangeLog or WithAcl or WithAcr, etc. would be independent pieces of work - i.e., translating a current SolidDataset that has an attached WithResourceInfo into an RDF/JS dataset would necessarily lose all that WithResourceInfo data, right?

Yes, but since we don't currently have a way to translate those into an RDF/JS Dataset either, there's no regression there. So what I was trying to say is that we can implement the types defined here without breaking anything; what pieces of work we implement when is of course a separate issue.

> So there's currently no lossless round-tripping (i.e., translating from SolidDataset -> RDF/JS Dataset -> SolidDataset -> RDF/JS Dataset -> etc.) of WithResourceInfo datasets, or WithAcl, or WithAcr, etc. - is that right?

There's no such thing as "WithResourceInfo datasets" - WithResourceInfo is not necessarily attached to a SolidDataset. So no, the round-tripping is indeed only for SolidDatasets.

src/pojo-interfaces.test.ts
Vinnl: I realised that we didn't even have to do anything special for Blank Nodes that only appear in Object position, since an empty `Predicates` (i.e. analogous to Turtle's empty `[]`) suffices there, so I only had to add support for them in Subject position: d2cf7c6e3ea74235fea36c1dd3dcce59d0917f9a

If you pull down the latest version and add your example Quad, you should see that it now does appear in the output.

Vinnl: As per your request, I also added a test that demonstrates that Lists can be represented: 8bf22219

PR #954, opened 2021-03-30
Prevent GitHub Actions from masking Pod URLs used in end-to-end tests

Vinnl:
Since the username is still an independent secret, this still doesn't look great:

```
Authenticated end-to-end tests against Pod [pod-compat.inrupt.com/***-compat/] and OIDC Issuer [broker.pod-compat.inrupt.com]:
```

https://github.com/inrupt/solid-client-js/runs/2228169943?check_suite_focus=true

But I suppose it's better than nothing?

### DIFF
--- a/src/e2e-node/e2e.test.ts
+++ b/src/e2e-node/e2e.test.ts
@@ -329,16 +329,22 @@ type AuthDetails = [Pod, OidcIssuer, ClientId, ClientSecret, RefreshToken];
 const serversUnderTest: AuthDetails[] = [
   // pod.inrupt.com:
   [
-    process.env.E2E_TEST_ESS_POD!,
-    process.env.E2E_TEST_ESS_IDP_URL!,
+    // Cumbersome workaround, but:
+    // Trim `https://` from the start of these URLs,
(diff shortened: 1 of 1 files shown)


src/e2e-node/e2e.test.ts
```diff
@@ -352,6 +358,8 @@ const serversUnderTest: AuthDetails[] = [
 describe.each(serversUnderTest)(
   "Authenticated end-to-end tests against Pod [%s] and OIDC Issuer [%s]:",
   (rootContainer, oidcIssuer, clientId, clientSecret, refreshToken) => {
+    rootContainer = "https://" + rootContainer;
```
pmcb55, line 361: Maybe just a matching comment to explain why we have to stitch back in the `https://` parts here...?

pmcb55 (approved):

src/e2e-node/e2e.test.ts
Vinnl: Good one: 854876a8

Vinnl:
(Also just realised I'd accidentally included #955 in this one - removed now.)

PR #958, opened 2021-04-01
Lock version of n3 to <1.9.0

### LINKED ISSUE #957: 400 Bad Request when storing a SolidDataset
### Search terms you've used

400 Bad Request, changelog, empty delete

### Bug description

The request does not seem to match the changelog.

Source: https://forum.solidproject.org/t/curious-400-error-when-removing-using-inrupt/4163

image

### To Reproduce

Not sure yet, will investigate...

1.
2.
3.
4.

**Minimal reproduction**

Not reproduced yet.

### Expected result

SolidDataset to be saved properly.

### Actual result

400 Bad Request.

### Environment

```
```

Vinnl:
There's no clear changelog, but there were changes that at least
affected our tests (see https://github.com/inrupt/solid-client-js/pull/918#issuecomment-800901838), and also seem to result in no longer being able to generate the correct Turtle from a SolidDataset's ChangeLog. (It results in a string consisting of just a single period (`.`).)

This PR fixes #957.

- [ ] I've added a unit test to test for potential regressions of this bug. (Probably N/A - this might be a bug in N3 directly, but will have to investigate that.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ The following changes have been implemented but not released yet:
 
 - Saving a dataset to an IRI with a hash fragment (e.g. the WebID) is processed
   as an update, and not anymore as an overwrite.
+- An update in one of our dependencies caused writes to a Pod to fail. This
+  dependency has now been pinned to an older, working version while we
+  investigate further. For more info, see
+  https://github.com/inrupt/solid-client-js/issues/957.
 
 The following sections document changes that have been released already:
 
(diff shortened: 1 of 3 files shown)


Vinnl:
CI builds being annoying all of a sudden - dependencies should be the same, but for some reason all of a sudden compilation fails in the browser-based end-to-end tests.

Either way, here's a confirmation that this does indeed fix the reported issue: https://forum.solidproject.org/t/curious-400-error-when-removing-using-inrupt/4163/6?u=vincent

NSeydoux (approved):

PR #967, opened 2021-04-06
Make CD job succeed for Dependabot

Vinnl:
Before, I marked the job as a whole as being skippable for
Dependabot, because it is not required for Pull Requests. However,
since other jobs that *are* required depend on it, skipping it
would also skip those jobs.

Thus, now only a step in the prepare-deployment job is skipped when
the actor is dependabot, so that the job as a whole still appears
to succeed.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -14,10 +14,6 @@ env:
   CI: true
 jobs:
   prepare-deployment:
-    # Dependabot does not have access to our secrets,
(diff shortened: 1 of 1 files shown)


NSeydoux (approved):
I assume that for these jobs that are dependant on this ones are also not skipped as a whole, but only skip the steps which require dependabot secrets, right ?

Vinnl:
Yes, although that's all of them - they're either publishing to npm (requiring the npm auth token), or testing the published package.

PR #969, opened 2021-04-06
Do not attempt to unpublish unpublished packages

Vinnl:
Dependabot updates are not published to npm by default, so there's
no need to unpublish those after merging.

### DIFF
--- a/.github/workflows/cd-teardown.yml
+++ b/.github/workflows/cd-teardown.yml
@@ -1,4 +1,4 @@
-name: CD
+name: CD — unpublish
 
 on: delete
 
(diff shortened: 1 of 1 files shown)


NSeydoux (approved):
For `solid-client-authn`, my approach to this was to skip the CD teardown when the actor is dependabot: if the CD goes through, and the PR is merged using the `@dependabot merge` command, then the actor doing the merge is dependabot. If manual intervention is required, dependabot won't merge the PR (I think), so that actor will be whoever merges. Do you think this approach would work too, and are there any specific reasons to approach this otherwise ?

Vinnl:
Yeah, I initially went with that approach as well, but then I realised the same thing as in that other PR: I don't think Dependabot is the actor when merging things to `main` for solid-client, because it keeps complaining it doesn't have access to merge it, and thus I always hit the merge button myself rather than asking Dependabot to merge.

PR #990, opened 2021-04-14
Temporarily disable scheduled CI runs

Vinnl:
They're currently consistently failing, generating a lot of noise. We can re-enable them once that's died down.

To manually run all tests, visit [the "CI" workflow page](https://github.com/inrupt/solid-client-js/actions/workflows/ci.yml), then hit "Run workflow", running it against the `main` branch:

image

This will also run our browser-based end-to-end tests, which can be a bit more flaky, but it's most important that the step `Run npm run e2e-test-node` completes successfully.

/cc @davidboweninrupt, @acoburn, @jholleran (I couldn't assign you as reviewers for some reason)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,10 +4,11 @@ on:
   push:
   # Allow manual triggering, e.g. to run end-to-end tests against Dependabot PRs:
   workflow_dispatch:
-  schedule:
+  # This is temporarily disabled while an issue is getting resolved:
+  # schedule:
   # Run every fifth minute.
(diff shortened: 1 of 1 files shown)


Vinnl:
I've just disabled _all_ GitHub Actions for now in the settings, since we'll have to apply a fix on our side, i.e. the failing tests aren't informative until then.

To re-enable them, go to the [Actions settings](https://github.com/inrupt/solid-client-js/settings/actions) and set "Actions permissions" to "Allow select actions" (ticking "Allow actions created by GitHub").

PR #991, opened 2021-04-15
CI refinements

Vinnl:
This:

- ~~Updates the TestCafe selectors for ESS's new approval screen~~ Done in #997.
- Have the Linux TestCafe tests in CI have a special user agent for easy identification in the logs.
- Log every request sent in the TestCafe tests and make them available for download. [Example.](https://github.com/inrupt/solid-client-js/actions/runs/753223963)

Note that these are all long-term fixes; i.e. this branch does not include the short-term workarounds needed to pass CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,6 +67,8 @@ jobs:
     - name: Run browser-based end-to-end tests (Linux)
       # TODO: Add Edge/merge with Windows setup once Edge is available on Linux:
       # Connections are flakier in CI, so retry failed network requests with --retry-test-pages.
+      # That also requires adding `--hostname localhost` to run over HTTPS:
+      # https://testcafe.io/documentation/402638/reference/configuration-file#retrytestpages
       # Setting a test-specific user agent is only possible for Chrome: https://stackoverflow.com/a/59358925
       run: npm run e2e-test-browser -- "firefox:headless,chrome:headless:userAgent='Browser-based solid-client end-to-end tests running in CI. Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) $(chrome --version) Safari/537.36'" --screenshots takeOnFails=true,path=./e2e-browser-failures --retry-test-pages --hostname localhost
       # The Node version does not influence how well our tests run in the browser,
@@ -86,6 +88,8 @@ jobs:
     - name: Run browser-based end-to-end tests (Windows)
       continue-on-error: true
       # Connections are flakier in CI, so retry failed network requests with --retry-test-pages.
+      # That also requires adding `--hostname localhost` to run over HTTPS:
+      # https://testcafe.io/documentation/402638/reference/configuration-file#retrytestpages
       run: npm run e2e-test-browser -- edge:headless,firefox:headless,chrome:headless --retry-test-pages --hostname localhost
       # The Node version does not influence how well our tests run in the browser,
       # so we only need to test in one.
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -67,7 +67,8 @@ jobs:
     - name: Run browser-based end-to-end tests (Linux)
       # TODO: Add Edge/merge with Windows setup once Edge is available on Linux:
       # Connections are flakier in CI, so retry failed network requests with --retry-test-pages.
-      run: npm run e2e-test-browser -- firefox:headless,chrome:headless --retry-test-pages --hostname localhost
+      # Setting a test-specific user agent is only possible for Chrome: https://stackoverflow.com/a/59358925
+      run: npm run e2e-test-browser -- "firefox:headless,chrome:headless:userAgent='Browser-based solid-client end-to-end tests running in CI. Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) $(chrome --version) Safari/537.36'" --retry-test-pages --hostname localhost
```
pmcb55, line 71: Just curious - why add the `--hostname localhost` option...?

src/e2e-browser/e2e.testcafe.ts
```diff
@@ -45,7 +47,31 @@ config({
...
+    await promises.mkdir(resolve(__dirname, "testcafe-requests"), {
+      recursive: true,
+    });
+    await promises.writeFile(
+      resolve(
+        __dirname,
+        "testcafe-requests",
+        `testcafe-requests-${Date.now()}.log`
+      ),
+      JSON.stringify(requestLogger.requests, undefined, 2)
```
pmcb55, line 72: Yeah, this capability (capturing requests/responses) is super-valuable I think. But when I do this on AuthN though it's spewing out 350MB per run! It's easy to tweak the output of course though - which I did for AuthN, and some very interesting insights in the continuing intermittent failures, which we can discuss offline (i.e., I'd suggest reducing the output being generated here before merging).

.github/workflows/ci.yml
Vinnl: Good question, should've been added to the comment I suppose, cause I had to look it up too :) Looks like it's required with `--retry-test-pages`: https://devexpress.github.io/testcafe/documentation/reference/configuration-file.html#retrytestpages

I'll try to remember adding that later today.

src/e2e-browser/e2e.testcafe.ts
Vinnl: The TestCafe tests for solid-client are pretty limited, so it's about 3MB per browser/OS combo (so with three Windows browsers, two on Linux and one on MacOS, that comes down to 18MB). That sounds pretty manageable, so I'd propose not throwing away data here.

.github/workflows/ci.yml
Vinnl: 03c7f48c61d26e050df5010168999bfa52934cbf

.github/workflows/ci.yml
```diff
@@ -145,13 +152,20 @@ jobs:
...
       uses: actions/upload-artifact@v2.2.3
       continue-on-error: true
       if: failure() && github.event_name != 'schedule'
       with:
         name: e2e-browser-failures
         path: e2e-browser-failures
+    - name: Archive browser-based end-to-end test request logs
+      uses: actions/upload-artifact@v2.2.3
+      continue-on-error: true
+      if: (failure() || success()) && matrix.node-version != '10.x' && matrix.node-version != '12.x' && github.event_name != 'schedule' && github.actor != 'dependabot[bot]'
```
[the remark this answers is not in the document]
Vinnl: I'm not sure exactly how it works, but by default, jobs without `if: failure()` do not get executed after a previous job fails. However, I was afraid that just adding that would result in the job not getting executed when there was _no_ failure, so I added the `|| success()` to be sure. (Trying out the different variations was too much effort :P )

PR #997, opened 2021-04-20
Temporarily disable end-to-end tests against ESS

Vinnl:
Inrupt's Enterprise Solid Server changed its authentication
mechanism to invalidate the refresh token after every use,
returning a new one to use instead. This does not mesh well with
our CI setup, which used a token set up once in the CI's
environment secrets.

The server team will probably be able to arrange a new setup for us
that does not require maintaining state across test runs, so while
we await that, we'll disable automated tests against it for a short
while.

This PR also includes an upgrade of `jose` which was flagged by `npm audit` and was therefore also failing CI: https://www.npmjs.com/advisories/1661
(Ironically, we only use that in our end-to-end tests :) )

Additionally, I've included the updated selectors for the browser-based tests from #991 for the updated broker screen.

Edit: I also added taking screenshots on failed tests so I could diagnose that I'd restored the _wrong_ secret to the broker's IDP URL >.<

### DIFF
--- a/src/e2e-browser/pageModels/broker.ts
+++ b/src/e2e-browser/pageModels/broker.ts
@@ -23,40 +23,20 @@ import { t, ClientFunction, Selector } from "testcafe";
 import { screen } from "@testing-library/testcafe";
 
 export class BrokerPage {
-  accessOpenIdCheckbox: Selector;
-  accessOfflineCheckbox: Selector;
-  accessWebIdCheckbox: Selector;
-  rememberForeverRadioButton: Selector;
-  rememberOneHourRadioButton: Selector;
-  rememberNotRadioButton: Selector;
   authoriseButton: Selector;
   denyButton: Selector;
 
   constructor() {
-    this.accessOpenIdCheckbox = screen.getByLabelText(
-      "log in using your identity"
-    );
-    this.accessOfflineCheckbox = screen.getByLabelText("offline access");
-    this.accessWebIdCheckbox = screen.getByLabelText("solid webid");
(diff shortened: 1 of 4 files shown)


src/e2e-browser/pageModels/broker.ts
```diff
@@ -23,40 +23,20 @@ import { t, ClientFunction, Selector } from "testcafe";
...
-    );
-    this.authoriseButton = screen.getByText("Authorize");
+    this.authoriseButton = screen.getByText("Approve");
     this.denyButton = screen.getByText("Deny");
   }
 
   async authoriseOnce() {
     await onAuthorisePage();
-    await t.click(this.rememberNotRadioButton).click(this.authoriseButton);
+    await t.click(this.authoriseButton);
```
pmcb55, line 36: Probably needs a general search-and-replace (e.g., line 34 too!)
```suggestion
    await t.click(this.authorizeButton);
```

Vinnl: The codebase internally has mostly used British English spelling for pretty much everything - not sure if it's feasible to change it all in one go (don't think I'd even be able to - I wouldn't know which are which), but let's not change them on a case-by-case basis.

pmcb55 (approved):
(I had assumed you'd disable the E2E tests via the GitHub action, but instead you just added '.skip' to the test suite instead - just out of curiosity, but did you have any reason for that preference...?!)

Vinnl:
@pmcb55 Yep! I initially went with that approach, but we also have e2e tests running against NSS, and I didn't want to disable those.

PR #999, opened 2021-04-22
Use root TypeScript version in TestCafe tests

Vinnl:
Without this, TestCafe would use the version of TypeScript in
/node_modules/testcafe/node_modules/typescript. The compiler path
is resolved relative to the TestCafe module path, so by referring
to the `typescript` folder in the parent of the TestCafe module,
we get the top-level TypeScript module.

(The trigger for this PR was https://github.com/inrupt/solid-client-authn-js/pull/1293 - TestCafe used an older version of TypeScript that SCAB breaks with.)

### DIFF
--- a/.testcaferc.json
+++ b/.testcaferc.json
@@ -6,5 +6,10 @@
   "clientScripts": [
     { "module": "@testing-library/dom/dist/@testing-library/dom.umd.js" }
   ],
+  "compilerOptions": {
+    "typescript": {
+      "customCompilerModulePath": "../typescript"
+    }
+  },
   "skipJsErrors": false
 }


pmcb55:
Hi Vinnie - just had a chat with NicAS on this one - can we schedule some time to discuss the details of this (e.g., how exactly was SCAB breaking?), as it seems this is really valuable knowledge for developers generally (so we also need to discuss how to capture this knowledge (e.g., as part of our Coding Guidelines, or our test suite README, or Confluence or something (since we can't add detailed comments to the TestCafe config itself (since it's pesky JSON that doesn't allow comments (so maybe a justification to move to JS config instead?!))).

Vinnl:
@pmcb55 I usually default to a JS config, but unfortunately it seems TestCafe doesn't support this.

The "problem" is that TestCafe has a dependency on TypeScript v3, whereas solid-client uses v4. When npm encounters that, it doesn't try to deduplicate the dependencies, it just installs both versions at the same time. This was generally not a problem. However, Nicolas [made use of a feature only introduced in TypeScript v4](https://github.com/inrupt/solid-client-authn-js/pull/1293) in SCA, which meant that its type definitions could only be parsed by TypeScript v4.

So I did two things:
- Submitted a PR to SCA to allow it to work with older versions of TS as well.
- Made sure that TestCafe uses the same instance of TypeScript to compile the solid-client code like we do when publishing it, when running tests. (That's this PR.)

You can see the breakage in action by looking at TestCafe recent runs, for example https://github.com/inrupt/solid-client-js/runs/2418345897?check_suite_focus=true

NSeydoux (approved):

PR #1000, opened 2021-04-22
Add getLinkedResourceUrlAll() and getEffectiveAccess()

### LINKED ISSUE #765: solid.getPermissions for ds.internal_resourceInfo.permissions
It would be nice to not have to use internal api to getPermissions on a datasource

### LINKED ISSUE #766: client.getTypes instead of  ds.internal_resourceInfo.linkedResources.type
Would be nice to have a function like client.getTypes to get the types of a datasource instead of using internal api

Vinnl:
This PR fixes #765 and fixes #766.

# New feature description

This PR adds a function `getEffectiveAccess` that allows developers to check what access the current user (and, if supported by the server, unauthenticated users) have to the given Resource. This enables e.g. reflecting in the UI that the current user is not allowed to change the given Resource.

Since it depended on reading Link headers for ACP, I decided to implement it in a way that fixed #766 while I was at it.

Note that we had initially decided to call this `getAccess`, but since we already have such a function in the universal access API (which requires being able to read ACRs or ACLs, so it differs from this one), I thought it might be better to give this a different name.

~~CI is currently failing because of #999.~~

# Checklist

- [ ] All acceptance criteria are met. - I renamed it from `getAccess` to `getEffectiveAccess` to distinguish it from the `getAccess` function in the universal access API.

### DIFF
--- a/src/resource/resource.ts
+++ b/src/resource/resource.ts
@@ -30,6 +30,7 @@ import {
   hasServerResourceInfo,
   SolidClientError,
   LinkedResourceUrlAll,
+  EffectiveAccess,
 } from "../interfaces";
 import { internal_toIriString } from "../interfaces.internal";
 import { fetch } from "../fetcher";
@@ -37,6 +38,7 @@ import {
   internal_isUnsuccessfulResponse,
   internal_parseResourceInfo,
 } from "./resource.internal";
+import { acp } from "../constants";
 
 /** @ignore For internal use only. */
 export const internal_defaultFetchOptions = {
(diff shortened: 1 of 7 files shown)


src/resource/resource.ts
```diff
@@ -176,6 +178,87 @@ export function isPodOwner(
...
+ * implementing Web Access Control, Resources can have an Access Control List
+ * Resource linked to it via the `acl` relation.
+ *
+ * @param resource A Resource fetched from a Solid Pod.
+ * @returns The URLs of Resources linked to the given Resource, indexed by the key that links them.
+ * @since Not released yet.
+ */
+export function getLinkedResourceUrlAll(
+  resource: WithServerResourceInfo
+): internal_LinkedResourceUrls {
```
NSeydoux, line 194: If the type is part of the external API, is it quite right that it is prefixed `internal_` ? Is it until we agree on it and make it part of the official API ?

src/resource/resource.ts
```diff
@@ -176,6 +178,87 @@ export function isPodOwner(
...
+    read: boolean;
+    append: boolean;
+    write: boolean;
+  };
+  public?: {
+    read: boolean;
+    append: boolean;
+    write: boolean;
+  };
+};
```
NSeydoux, line 217: Would this have its place in the  `interface.ts` file ?

NSeydoux (approved):
Some superficial comments, but overall looks good to me.

src/resource/resource.ts
Vinnl: No good catch, that would have resulted in annoying `any`s breaking type checking for developers. Fixed in 3f19b833.

src/resource/resource.ts
Vinnl: Yeah good point. We should probably consider at some point just getting rid of `interfaces.ts` altogether and just move the interfaces to the relevant modules, but as long as we haven't done that, best to be consistent: 911fa11f

PR #1011, opened 2021-04-26
Further increase TestCafe handleRedirect timeout

Vinnl:
It's currently timing out a lot in CI, so let's try giving it some
more time.


Vinnl:
Well what do you know, it actually passed CI in one go :D

NSeydoux (approved):

PR #1012, opened 2021-04-26
Allow for ridiculous timeouts when loggin in in CI

Vinnl:
The servers seem particularly slow today, and CI probably isn't
helping either, so here's a desperate attempt to make it less
flaky.

Also this was @NSeydoux's idea, so he should share the blame. Proof:

image


NSeydoux (approved):
I'm not sure I appreciate finger-pointing, and also how you interpret memes I send you is your responsability :p.

Vinnl:
I'm starting to think the timeout isn't the problem though:

image

PR #1017, opened 2021-04-28
DOCS-296: Use shared docs-assets

kay-kim:
Use a shared public [inrupt/docs-assets](https://github.com/inrupt/docs-assets) repo for our docs css/templates/theme.  Currently added as a git submodule (to get it to work with vercel, added using https instead of ssh).  Give a holler if you prefer we don't use git submodules. (We could also just tweak the docs makefile to just git clone locally during the build rather than use a submodule.)

Rendered: https://lit-pod-25ihz6338-inrupt.vercel.app/


Vinnl (changes requested):
Could you remove `docs/api/.DS_Store`?

And yeah, if this is only used in the build script, I think I'd prefer not to use submodules. The reason is that I think that this would mean that every time you update the shared repo, Git would report there being uncommitted changes after running the build script. If that assumption is incorrect then let me know :)

kay-kim:
Will update to remove submodules and .DS_Store.  I think if  I remove the `git submodule update --remote` Git might not report about uncommitted changes to the  submodules, but then, of course, we'd be running w. out-of-date submodules 😄

kay-kim:
Updated to remove submodules.

Vinnl (approved):
Thanks! Glad to have all those files ejected from the repo :)

PR #1036, opened 2021-05-12
Make SolidDatasets and Things atomic objects

Vinnl:
This implements the data structure proposed in #947. It's still in draft while that is awaiting its final feedback round, but it'd be good to already collect @NSeydoux's feedback on the actual implementation while we await that, since that's been lingering for a while.

While I was working on the parsing, I extracted out the parser interface to enable passing your own parsers for other data types (JSON-LD, Trig, whatever). I haven't mentioned that in the changelog/docs yet though, since we might want to give that interface some more thought.

Likewise, I haven't exposed the functions yet to convert to and from an RDF/JS Dataset. We'll probably want to give that some explicit attention before doing so too, and not having that is not a regression anyway. ~~I'll still have to inspect the API docs to make sure that those new types/functions are not exposed yet.~~ Edit: done.

I added two behaviour changes that I thought make sense and would not break downstreams - they're listed in the changelog under "Changed". Feel free to add comments there about whether they're a good idea.

The E2E tests are still succeeding without modification, so that's a good sign. ~~I'll try it later on Penny and PodBrowser too.~~ Edit: Done, Penny and PodBrowser work too!


src/resource/resource.internal.ts
```diff
@@ -136,49 +135,13 @@ export function internal_cloneResource<ResourceExt extends object>(
...
-      typeof value.constructor === "undefined" ||
-      value.constructor.name !== "Object"
-    ) {
-      return;
-    }
-
-    copy[key] = value;
-  });
-
-  return copy;
```
Vinnl, line 181: Good riddance to this hack and inefficient beast.

src/resource/solidDataset.ts
```diff
@@ -49,28 +45,159 @@ import {
...
+    },
+  });
+}
+
+export type Parser = {
+  onQuad: (onQuadCallback: (quad: Quad) => void) => void;
+  onError: (onErrorCallback: (error: unknown) => void) => void;
+  onComplete: (onCompleteCallback: () => void) => void;
+  parse: (source: string, resourceInfo: WithServerResourceInfo) => void;
+};
```
NSeydoux, line 90: Out of curiosity, besides n3, have you considered any parser that could instantiate this interface ?

NSeydoux (commented):
I'm not quite done with the review yet and will pick it up Monday :)

CHANGELOG.md
```diff
@@ -6,6 +6,22 @@ The following changes have been implemented but not released yet:
...
+
+- `getUrl` now returns local URLs (e.g. #my-thing)
+- `getThing` now returns empty Things if they were added, rather than `null`.
+- The internal structure of SolidDatasets has been reworked. This shouldn't
+  affect you if you relied on just our public API's, but if you were
+  manipulating the data directly, you might need to update to use our public
+  API's. If something's not covered that you need, please [file an
+  issue](https://github.com/inrupt/solid-client-js/issues/new/choose).
+
+### New features
```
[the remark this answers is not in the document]
Vinnl: Mutability didn't work already (i.e. would break as soon as you tried to save the SolidDataset), and for that reason RDF/JS compatibility was never advertised. Since breaking changes only concern public API's, this particular change is not breaking in the SemVer sense that it warrans a major version bump.

The changes listed under "Changes" _can_ theoretically be considered breaking changes. However, since major version bumps are intended to communicate to downstreams that they'll have to pay extra attention to the new release and carefully read the release notes for potential migration paths, and since as far as I'm aware nobody will _actually_ need to make any code changes to accommodate those changes, I think we should not treat them as such.

src/resource/solidDataset.ts
Vinnl: One thing to note is that n3 does not conform to this interface. Instead, I have considered other parsers in the sense that theoretically, you should be able to add a wrapper similar to the one I wrote for n3 to make other parsers work as well. However, I would like to actually write those for at least JSON-LD and a different Turtle parser before making this part of our official public API.

(So add the functionality now, but list it in the changelog and API docs later.)

src/thing/get.ts
```diff
@@ -51,15 +56,15 @@ export function getUrl(
...
+    throw new ValidPropertyUrlExpectedError(property);
+  }
+  const predicateUrl = internal_toIriString(property);
+  const firstUrl = thing.predicates[predicateUrl]?.namedNodes?.[0] ?? null;
+  if (firstUrl === null) {
     return null;
   }
-
-  return matchingQuad.object.value;
+  return isLocalNodeIri(firstUrl) ? `#${getLocalNodeName(firstUrl)}` : firstUrl;
```
NSeydoux, line 67: \``#${getLocalNodeName(firstUrl)}`\` is going to be a construct that may be duplicated a bunch of times. It it worth, for readability, to factor it out in a function `serializeLocalNodeName` ?

src/thing/remove.ts
```diff
@@ -64,13 +60,14 @@ export function removeAll(thing: Thing, property: Url | UrlString): Thing {
   if (!internal_isValidUrl(property)) {
     throw new ValidPropertyUrlExpectedError(property);
   }
-  const predicateNode = asNamedNode(property);
+  const predicateIri = internal_toIriString(property);
+  const newPredicates = { ...thing.predicates };
+  delete newPredicates[predicateIri];
```
NSeydoux, line 65: Hurray for efficient and compact removal ! (Not the only place, but worth mentioning somewhere)

NSeydoux (approved):

CHANGELOG.md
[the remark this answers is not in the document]
NSeydoux: (so apparently you cannot remove a reaction emoji, even in case of misclick ^^)

NSeydoux:
I don't have any significant feedback on the implementation of this feature: the internal code touching datasets is improved, and this adds both efficiency and safety to our codebase.

CHANGELOG.md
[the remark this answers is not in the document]
Vinnl: Are you sure? I just removed one by clicking the reaction itself.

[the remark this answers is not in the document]
NSeydoux: There must have been a bug: I clicked on the +1, it removed both ^^

src/rdf.internal.ts
```diff
@@ -0,0 +1,585 @@
...
+  const chainBlankNodes = blankNodeSubjects
+    .concat(blankNodeObjects)
+    .filter((blankNode) => {
+      if (
+        cycleBlankNodes.some((cycleBlankNode) =>
+          cycleBlankNode.equals(blankNode)
+        )
+      ) {
+        return false;
+      }
```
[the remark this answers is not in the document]
Vinnl: Given the complexity I definitely should have done this the first time around. Of course, since it's recursion it's still fairly complex, but I gave it a go here: cb213c7b.

src/rdf.internal.ts
```diff
@@ -0,0 +1,585 @@
...
+  const dataset: ImmutableDataset = {
+    graphs: { default: {} },
+    type: "Dataset",
+  };
+
+  const quads = Array.from(rdfJsDataset);
+
+  const chainBlankNodes = getChainBlankNodes(quads);
+
+  // Quads with chain Blank Nodes as their Subject will be parsed when those
```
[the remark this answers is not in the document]
Vinnl: Also attempted to explain in cb213c7b.

src/thing/get.ts
Vinnl: I think it only occurs here. It's a bit of an odd one out too, because most of the time local references should already be fully resolved after saving the SolidDataset. The only time this occurs is when you call `getIri` (or `getIriAll`) to obtain a URL you just added to a Thing - if you've fetched it, or already saved it, you'll get the full URL, and our own codebase just works with the skolemised URL.

src/resource/solidDataset.ts
Vinnl: Not-yet-public interfaces hidden in 5e5a5e4a2410c967c4a5cce63cbe1b8365949c75.

PR #1043, opened 2021-05-18
Re-enable Node-based end-to-end tests against ESS

Vinnl:
~~This includes the fix from #1042, since that bug arises in our end-to-end tests and thus they wouldn't succeed without it. It's probably easiest to review this after that is merged, or just [look at the commit after that](https://github.com/inrupt/solid-client-js/pull/1043/commits/bf425c5289453f851339587d5eb747b0fff4c973).~~ #1042's been merged.

This replaces our devDependency on SCAN with the version from https://github.com/inrupt/solid-client-authn-js/pull/1329. That allows us to use client credentials to authenticate in our end-to-end tests. This is needed because the server was updated to rotate refresh tokens every time they were used, blocking the ability to store that token once and re-use it for every test. Instead, the client credentials that are currently used have been preregistered with ESS to enable them to be longer-lived.

This is not a long-term solution since obviously not everybody can go and ask the server host to preregister their client, which is why we're depending on a preview version of SCAN. Once a longer term solution has been designed and implemented, we can switch to that.

(This is our only usage of SCAN, so depending on the preview version should not cause issues elsewhere.)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4400,9 +4400,9 @@
       }
     },
     "@inrupt/solid-client-authn-core": {
-      "version": "1.8.0",
-      "resolved": "https://registry.npmjs.org/@inrupt/solid-client-authn-core/-/solid-client-authn-core-1.8.0.tgz",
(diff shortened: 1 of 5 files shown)


pmcb55 (approved):

pmcb55 (approved):

Vinnl:
Superseded by #1181.

PR #1063, opened 2021-06-01
Fix missing of build-time errors for asUrl()

Vinnl:
With ThingLocal now being a subset of ThingPersisted thanks to its
use of a skolemised URL, editors could no longer recognise that a
ThingLocal passed to asUrl() also required a baseUrl. With this
somewhat involved type definition, developers will once again have
red squigglies if they forget to pass that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ The following changes have been implemented but not released yet:
 
 ## [Unreleased]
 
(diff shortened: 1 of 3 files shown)


Vinnl:
Never mind, looks like I was counting my eggs before they hatched - I'm still running into errors elsewhere...

PR #1068, opened 2021-06-03
Fix missing build-time error for asUrl()

Vinnl:
With ThingLocal now being a subset of ThingPersisted thanks to its
use of a skolemised URL, editors could no longer recognise that a
ThingLocal passed to asUrl() also required a baseUrl. With this
updated type definition, developers will once again have
red squigglies if they forget to pass a baseUrl together with a
ThingLocal.

Unfortunately, a Thing of which it is unknown whether it's a
ThingLocal or a ThingPersisted is indistinguishable from a
ThingPersisted, so in those cases developers will still have to
rely on the runtime error.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ The following changes have been implemented but not released yet:
 
 ## [Unreleased]
 
+### Bugs fixed
+
+- Since version 1.8.0, TypeScript would no longer warn you if you omitted the
+  second argument to `asUrl` in cases where it's required, thereby risking
+  runtime errors. In most of these cases, TypeScript should now warn you again.
+
 The following sections document changes that have been released already:
 
 ## [1.8.1] - 2021-05-25
--- a/src/thing/thing.test.ts
+++ b/src/thing/thing.test.ts
@@ -801,6 +801,36 @@ describe("asIri", () => {
     );
   });
 
+  it("triggers a TypeScript error when passed a ThingLocal without a base IRI", () => {
+    const localThing: ThingLocal = createThing();
+
+    // @ts-expect-error
+    expect(() => asUrl(localThing)).toThrow();
+  });
+
+  // This currently fails because a plain `Thing` always has a `url` property that is a string,
+  // and is therefore indistinguishable from a `ThingPersisted`. Not sure what the solution is yet.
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,13 @@ The following changes have been implemented but not released yet:
 
 ## [Unreleased]
 
+### Bugs fixed
+
+- Since version 1.8.0, TypeScript would no longer warn you if you omitted the
+  second argument to `asUrl` in cases where it was should not be omitted,
```
pmcb55, line 12: Just to void the double negative:
```suggestion
  second argument to `asUrl` in cases where it's required,
```

src/thing/thing.test.ts
```diff
@@ -801,6 +801,36 @@ describe("asIri", () => {
...
+    const localThing: ThingLocal = createThing();
+
+    // @ts-expect-error
+    expect(() => asUrl(localThing)).toThrow();
+  });
+
+  // This currently fails because a plain Thing always has a `url` property that is a string,
+  // and is therefore indistinguishable from a ThingPersisted. Not sure what the solution is yet.
+  // Meanwhile TS users won't get a build-time error if they're passing a regular Thing,
+  // which is annoying but not a major issue.
```
pmcb55, line 814: I'm a bit unclear on the terminology used here, i.e., 'plain thing', 'regular thing', 'persisted thing' - and it doesn't even mention a 'local thing'. So is a 'plain thing' the same as a 'regular thing' the same as a 'local thing'...?

Vinnl: Yep, I reworded that to use "plain" twice instead of switching to "regular", and added backticks around the type names (`Thing` and `ThingPersisted`) to hopefully clarify I'm referring to those raw types: 2ba8d5ee.

Basically, we went from `ThingLocal` and `ThingPersisted` from being distinct types, and `Thing` being the union of those two`, to `ThingLocal` being a subset of `ThingPersisted` (i.e. it has a full URL, but that URL is a skolem that indicates that it's a local URL) and `Thing` being equal to `ThingPersisted`. Not great and I hope we'll find a solution to that at some point, but it's not a major issue either.

CHANGELOG.md
Vinnl: 2ba8d5ee

pmcb55 (approved):

PR #1075, opened 2021-06-07
Upgrade libraries dependabot missed

Vinnl:
I'm not sure why, but every so often Dependabot seems to lose track of some libraries (TypeScript seems to be affected relatively often). This bumps the versions of Jest ~~and TypeScript~~ because Dependabot didn't.


Vinnl:
Removed TypeScript upgrade because TypeDoc is having issues with the new version for now: https://github.com/TypeStrong/typedoc/issues/1589

Once that gets released I'll create a new PR for that.

pmcb55 (approved):

PR #1080, opened 2021-06-08
Fix issues identified by npm audit

Vinnl:
Specificially: https://www.npmjs.com/advisories/1751

`npm audit fix` could automatically fix some transitive dependencies, but it advised upgrading to a version of testcafe that [does not appear to be intended for use](https://github.com/DevExpress/testcafe/issues/4885#issuecomment-601124382). I used [npm-force-resolutions](https://www.npmjs.com/package/npm-force-resolutions) to upgrade the dependency there.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7212,7 +7212,7 @@
         "fast-deep-equal": "^3.1.3",
         "file-entry-cache": "^6.0.1",
         "functional-red-black-tree": "^1.0.1",
-        "glob-parent": "^5.1.2",
+        "glob-parent": "5.1.2",
         "globals": "^13.6.0",
         "ignore": "^4.0.6",
         "import-fresh": "^3.0.0",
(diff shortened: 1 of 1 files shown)


pmcb55 (approved):
Jeez - these manual edits of the `package-lock.json` sure are very scary ('cos doesn't it mean we can't just delete and re-generate it and be sure all is well, even if you just want to work locally?).
Approving this anyway, but I'm just not sure what a longer-term approach to these dependency problems should be - any ideas...?

Vinnl:
It's not _super_ manual (as in, I use `npm-force-resolutions`), and I don't think it should cause further problems in the sense that future `npm installs` should not revert this - "just deleting and regenerating it" is a good way to cause "works on my machine" issues, so I wouldn't do that anyway.

That said, I agree it's not great. Unfortunately, I don't have better ideas either :/ Not blocking things on `npm audit` failures is an idea, but I wouldn't classify that as "better".

PR #1083, opened 2021-06-09
Resolve npm audit resolutions once and for all

Vinnl:
This adds npm-force-resolutions ~~as a dependency~~ and automatically
runs it before installing something, to hopefully keep transitive
dependencies at a non-vulnerable version even if their parent
dependencies have not updated yet, even across multiple dependency
updates.

(Edit: adding it as a dependency doesn't work, since it runs before installing something - like itself. Ran with `npx` now and pinned the version there. Unfortunately that won't be recognised by dependabot, but I guess it's the best I can do for now.)

I'm not sure yet if a preinstall script will work for maintaining
it across dependabot updates and with `npm ci` runs, but I figured
we can at least try and see if it works.

Note that this would not remove the need to run `npm audit fix` or add a resolution every now and then, but at least once a resolution has been set, it should hopefully not need to be set again.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12504,6 +12504,17 @@
               "dev": true,
               "requires": {
                 "is-glob": "^4.0.1"
+              },
+              "dependencies": {
+                "is-glob": {
+                  "version": "3.1.0",
+                  "resolved": "https://registry.npmjs.org/is-glob/-/is-glob-3.1.0.tgz",
+                  "integrity": "sha1-e6WuJCF4BKxwcHuWkiVnSGzD6Eo=",
+                  "dev": true,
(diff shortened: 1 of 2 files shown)


pmcb55 (approved):
It might be an idea to write up all the learnings you seem to gathering on all this @Vinnl. Seems like once you have a good idea on how best to handle it, we should share with Jack, Ruben and Matthieu (i.e., with all the other JS codebases we have), since I assume the principle ideas will apply equally to any JS codebase, right...!?

Vinnl:
OK, seems like this isn't the solution either, since it breaks for people not using npm lockfiles...

PR #1088, opened 2021-06-14
Date data type

### LINKED ISSUE #1076: Add support for `date` data type
### Search terms you've used
date, data type, addDate, getDate, setDate, removeDate

### Feature suggestion

There currently are `datetime` functions in solid-client, but none to specify a date without a time.

### Expected functionality/enhancement

Functions `getDate`, `addDate`, `setDate` and `removeDate` that are like their `datetime` equivalents, but disregard the time, and store the value as an `xsd:date`.

### Actual functionality/enhancement

Currently you're forced to create a `datetime`.

### Use Cases

Some times there's no time that would be relevant - e.g. for all-day events, like stating when Easter is this year.

### Additional information

To implement this:

- Add `xsd:date` to the `schemaTypes` we support [in `datatypes.ts`](https://github.com/inrupt/solid-client-js/blob/2ea2ab5b447bae24e5913b32520f6c960bdbac62/src/datatypes.ts#L34).
- Add `serializeDate` and `deserializeDate` to `datatypes.ts` in a way similar to [its datetime equivalents](https://github.com/inrupt/solid-client-js/blob/2ea2ab5b447bae24e5913b32520f6c960bdbac62/src/datatypes.ts#L69-L154), based on the spec that determines [what the stringified format of an `xsd:date` looks like](https://www.w3.org/TR/xmlschema-2/#date-lexical-representation).
- Add the relevant functions to `src/thing/get.ts`, `add.ts`, `set.ts` and `remove.ts` similar to the datetime equivalents.
- Don't forget to update the tests as well :)

Also see #1077 for a similar issue that you can pick up after this.

dgbarclay:
# Get/Set/Add/Remove methods for date data type

Fixes #1076.

# Checklist

- [] Relevant documentation, if any, has been written/updated.


src/datatypes.ts
```diff
@@ -153,6 +156,75 @@ export function deserializeDatetime(literalString: string): Date | null {
...
+  );
+
+  // For the year, values from 0 to 99 map to the years 1900 to 1999. Since the serialisation
+  // always writes out the years fully, we should correct this to actually map to the years 0 to 99.
+  // See
+  // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date#Individual_date_and_time_component_values
+  if (utcFullYear >= 0 && utcFullYear < 100) {
+    // Note that we base it on the calculated year, rather than the year that was actually read.
+    // This is because the year might actually differ from the value listed in the serialisation,
+    // i.e. when moving the timezone offset to UTC pushes it into a different year:
```
Vinnl, line 222: I think this comment is irrelevant for `date` - since there's no time, a timezone offset can't push it into a different year. (i.e. in a timezone of `+02:00`, a time of `23:00` could mean the actual date could differ - but if there's no time, it can't be `23:00` either :)

src/datatypes.ts
```diff
@@ -153,6 +156,75 @@ export function deserializeDatetime(literalString: string): Date | null {
...
+ * @see https://www.w3.org/TR/xmlschema-2/#date-lexical-representation
+ */
+export function deserializeDate(literalString: string): Date | null {
+  // Date in the format described at
+  // https://www.w3.org/TR/xmlschema-2/#date-lexical-representation
+  // (without constraints on the value).
+  // -? - An optional leading `-`.
+  // \d{4,}- - Four or more digits followed by a `-` representing the year. Example: "3000-".
+  // \d\d-\d\d - Two digits representing the month and two representing the day of the month,
+  //             separated by a `-`. Example: "11-03".
```
Vinnl, line 186: Here, too, I think it's missing the fact that it can optionally also contain a timezone (i.e. `Z`, or `+01:30` or `-02:00`).

src/thing/add.test.ts
```diff
@@ -607,6 +608,173 @@ describe("addDatetime", () => {
   });
 });
 
+describe("addDate", () => {
+  it("adds the given datetime value for the given predicate", () => {
```
Vinnl, line 612:
```suggestion
  it("adds the given date value for the given predicate", () => {
```

src/thing/add.test.ts
```diff
@@ -607,6 +608,173 @@ describe("addDatetime", () => {
...
+      thing,
+      "https://some.vocab/predicate",
+      new Date(Date.UTC(1990, 10, 12))
+    );
+
+    expect(
+      updatedThing.predicates["https://some.vocab/predicate"].literals![
+        "http://www.w3.org/2001/XMLSchema#date"
+      ]
+    ).toStrictEqual(["1990-11-12T00:00:00.000Z"]);
```
Vinnl, line 625: I think this is the result of the `serializeDate` function just calling `toISOString()`, but I think this should be:

```suggestion
    ).toStrictEqual(["1990-11-12Z"]);
```

src/thing/get.test.ts
```diff
@@ -871,6 +990,136 @@ describe("getDatetimeAll", () => {
...
+
+    expect(
+      getDateAll(
+        thingWithDates,
+        DataFactory.namedNode("https://some.vocab/predicate")
+      )
+    ).toEqual([expectedDate1, expectedDate2]);
+  });
+
+  it("returns an empty array if no datetime values were found", () => {
```
Vinnl, line 1027: There's a couple of other mentions of `datetime`s below too :)

```suggestion
  it("returns an empty array if no date values were found", () => {
```

src/thing/remove.test.ts
```diff
@@ -1064,6 +1065,230 @@ describe("removeDatetime", () => {
...
+
+    expect(
+      updatedThing.predicates["https://some.vocab/predicate"].literals
+    ).toStrictEqual({
+      "http://www.w3.org/2001/XMLSchema#date": [],
+    });
+  });
+
+  it("does not modify the input Thing", () => {
+    const thingWithDatetime = getMockThingWithLiteralFor(
```
Vinnl, line 1110: In this file too check for references to dateTime in your date tests :)

```suggestion
    const thingWithDate = getMockThingWithLiteralFor(
```

src/thing/set.ts
```diff
@@ -106,6 +107,23 @@ export const setDatetime: SetOfType<Date> = (thing, property, value) => {
   return addDatetime(removeAll(thing, property), property, value);
 };
 
+/**
+ * Create a new Thing with existing values replaced by the given datetime for the given Property.
```
Vinnl, line 111:
```suggestion
 * Create a new Thing with existing values replaced by the given date for the given Property.
```

Vinnl (changes requested):
A good start, but as far as I can see the serialise function does not match the spec, and the deserialize function is throwing away the timezone information. If you can fix that (and add the related tests) then I think we're probably good to go :)

src/datatypes.ts
dgbarclay: In this case, can deserialize essentially ignore the timezone and just return a Date object with T00:00:00?

Vinnl: Yeah I think that's the way to go. It's a bit fiddly though - I'd set it to 0:00 in the timezone set in the serialisation (or UTC, if none was set), but since JS `Date` objects only support either the local time or UTC, the time you get locally might still be somewhat different (e.g. it'd be 22:00 the previous day here).

Vinnl: Actually, maybe set it to 12:00? That way I think it should always match the local and UTC date regardless of the timezone, and the time part should be ignored anyway.

(Dates in JavaScript really is a mess. There's a proposal for a replacement, [Temporal](https://tc39.es/proposal-temporal/docs/), that looks like it will be added to JavaScript somewhere in the coming few years. But that doesn't help us now...

dgbarclay: Yeah, I was wondering what would happen for a date with a - timezone for setting at 0:00. I'll change now.

src/datatypes.ts
```diff
@@ -153,6 +156,83 @@ export function deserializeDatetime(literalString: string): Date | null {
...
+      ? [-1, signedDateString.substring(1)]
+      : [1, signedDateString];
+  const [yearString, monthString, dayAndTimezoneString] = dateString.split("-");
+
+  const dayString = (dayAndTimezoneString.length > 2 ? dayAndTimezoneString.substring(0, 2) : dayAndTimezoneString);
+
+  const utcFullYear = Number.parseInt(yearString, 10) * yearMultiplier;
+  const utcMonth = Number.parseInt(monthString, 10) - 1;
+  const utcDate = Number.parseInt(dayString, 10);
+  const hour = 12;
```
Vinnl, line 214: Ah right, we're basically stripping the timezone information from the input string, because there's no way to represent that using `Date`. Reeeaallly looking forward to JS getting a better date&time API :angry:

src/thing/remove.ts
```diff
@@ -169,6 +170,28 @@ export const removeDatetime: RemoveOfType<Date> = (thing, property, value) => {
...
+ */
+export const removeDate: RemoveOfType<Date> = (thing, property, value) => {
+  internal_throwIfNotThing(thing);
+  // deserialize date will return Date object at midday (43200000 ms)
+  return removeLiteralMatching(
+    thing,
+    property,
+    xmlSchemaTypes.date,
+    (foundDate) =>
+      deserializeDate(foundDate)?.getTime() === (value.getTime() + 43200000)
```
Vinnl, line 191: Hmm.... So this line gets me thinking: if the developer has to pass in a `Date` object not only with the correct date, but also with a certain time - how much value is it providing that we're using a `Date` object here?

Would it potentially make more sense to use a custom plain object here using only the properties we're interested in? So something like this:

```
type Date = {
  year: number;
  month: number;
  day: number;
  timezoneHourOffset?: number;
  timezoneMinuteOffset?: number;
};
```

Alternatively, if that's cumbersome too, maybe we should only check on `foundDate`'s and `value`'s year, month and day rather than adding 43200000ms?

Sorry, I'm being a little annoying here bring this up after implementing it one way, but it's the implementation experience that brought the insight...

Vinnl (commented):
It mostly looks good, but I'm just being very annoying and having second thoughts about this approach - sorry! :/ I added a suggestion inline, but would love to hear your thoughts - and do speak up if you think it's rubbish, cause I'm certainly not sure about it.

As an aside, some process thingies to note that might be helpful in the future:

- If the PR is ready for another review, be sure to press the little "recycle" icon next to a reviewer's name, so they get a notification that they're expected to do something :)
image
- Before a PR can be merged, all checks should succeed.
image
You can click the "Details" link to see what went wrong - in this case, it looks like you didn't run Prettier on a number of files. This should generally run automatically when committing, so that's a bit weird. Either way, your editor probably can do it for you - in VSCode, you can find "Format Selection" if you press Ctrl+Shift+P.
- The PR template mentions "Commits in this PR are minimal and have descriptive commit messages." During review you can sometimes collect a few minor fixing commits, but please when merging squash them all together if the PR is essentially just a single change. (Ideally you'd rebase locally, creating the commits that make sense, but that does involve some more advanced Git knowledge :) ) Not relevant yet, but when finishing up the PR :)

src/thing/remove.ts
dgbarclay: Yeah, it's a fair point. On the set/get/add end, the user can pass in any date object and the serializing will handle all of the annoying date-related things. So, for the user, it's still fairly simple/easy. For example, the following code will add "1990-11-12Z" to the thing, ignoring the time :

```
const updatedThing = addDate(     
 thing,      
"https://some.vocab/predicate", 
 new Date(Date.UTC(1990, 10, 12, 3, 3, 3))
);
```
The difficulty, as you say, is on the remove side. In this case, they would have to provide the full date object with the time set to 0 in order to get a match, not ideal or intuitive.

I'm reluctant to completely ditch the Date object because I like how smoothly it works for appending to things, but for the remove method, we need a better solution. In my opinion, I think modifying the code to check for a date match should be enough. In that use case, a user just has to match the date they set originally which I think is pretty okay from a UX point of view, and, we can remove the clunky 43200000ms...

dgbarclay: Also, thanks for git tips, useful stuff to know for the future.

Vinnl: OK, sounds good!

And how about time: would a custom object make sense there, since it's somewhat more clunky to construct a `Date` just for the time?

dgbarclay: Yeah, I think that's a good idea. Returning the user 1970 Jan 1st attached to their time isn't nice. Something like this...

```
type Time = {
  hour: number;
  minute: number;
  second: number;
  timezoneHourOffset?: number;
  timezoneMinuteOffset?: number;
};
```

We'd need to document this change well, is that something I need to do before the PR is merged?

Vinnl: You're still working on time right? So it needn't block this PR.

For the time PR/branch, you can define that type and then export it from `index.ts` and add a TypeDoc comment block of documentation to that type as well.

dgbarclay: Yeah oops for the other PR. Okay, I'll do that.

Vinnl (approved):
Quite a few more edge cases here than I expected, but you did it! Thanks a lot :)

Vinnl:
Ah, and of course there were parallel changes in `main` that this has to align with too: abeda783.

src/thing/remove.ts
```diff
@@ -169,6 +170,36 @@ export const removeDatetime: RemoveOfType<Date> = (thing, property, value) => {
...
+    property,
+    xmlSchemaTypes.date,
+    function (foundDate) {
+      if (deserializeDate(foundDate)) {
+        return (
+          deserializeDate(foundDate)!.getFullYear() === value.getFullYear() &&
+          deserializeDate(foundDate)!.getMonth() === value.getMonth() &&
+          deserializeDate(foundDate)!.getDate() === value.getDate()
+        );
+      } else {
```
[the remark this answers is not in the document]
dgbarclay: Ahhh I see, thanks for sorting.

src/datatypes.test.ts
```diff
@@ -262,6 +264,45 @@ describe("deserializeDatetime", () => {
...
+  it("properly parses a serialised date", () => {
+    const expectedDate = new Date(Date.UTC(1990, 10, 12, 12));
+    expect(deserializeDate("1990-11-12Z")).toEqual(expectedDate);
+
+    const expectedDateWithNegativeYear = new Date(Date.UTC(-42, 10, 12, 12));
+    expect(deserializeDate("-0042-11-12Z")).toEqual(
+      expectedDateWithNegativeYear
+    );
+
+    const dateBeforeTheYear100 = new Date(0, 0, 1, 12);
```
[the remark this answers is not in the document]
dgbarclay: That's a fun one, always have reviewers in different timezones!!

PR #1089, opened 2021-06-14
Add buildThing()

Vinnl:
# New feature description

Although not yet a full-fledged fluent API for RDF data manipulation, this will at least allow developers to avoid bugs like this that we've seen people run into:

```javascript
let newThing = createThing();
addUrl(newThing, rdf.type, schema.Person);
addStringNoLocale(newThing, schema.givenName, "Vincent");
```

(i.e. forgetting to reassign to `newThing`, so it would still be empty.)

Instead, you can now create/update a Thing in a single expression:

```javascript
const newThing = buildThing()
  .addUrl(rdf.type, schema.Person)
  .addStringNoLocale(schema.givenName, "Vincent")
  .build();
```

`buildThing` can also take an existing Thing as an argument on which it will then operate, or parameters that it will pass to `createThing` internally.

I've also added a test that makes sure that if we ever add additional `add*`, `set*` and `remove*` functions, we'll be warned to also add it here.

# Checklist

### DIFF
--- a/src/thing/build.test.ts
+++ b/src/thing/build.test.ts
@@ -1,0 +1,238 @@
+/**
+ * Copyright 2020 Inrupt Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to use,
+ * copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the
+ * Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,
+ * INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A
+ * PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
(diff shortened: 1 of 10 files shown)


src/thing/build.test.ts
```diff
@@ -0,0 +1,217 @@
...
+      (adderName) => adderName.substring(0, 3) === "add"
+    );
+    expect.assertions(adderNames.length * 2);
+
+    const builder = buildThing();
+
+    adderNames.forEach((adderName) => {
+      expect(
+        (builder as Record<string, Function>)[adderName]
+      ).not.toBeUndefined();
```
pmcb55, line 175: Maybe there's a subtle difference, but is equivalent (and more readable)?
```suggestion
      ).toBeDefined();
```

src/thing/build.test.ts
```diff
@@ -0,0 +1,217 @@
...
+      "https://some.vocab/predicate",
+      1337
+    );
+    nonBuilderThing = adders.addStringWithLocale(
+      nonBuilderThing,
+      "https://some.vocab/predicate",
+      "Yet another string",
+      "en-gb"
+    );
+    nonBuilderThing = removers.removeStringWithLocale(
```
pmcb55, line 96: I'm not sure why all these calls to `remove` and `set` - isn't this test just asserting that the builder builds a dataset as expected?

ajacksified: Agreed - this test looks like it's trying to cover a _lot_ of functionality. Each function should have its own test, or failures will be difficult to diagnose - or potentially hidden by all of the functions being called.

src/thing/build.test.ts
Vinnl: Heh, that was mostly my mental model while writing it it seems, leading to me missing the equivalence. bc943d77

src/thing/build.test.ts
Vinnl: Yeah, I initially just wanted a relatively small test to verify that it did the same as one of them, since it's practically trivial that if one works, the rest will too. But then of course it was complaining about incomplete coverage, so I just added a couple of other calls.

Split up in 3f0f83c2.

pmcb55 (approved):

PR #1098, opened 2021-06-18
Updates per typedoc changes

kay-kim:
Fix missing module.md:
- Instead of copying specific directories over, do a bulk copy at the typedoc generated api/ directory to our build directory; and then remove the redundant index.md since we're using index.rst

### DIFF
--- a/docs/api/Makefile
+++ b/docs/api/Makefile
@@ -36,15 +36,15 @@ migrate: clean
 
    # Copying to SOURCECOPYDIR instead of copying source dir to BUILDDIR
    # in case someone forgets to backslash after build/
-   # Copying source/api/modules -> build/source/modules
-   # Copying source/api/classes -> build/source/classes
-   # Otherwise, after building the docs, the url is dev-tools/api/js/solid-client/api/modules or such.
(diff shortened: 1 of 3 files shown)


kay-kim:
@Vinnl  -- should I have just uploaded the patch directly to  https://github.com/inrupt/solid-client-js/pull/1097 ?

Vinnl:
@kay-kim Thanks for getting on this, I hadn't gotten around to looking at it yet. It appears that it might indeed need the changes in that PR as well to work - I'll try that and see if it then succeed.

kay-kim:
@Vinnl  -- Still unclear why it was complaining that the index.md doesn't exist even though it would have been copied.  I did add a file check for existence before removing, and that did get rid of the error and it builds normally now. But wasn't able to actually recreate the error locally to see why it complained before.

Vinnl:
@kay-kim I found out why it complained - I'd already run into the duplicate `index.md` before, and removed it in the `build-api-docs` script in `package.json` rather than in the Makefile. Removed that in 740ae12c722ff66fa8cacc058addca044f62715f.

PR #1111, opened 2021-06-24
Time data type

### LINKED ISSUE #1077: Add support for `time` data type
### Search terms you've used
time, data type, addTime, getTime, setTime, removeTime

### Feature suggestion

There currently are `datetime` functions in solid-client, but none to specify a time without a date.

### Expected functionality/enhancement

Functions `getTime`, `addTime`, `setTime` and `removeTime` that are like their `datetime` equivalents, but disregard the date, and store the value as an `xsd:time`.

### Actual functionality/enhancement

Currently you're forced to create a `datetime` with an arbitrary date.

### Use Cases

Some times there's no date that would be relevant - e.g. when storing when school typically closes.

### Additional information

To implement this:

- Add `xsd:time` to the `schemaTypes` we support [in `datatypes.ts`](https://github.com/inrupt/solid-client-js/blob/2ea2ab5b447bae24e5913b32520f6c960bdbac62/src/datatypes.ts#L34).
- Add `serializeTime` and `deserializeTime` to `datatypes.ts` in a way similar to [its datetime equivalents](https://github.com/inrupt/solid-client-js/blob/2ea2ab5b447bae24e5913b32520f6c960bdbac62/src/datatypes.ts#L69-L154), based on the spec that determines [what the stringified format of an `xsd:time` looks like](https://www.w3.org/TR/xmlschema-2/#time-lexical-representation).
- Add the relevant functions to `src/thing/get.ts`, `add.ts`, `set.ts` and `remove.ts` similar to the datetime equivalents.
- Don't forget to update the tests as well :)

Also see #1076 for a similar issue that you can pick up after this.

dgbarclay:
# Time data type
Support added for time data type. This includes set/get/add/remove methods. Fixes #1077.

# Checklist

### DIFF
--- a/src/datatypes.ts
+++ b/src/datatypes.ts
@@ -104,7 +104,7 @@ export function deserializeDatetime(literalString: string): Date | null {
   //                       the hour offset and two for the minute offset, separated by a `:`.
   //                       Example: "+13:37".
   const datetimeRegEx =
-    /-?\d{4,}-\d\d-\d\dT\d\d:\d\d:\d\d(\.\d+)?(Z|(\+|-)\d\d:\d\d)/;
+    /-?\d{4,}-\d\d-\d\dT\d\d:\d\d:\d\d(\.\d+)?(Z|(\+|-)\d\d:\d\d)?/;
   if (!datetimeRegEx.test(literalString)) {
     return null;
   }
@@ -230,12 +230,17 @@ export function deserializeDate(literalString: string): Date | null {
  * @returns A tuple [timeString, timezoneString].
  * @see https://www.w3.org/TR/xmlschema-2/#time-lexical-repr
  */
-function splitTimeFromTimezone(timeString: string): [string, string] {
+function splitTimeFromTimezone(timeString: string): [string, string?] {
   if (timeString.endsWith("Z")) {
     return [timeString.substring(0, timeString.length - 1), "Z"];
   }
   const splitOnPlus = timeString.split("+");
   const splitOnMinus = timeString.split("-");
+
+  if (splitOnPlus.length === 1 && splitOnMinus.length === 1) {
+    return [splitOnPlus[0], undefined];
+  }
+
   return splitOnPlus.length > splitOnMinus.length
     ? [splitOnPlus[0], "+" + splitOnPlus[1]]
     : [splitOnMinus[0], "-" + splitOnMinus[1]];
--- a/src/datatypes.ts
+++ b/src/datatypes.ts
@@ -33,6 +33,7 @@ export const xmlSchemaTypes = {
   boolean: "http://www.w3.org/2001/XMLSchema#boolean",
   dateTime: "http://www.w3.org/2001/XMLSchema#dateTime",
   date: "http://www.w3.org/2001/XMLSchema#date",
+  time: "http://www.w3.org/2001/XMLSchema#time",
   decimal: "http://www.w3.org/2001/XMLSchema#decimal",
   integer: "http://www.w3.org/2001/XMLSchema#integer",
   string: "http://www.w3.org/2001/XMLSchema#string",
@@ -65,6 +66,131 @@ export function deserializeBoolean(value: string): boolean | null {
   } else {
     return null;
   }
+}
+
+/**
+ * Time type for time data type attributes
+ */
+export type Time = {
+  hour: number;
+  minute: number;
+  second: number;
+  millisecond?: number;
+  timezoneHourOffset?: number;
+  timezoneMinuteOffset?: number;
+};
+
+/**
+ * @internal
+ * @param value Value to serialise.
+ * @returns String representation of `value` in UTC.
+ * @see https://www.w3.org/TR/xmlschema-2/#time-lexical-repr
+ */
+export function serializeTime(value: Time): string {
+  let millisecondString;
+  let timezoneString;
+
+  if (value.millisecond) {
+    if (value.millisecond < 10) {
+      millisecondString = "00" + value.millisecond;
+    } else if (value.millisecond < 100) {
+      millisecondString = "0" + value.millisecond;
+    } else {
+      millisecondString = value.millisecond;
+    }
+  }
+
+  if (typeof value.timezoneHourOffset === "number") {
+    const timezoneFormatted =
+      Math.abs(value.timezoneHourOffset) < 10
+        ? "0" + Math.abs(value.timezoneHourOffset)
+        : Math.abs(value.timezoneHourOffset);
+
+    timezoneString =
+      value.timezoneHourOffset >= 0
+        ? "+" + timezoneFormatted
+        : "-" + timezoneFormatted;
+
+    if (value.timezoneMinuteOffset) {
+      timezoneString =
+        timezoneString +
+        ":" +
+        (value.timezoneMinuteOffset < 10
+          ? "0" + value.timezoneMinuteOffset
+          : value.timezoneMinuteOffset);
+    } else {
+      timezoneString = timezoneString + ":00";
+    }
+  }
+
+  return (
+    (value.hour < 10 ? "0" + value.hour : value.hour) +
+    ":" +
+    (value.minute < 10 ? "0" + value.minute : value.minute) +
+    ":" +
+    (value.second < 10 ? "0" + value.second : value.second) +
+    (value.millisecond ? "." + millisecondString : "") +
+    (timezoneString ? timezoneString : "")
+  );
+}
+
+/**
+ * @internal
+ * @param literalString Value to deserialise.
+ * @returns Deserialized time, or null if the given value is not a valid serialised datetime.
+ * @see https://www.w3.org/TR/xmlschema-2/#time-lexical-repr
+ */
+export function deserializeTime(literalString: string): Time | null {
+  // Time in the format described at
+  // https://www.w3.org/TR/xmlschema-2/#time-lexical-repr
+  // \d\d:\d\d:\d\d - Two digits for the hour, minute and second, respectively, separated by a `:`.
+  //                  Example: "13:37:42".
+  // (\.\d+)? - Optionally a `.` followed by one or more digits representing milliseconds.
+  //            Example: ".1337".
+  // (Z|(\+|-)\d\d:\d\d) - The letter Z indicating UTC, or a `+` or `-` followed by two digits for
+  //                       the hour offset and two for the minute offset, separated by a `:`.
+  //                       Example: "+13:37".
+  const timeRegEx = /\d\d:\d\d:\d\d(\.\d+)?(Z|(\+|-)\d\d:\d\d)?/;
+  if (!timeRegEx.test(literalString)) {
+    return null;
+  }
+  const [timeString, timezoneString] = splitTimeFromTimezone(literalString);
+  const [hourString, minuteString, timeRest] = timeString.split(":");
+  let utcHours = Number.parseInt(hourString, 10);
+  let utcMinutes = Number.parseInt(minuteString, 10);
+  const [secondString, optionalMillisecondString] = timeRest.split(".");
+  const utcSeconds = Number.parseInt(secondString, 10);
+  const utcMilliseconds = optionalMillisecondString
+    ? Number.parseInt(optionalMillisecondString, 10)
+    : undefined;
+
+  if (utcMinutes >= 60) {
+    utcHours = utcHours + 1;
+    utcMinutes = utcMinutes - 60;
+  }
+
+  const deserializedTime: Time = {
+    hour: utcHours,
+    minute: utcMinutes,
+    second: utcSeconds,
+  };
+  if (typeof utcMilliseconds === "number") {
+    deserializedTime.millisecond = utcMilliseconds;
+  }
+  if (typeof timezoneString === "string") {
+    const [hourOffset, minuteOffset] = getTimezoneOffsets(timezoneString);
+    if (
+      typeof hourOffset !== "number" ||
+      hourOffset > 24 ||
+      typeof minuteOffset !== "number" ||
+      minuteOffset > 59
+    ) {
+      return null;
+    }
+    deserializedTime.timezoneHourOffset = hourOffset;
+    deserializedTime.timezoneMinuteOffset = minuteOffset;
+  }
+  return deserializedTime;
 }
 
 /**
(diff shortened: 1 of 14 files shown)


src/datatypes.ts
```diff
@@ -24,6 +24,7 @@ import { DataFactory } from "./rdfjs.internal";
 import { IriString, Iri, SolidClientError, LocalNode } from "./interfaces";
 import { internal_toIriString } from "./interfaces.internal";
 import { getLocalNodeName, isLocalNodeIri } from "./rdf.internal";
+import { literal } from "@rdfjs/dataset";
```
Vinnl, line 27:
```suggestion
```

src/datatypes.ts
```diff
@@ -153,6 +156,110 @@ export function deserializeDatetime(literalString: string): Date | null {
...
+  millisecond?: number;
+  timezoneHourOffset?: number;
+  timezoneMinuteOffset?: number;
+};
+
+/**
+ * @internal
+ * @param value Value to serialise.
+ * @returns String representation of `value` in UTC.
+ * @see https://www.w3.org/TR/xmlschema-2/#time-lexical-representation
```
Vinnl, line 175: The URL appears to be different for some reason:

```suggestion
 * @see https://www.w3.org/TR/xmlschema-2/#time-lexical-repr
```

src/datatypes.ts
```diff
@@ -153,6 +156,110 @@ export function deserializeDatetime(literalString: string): Date | null {
...
+ */
+export function serializeTime(value: Time): string {
+  const utcHours =
+    value.hour + (value.timezoneHourOffset ? value.timezoneHourOffset : 0);
+  const utcMinutes =
+    value.minute +
+    (value.timezoneMinuteOffset ? value.timezoneMinuteOffset : 0);
```
Vinnl, line 182: Hmm, by normalising everything to UTC, we're throwing away information. For example, if I set my alarm to 8:00, I don't want that to suddenly become 7:00 when Daylight Saving Time starts (or ends, whichever is appropriate :P ). So I think rather than adding the offsets to the hour and minute values, it's probably best to explicitly set the offset to the given values, or `Z` if none are passed.

src/datatypes.ts
```diff
@@ -153,6 +156,110 @@ export function deserializeDatetime(literalString: string): Date | null {
...
+  // get millisecond into three digits
+  if (value.millisecond) {
+    if (value.millisecond < 10) {
+      millisecondString = "00" + value.millisecond;
+    } else if (value.millisecond < 100) {
+      millisecondString = "0" + value.millisecond;
+    } else {
+      millisecondString = value.millisecond;
+    }
+  }
```
Vinnl, line 194: Looking at [the docs](https://www.w3.org/TR/xmlschema-2/#dateTime-lexical-representation), as I understand it the milliseconds don't actually need to be of a particular length, as opposed to the hour, minute and second representations.

src/datatypes.ts
```diff
@@ -153,6 +156,110 @@ export function deserializeDatetime(literalString: string): Date | null {
...
+
+/**
+ * @internal
+ * @param literalString Value to deserialise.
+ * @returns Deserialized time, or null if the given value is not a valid serialised datetime.
+ * @see https://www.w3.org/TR/xmlschema-2/#time-lexical-representation
+ */
+export function deserializeTime(literalString: string): Time | null {
+  // Time in the format described at
+  // https://www.w3.org/TR/xmlschema-2/#time-lexical-representation
```
Vinnl, line 214:
```suggestion
  // https://www.w3.org/TR/xmlschema-2/#time-lexical-repr
```

src/datatypes.ts
```diff
@@ -153,6 +156,110 @@ export function deserializeDatetime(literalString: string): Date | null {
...
+  // (without constraints on the value).
```
Vinnl, line 215: Hmm, so this is a leftover from the dateTime equivalent, which seems to accept e.g. `25:61` for the time part just fine. :shrug: 

```suggestion
```

src/datatypes.ts
```diff
@@ -153,6 +156,110 @@ export function deserializeDatetime(literalString: string): Date | null {
...
+    return null;
+  }
+  const [timeString, timezoneString] = splitTimeFromTimezone(literalString);
+  const [hourOffset, minuteOffset] = getTimezoneOffsets(timezoneString);
+  if (minuteOffset > 59) {
+    return null;
+  }
+  const [hourString, minuteString, timeRest] = timeString.split(":");
+  let utcHours = Number.parseInt(hourString, 10) + hourOffset;
+  let utcMinutes = Number.parseInt(minuteString, 10) + minuteOffset;
```
Vinnl, line 234: Here too I think it's probably better to return the original hour and minute values, and the timezone offsets separately, to avoid losing data.

src/thing/get.test.ts
```diff
@@ -871,6 +885,284 @@ describe("getDatetimeAll", () => {
...
+      "13:37:42Z",
+      "time"
+    );
+    const expectedTime = {
+      hour: 13,
+      minute: 37,
+      second: 42,
+    };
+
+    expect(getTime(thingWithTime, "https://some.vocab/predicate")).toEqual(
```
Vinnl, line 901: This holds for the other tests as well, but if you use `toStrictEqual` it will also fail the test if there are additional properties that we did not expect. (You might not see this everywhere in the codebase yet since it's a lesson we learned after many tests that still use regular `.toEqual` were already written.)

```suggestion
    expect(getTime(thingWithTime, "https://some.vocab/predicate")).toStrictEqual(
```

src/thing/remove.test.ts
```diff
@@ -2290,15 +2603,38 @@ describe("removeLiteral", () => {
...
+      )
+    );
+
+    expect(
+      updatedThing.predicates["https://some.vocab/predicate"].literals
+    ).toStrictEqual({
+      "http://www.w3.org/2001/XMLSchema#time": [],
+    });
+  });
+
```
Vinnl, line 2628: I think this is in the wrong `describe` block?

src/thing/remove.test.ts
```diff
@@ -2290,15 +2603,38 @@ describe("removeLiteral", () => {
...
-    const thingWithDatetime = getMockThingWithLiteralFor(
+    const thingWithDateTime = getMockThingWithLiteralFor(
       "https://some.vocab/predicate",
       "1990-11-12T13:37:42Z",
       "dateTime"
     );
 
     const updatedThing = removeLiteral(
-      thingWithDatetime,
+      thingWithDateTime,
```
Vinnl, line 2637: These changes seem unintended? We're spelling it `Datetime` everywhere else if I recall correctly, so for consistency's sake probably better to revert this.

src/thing/remove.ts
```diff
@@ -169,6 +171,37 @@ export const removeDatetime: RemoveOfType<Date> = (thing, property, value) => {
...
+    thing,
+    property,
+    xmlSchemaTypes.time,
+    function (foundTime) {
+      const deserializedTime = deserializeTime(foundTime);
+      if (deserializedTime) {
+        return (
+          deserializedTime.hour === value.hour &&
+          deserializedTime.minute === value.minute &&
+          deserializedTime.second === value.second
```
Vinnl, line 196: Shouldn't we also check the milliseconds and time zone offsets?

Vinnl (changes requested):
Overall looks good, but I think it's important to preserve the timezone data, if given.

Btw, you checked the "The changelog has been updated" checkbox, but this feature isn't listed there yet - same goes for the `Date` support. Could you add that too?

src/datatypes.ts
dgbarclay: Okay, will do both. I only chose to do it this way because we store datetime and date as UTC using this method. If you think that this isn't the best solution for time then I'll get it changed.

src/thing/remove.test.ts
dgbarclay: yep, that was unintended...

src/thing/remove.ts
dgbarclay: I omitted them because they're not always guaranteed, may make sense to have the check for those which do tho

src/datatypes.ts
dgbarclay: This was added so that if a user were to create a time as such:
time = {
...
millisecond: 42
}
It would be displayed as 0.42 instead of correctly storing it as 0.042. We need some kind of correction mechanism, but if there's no fixed length, I'll just use a division by 1000.

dgbarclay: Possibly just a division will work

src/datatypes.ts
dgbarclay: Yep, fair point. I was trying to match the standards we have for Date and DateTime.

src/thing/remove.ts
dgbarclay: > Overall looks good, but I think it's important to preserve the timezone data, if given.
> 
> Btw, you checked the "The changelog has been updated" checkbox, but this feature isn't listed there yet - same goes for the `Date` support. Could you add that too?

Where can I find the changelog??

src/datatypes.ts
dgbarclay: Aren't milliseconds inherently limited to 999? If we want there to be no fixed length then maybe we need a better naming convention or an option for more specificity.

src/datatypes.ts
Vinnl: Yeah that makes sense, unfortunately in the case of `Date` we're forced to throw away that info because Date doesn't allow us to represent timezones, but since we have our own data structure here we can avoid that :)

src/datatypes.ts
Vinnl: Hmm, so I was thinking that `0.42` was the correct value, but I now think I might indeed be misinterpreting it:

>  All ·minimally conforming· processors ·must· support (...) a minimum fractional second precision of milliseconds or three decimal digits (i.e. s.sss)

High school maths is a while ago, but it might indeed be the case that you need those leading zeroes to preserve the precision? Either way I don't think there's real harm in prepending them, so let's leave them in.

There's not necessarily a fixed length, it seems, but it's OK to have one:

> ·minimally conforming· processors ·may· set an application-defined limit on the maximum number of digits they are prepared to support in these two cases, in which case that application-defined maximum number ·must· be clearly documented.

src/datatypes.ts
Vinnl: Yeah as mentioned elsewhere, when using `Date` the developer can't explicitly choose a time zone, so there's no information we're throwing away there. Once [Temporal](https://github.com/tc39/proposal-temporal) makes it to JavaScript and we add support, we'll have to preserve the timezone too for Date and Datetime when passed it.

src/thing/remove.ts
Vinnl: In the root of the repository there is a file `CHANGELOG.md`. You can look at the existing entries for the format, it's roughly in line with https://keepachangelog.com/.

src/thing/remove.test.ts
dgbarclay: I believe this is needed. We do the same check for DateTime, seeing if removeLiteral accepts that data type, so I think it makes sense to do this for time too??

Vinnl: Ah sorry! I thought this was in the tests for Datetime, but it's actually in the block for `removeLiteral`. Never mind then!

src/datatypes.ts
```diff
@@ -193,71 +187,120 @@ export function serializeTime(value: Time): string {
     }
   }
 
+  if (value.timezoneHourOffset || value.timezoneHourOffset == 0) {
```
Vinnl, line 190: If I'm understanding this correctly you basically want to check whether `value.timezoneHourOffset` is defined, but since 0 is implicitly cast to `false`, you also have to check for the value 0 (or rather, you're checking for values that can be cast to 0 by using `==` rather than `===`, so `false == 0` also matches).

To avoid potential footguns like that, when working in JavaScript it's generally a good habit to avoid implicit type casting:

```suggestion
  if (typeof value.timezoneHourOffset !== "undefined") {
```

or even checking explicitly whether it is of the correct type:

```suggestion
  if (typeof value.timezoneHourOffset === "number") {
```

src/datatypes.ts
```diff
@@ -193,71 +187,120 @@ export function serializeTime(value: Time): string {
...
+  // https://www.w3.org/TR/xmlschema-2/#time-lexical-repr
   // \d\d:\d\d:\d\d - Two digits for the hour, minute and second, respectively, separated by a `:`.
   //                  Example: "13:37:42".
   // (\.\d+)? - Optionally a `.` followed by one or more digits representing milliseconds.
   //            Example: ".1337".
   // (Z|(\+|-)\d\d:\d\d) - The letter Z indicating UTC, or a `+` or `-` followed by two digits for
   //                       the hour offset and two for the minute offset, separated by a `:`.
   //                       Example: "+13:37".
-  const datetimeRegEx = /\d\d:\d\d:\d\d(\.\d+)?(Z|(\+|-)\d\d:\d\d)/;
+  const datetimeRegEx = /\d\d:\d\d:\d\d(\.\d+)?(Z|(\+|-)\d\d:\d\d)?/;
```
Vinnl, line 240: Heh... Super minor issue, but I guess this is a `timeRegEx` only:

```suggestion
  const timeRegEx = /\d\d:\d\d:\d\d(\.\d+)?(Z|(\+|-)\d\d:\d\d)?/;
```

src/datatypes.ts
```diff
@@ -193,71 +187,120 @@ export function serializeTime(value: Time): string {
...
     ? Number.parseInt(optionalMillisecondString, 10)
     : undefined;
 
-  if (utcMinutes < 0) {
-    utcHours = utcHours - 1;
-    utcMinutes = 60 + utcMinutes;
-  } else if (utcMinutes >= 60) {
+  if (utcMinutes >= 60) {
     utcHours = utcHours + 1;
     utcMinutes = utcMinutes - 60;
```
Vinnl: Edge case, but what if `utcMinutes > 120`?

```suggestion
    utcHours = utcHours + Math.floor(utcMinutes / 60);
    utcMinutes = utcMinutes % 60;
```

Vinnl (approved):
Good job :+1: Since I'm more familiar with the changes in `main` I'll resolve the merge conflicts, fix up the teeny tiny remarks I still had and merge it in.

src/datatypes.ts
Vinnl: Ah right, that can't happen because there's just two characters for the minutes :)

Vinnl:
Heh, turns out the merge conflicts were with your earlier Date data type changes :P

PR #1122, opened 2021-06-29
Only prepare end-to-end tests when running them

Vinnl:
This prepare job fails relatively often when npm is trying to
install things and fails fetching the relevant metadata. Since
we're not actually running the browser-based end-to-end tests all
the time, we should be able to cut down on those failures by only
running the preparation step when we're actually going to run
browser-based end-to-end tests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,6 +67,7 @@ jobs:
         # Run the end-to-end tests against the code in this branch specifically:
         npm install ../../
         cd ../..
+      if: matrix.node-version != '14.x' && matrix.node-version != '12.x' && github.actor != 'dependabot[bot]'
     - name: Run browser-based end-to-end tests (Linux)
       # TODO: Add Edge/merge with Windows setup once Edge is available on Linux:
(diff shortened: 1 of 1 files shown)


ajacksified (approved):

.github/workflows/ci.yml
```diff
@@ -67,6 +67,7 @@ jobs:
         # Run the end-to-end tests against the code in this branch specifically:
         npm install ../../
         cd ../..
+      if: matrix.node-version != '14.x' && matrix.node-version != '12.x' && github.actor != 'dependabot[bot]'
```
pmcb55, line 70: Just a minor nit, but rather than these explicit references to version numbers, can we define a new 'node-version-latest' matrix param:
```yaml
strategy:
      matrix:
        # Available OS's: https://help.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners
        os: [ubuntu-20.04, windows-2019, macos-10.15]
        node-version-latest: 16.x
        node-version: [node-version-latest, 14.x, 12.x]
```
...and then this line becomes:
```yaml
if: matrix.node-version = matrix.node-version-latest && github.actor != 'dependabot[bot]'
```

PR #1173, opened 2021-07-28
Bail out when chain detection is expensive + edge case fix

### LINKED ISSUE #1158: getSolidDataset is blocking
### Search terms you've used

- skimmed through all issues
- solidDataset, getSolidDataset, blocking, synchronous

#### Overview
When executing `getSolidDataset` with a particularly large `turtle` file, a javascript app freezes for a while
The `async` nature of `getSolidDataset` would suggest that the code is non-blocking, but the actual behaviour suggests otherwise.

#### Example

An app would freeze for 71 seconds when running `getSolidDataset('https://ruben.verborgh.org/profile/#me')`

#### Detail
On a closer look, it seems that the blocking code is the following:

Blocks for 60 seconds:
https://github.com/inrupt/solid-client-js/blob/main/src/rdfjs.internal.ts#L317-L319

Blocks for 11 seconds:
https://github.com/inrupt/solid-client-js/blob/main/src/rdfjs.internal.ts#L326-L347

Maybe these pieces of code could be more efficient? Maybe they run with high time complexity? (~~Didn't study it in detail~~ _seems like a long-running cycle detecting algorithm_)

The time is just an example. Of course it will take different times in different context.

### To Reproduce

1. `import { getSolidDataset } from '@inrupt/solid-client'`
2. run `await getSolidDataset('https://ruben.verborgh.org/profile/#me')`
3. see that things freeze for a long time

**Minimal reproduction**

_Used a custom code sandbox, not the one provided here._

code sandbox: https://codesandbox.io/s/gracious-buck-7xyvj?file=/src/index.ts
resulting app: https://7xyvj.csb.app/

### Expected result

When `await getSolidDataset('https://ruben.verborgh.org/profile/#me')` is executed within an app, the app keeps running.
The `getSolidDataset` either runs faster, or runs asynchronously or both.

### Actual result

The app freezes for a long time.

### Environment

```
```

```
 System:
    OS: Linux 5.12 Arch Linux
    CPU: (4) x64 Intel(R) Core(TM) i5-3320M CPU @ 2.60GHz
    Memory: 229.06 MB / 5.62 GB
    Container: Yes
    Shell: 😛 not relevant
  Binaries:
    Node: 16.2.0 - /usr/bin/node
    Yarn: 1.22.10 - /usr/bin/yarn
    npm: 7.15.0 - /usr/bin/npm
  Browsers:
    😛 not relevant
  npmPackages:
    @inrupt/solid-client: ^1.10.0 => 1.10.0 
    node-fetch: ^2.6.1 => 2.6.1 
  npmGlobalPackages:
    😛 not relevant
```

### Additional

If you want to see a production app running into this issue, you can open a [solid friends crawler](https://friends.livegraph.org), stay logged out and watch for a while, until the app reaches the profile `https://ruben.verborgh.org/profile/#me` and freezes

Vinnl:
This PR fixes #1158. Working on that also helped me discover that there was a (fairly common) case in which we did not correctly detect non-cyclic chains and thus generated non-deterministic identifiers for blank nodes in cases where that was not necessary - this is fixed in the first commit c0960e252a90c26b969e97018a23833e68bf258c.

The second commit (61d0c89a16a39be0e54af07b6874fc2d4331a2d7) bails out of chain detection when a Resource contains more than 20 Blank Nodes. Chain detection is really an optimisation: when a Resource contains non-cyclic Blank Nodes (e.g. when using RDF lists, or mashlib's default representation of trusted apps), it ensures that the resulting SolidDataset will always have the exact same properties whenever it's parsed. This helps e.g. tools like SWR use something like [dequal](https://github.com/lukeed/dequal) to verify that the same Resource fetched twice was unchanged. However, there will always be cases when we cannot do this, e.g. when there's a cycle of Blank Nodes. Therefore, it's not a problem to simply skip that work when it would take a lot of time.

- [x] The changelog has been updated, if applicable. (Didn't think it was particularly interesting, but can add a blurb if someone disagrees.)

### DIFF
--- a/src/resource/solidDataset.ts
+++ b/src/resource/solidDataset.ts
@@ -233,11 +233,19 @@ export async function responseToSolidDataset(
           solidDataset = addRdfJsQuadToDataset(solidDataset, quad);
         }
       });
-      parser.onComplete(() => {
+      parser.onComplete(async () => {
+        // If a Resource contains more than this number of Blank Nodes,
+        // we consider the detection of chains (O(n^2), I think) to be too
+        // expensive, and just incorporate them as regular Blank Nodes with
+        // non-deterministic, ad-hoc identifiers into the SolidDataset:
+        const maxBlankNodesToDetectChainsFor = 20;
         // Some Blank Nodes only serve to use a set of Quads as the Object for a
         // single Subject. Those Quads will be added to the SolidDataset when
         // their Subject's Blank Node is encountered in the Object position.
-        const chainBlankNodes = getChainBlankNodes(quadsWithBlankNodes);
+        const chainBlankNodes =
+          quadsWithBlankNodes.length <= maxBlankNodesToDetectChainsFor
+            ? getChainBlankNodes(quadsWithBlankNodes)
+            : [];
         const quadsWithoutChainBlankNodeSubjects = quadsWithBlankNodes.filter(
           (quad) =>
             chainBlankNodes.every(
--- a/src/rdf.test.ts
+++ b/src/rdf.test.ts
@@ -587,7 +587,7 @@ describe("fromRdfJsDataset", () => {
       );
     });
 
-    it("can parse chained Blank Nodes with a single lnk that end in a dangling Blank Node", () => {
+    it("can parse chained Blank Nodes with a single link that end in a dangling Blank Node", () => {
       const mockDataset: ImmutableDataset = {
         type: "Dataset",
         graphs: { default: {} },
--- a/src/rdfjs.internal.ts
+++ b/src/rdfjs.internal.ts
@@ -502,9 +502,11 @@ function getCycleBlankNodes(
     )
     .map((quad) => quad.object as RdfJs.BlankNode);
 
-  // If no Blank Nodes are connected to `currentNode`, we're done:
+  // If no Blank Nodes are connected to `currentNode`, and `currentNode` is not
+  // part of a cycle, we're done; the currently traversed Nodes do not form a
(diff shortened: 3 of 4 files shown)


src/resource/solidDataset.ts
```diff
@@ -233,11 +233,19 @@ export async function responseToSolidDataset(
           solidDataset = addRdfJsQuadToDataset(solidDataset, quad);
         }
       });
-      parser.onComplete(() => {
+      parser.onComplete(async () => {
+        // If a Resource contains more than this number of Blank Nodes,
+        // we consider the detection of chains (O(n^2), I think) to be too
+        // expensive, and just incorporate them as regular Blank Nodes with
+        // non-deterministic, ad-hoc identifiers into the SolidDataset:
+        const maxBlankNodesToDetectChainsFor = 20;
```
NSeydoux, line 241: Should this be a constant in the `constants.ts` file ? So far it's only RDF constants, but it feels weird to have a hard-coded limit defined so locally.

NSeydoux (approved):

src/resource/solidDataset.ts
Vinnl: Not sure, since it's really only used locally, i.e. there's just a single mention of this variable - I mostly extracted it to be able to give it a name that explained its purpose, rather than for it to be a magic number. I'm afraid that pulling it out of its context will make it _harder_ to understand what it does, and it's not meant to be re-used elsewhere in the codebase either.

PR #1177, opened 2021-07-29
Account for Link with rel="acl" pointing to ACRs

Vinnl:
# New feature description

The ACP proposal is going to be updated such that Link headers with
rel="acl" can also point to Access Control Resources, rather than
being used by WAC exclusively in order to point to ACLs.

I'm still not entirely sure that we've covered all potential
problems this may cause with this change, but I expect it to be
enough.

Ideally, WAC would also mandate a rel="type" Link header for Access
Control Lists, so we could not just check that it is _not_ of type
ACR, but explicitly check that it _is_ of type ACL.

@matthieubosquet I reworked our initial approach such that `fetchAcl` no longer throws an error, as I realised that that would be breaking for people who (justifiably) assumed that they should be able to call that after calling `hasAccessibleAcl`. So I've now created an internal-only specific error class for that, and use that to detect the case of the ACL being an ACR and then returning `null` instead to indicate that the Resource and fallback ACL could not be fetched.

# Checklist

### DIFF
--- a/src/acl/acl.test.ts
+++ b/src/acl/acl.test.ts
@@ -163,6 +163,50 @@ describe("fetchAcl", () => {
     );
   });
 
+  it("returns null for both ACLs if the Resource points to an ACR instead", async () => {
+    const mockFetch = jest.fn((url) => {
+      if (url === "https://some.pod/resource.acl") {
+        return Promise.resolve(
+          mockResponse(undefined, {
+            url: "https://some.pod/resource.acl",
+            headers: {
+              Link: '<http://www.w3.org/ns/solid/acp#AccessControlResource>; rel="type"',
+              "Content-Type": "text/turtle",
+            },
+          })
+        );
+      }
+
+      return Promise.resolve(
+        mockResponse(undefined, {
+          headers: {
+            "Content-Type": "text/turtle",
+            Link: '<resource.acl>; rel="acl"',
+          },
+          url: url as string,
+        })
+      );
+    });
+
+    const mockResourceInfo: WithServerResourceInfo = {
+      internal_resourceInfo: {
+        sourceIri: "https://some.pod/resource",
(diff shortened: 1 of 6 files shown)


src/acl/acl.test.ts
```diff
@@ -310,6 +354,66 @@ describe("fetchResourceAcl", () => {
...
+    await expect(
+      internal_fetchResourceAcl(sourceDataset, {
+        fetch: mockFetch,
+      })
+    ).rejects.toThrow(
+      "[https://some.pod/resource] is governed by Access Control Policies in [https://some.pod/resource?ext=acr] rather than by Web Access Control."
+    );
+  });
+
+  it("throws an error if the linked Resource is an ACP ACR and thus has another type", async () => {
```
NSeydoux, line 388: I'm not sure about the phrasing of this test description. If I get it right, the intent is to test that `internal_fetchResourceAcl` throws if dereferencing the target resource return several Link headers, one of them indicating that it is an ACR. Is that correct ? If so, what happens if one of the Link headers describes the resource as governed by an ACL as well (in other words, if the Link headers are contradictory) ?

I mean, the current implementation throws, and that may be the behaviour that we want, but it's not entirely explicit.

NSeydoux (approved):

src/acl/acl.test.ts
Vinnl: Hmm, that's a good question... The problem is that there is no Link header that describes a Resource as an ACL (which is why we only check if it's _not_ an ACR). But yes, if the Link headers are contradictory, I think we'd still want to throw, since if something so idiosyncratic is the case, we cannot reliably assume that it's an ACL. I've updated the test description to no longer conclude that it's definitely not an ACL if it's advertised as an ACR: dc3f9ffc

PR #1187, opened 2021-08-03
Prepare for the use of WAC Modes in ACP

Vinnl:
This renames the existing APIs that set Allow and Deny Modes in ACP
to append *V1, and creates new *V2 versions that use the WAC
vocabulary. It also prepares an acp_v4 export that will carry all
the updated APIs for the upcoming ACP changes. (This isn't public yet though.)

To avoid this breaking for developers, the acp_v1 through acp_v3
exports still carry the *V1 functions under their old names. 
Additionally, to avoid breaking submodule imports, the old
names are also aliased inside the acp/policy module to point to
the *V1 versions. However, those will have to eventually be
updated to point to the *V2 functions, which will technically be
breaking. However, given how the ecosystem hasn't yet widely
moved to submodule imports, how I haven't heard of anybody using
those yet, and how we've explicitly warned that the ACP APIs can
break in non-major versions, I don't expect that to harm anyone
in practice.

Once servers move to implement the updated vocabs, apps will of
course still break. If all relevant servers (in practice, just
pod.inrupt.com, as that's the only one that implements ACPs)
update at the same time, then developers can switch around to the
new APIs at that point in time.

/cc @matthieubosquet

# Checklist


src/acp/policy.ts
```diff
@@ -177,6 +178,48 @@ export function setPolicy<Dataset extends SolidDataset>(
...
+  policy: P,
+  modes: AccessModes
+): P {
+  let newPolicy = removeAll(policy, acp.allow);
+
+  if (modes.read === true) {
+    newPolicy = addIri(
+      newPolicy,
+      acp.allow,
+      internal_accessModeIriStrings.read
```
NSeydoux, line 203: Just to make sure I get this straight, the access modes changed to be compatible with ACL, but the rest of the ACP vocabulary/spec is unchanged ?

NSeydoux (approved):

src/acp/policy.ts
Vinnl: Yep! Had to check a couple of times during implementation myself as well.

PR #1193, opened 2021-08-05
Parcel fixes

Vinnl:
These don't quite fix  the end-to-end tests hanging in CI, but they do get the MacOS runs to succeed again - before, they'd crash due to error messages thrown onto the page by Parcel.

### DIFF
--- a/.codesandbox/sandbox/end-to-end-test.html
+++ b/.codesandbox/sandbox/end-to-end-test.html
@@ -10,7 +10,7 @@
 	<input type="url" name="idp" id="idp" value="https://broker.pod.inrupt.com"/>
 	<button>Log in</button>
 
-	<script src="src/end-to-end-test.ts">
+	<script src="src/end-to-end-test.ts" type="module">
 	</script>
 </body>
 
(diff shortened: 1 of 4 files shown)


.codesandbox/sandbox/end-to-end-test.html
```diff
@@ -10,7 +10,7 @@
 	<input type="url" name="idp" id="idp" value="https://broker.pod.inrupt.com"/>
 	<button>Log in</button>
 
-	<script src="src/end-to-end-test.ts">
+	<script src="src/end-to-end-test.ts" type="module">
```
Vinnl, line 13: This is apparently needed now to tell Parcel to correctly parse `import` statements.

.codesandbox/sandbox/package.json
```diff
@@ -2,7 +2,6 @@
   "name": "solid-client-sandbox",
   "version": "1.0.0",
   "description": "Example project calling out to solid-client",
-  "main": "index.html",
```
Vinnl, line 5: And for this it was complaining that it's not a JS file. And since it was more informative than actually useful, I just removed it :)

NSeydoux (approved):

Vinnl:
@NSeydoux I added a5494b9c4887e3f17403a5d405cfaf61c36f6f85 to rely on just the MacOS tests for now, could you take a look whether that makes sense? I'll add a ticket to investigate this in the future.

NSeydoux:
The change makes sense to me, even if it's still unclear what triggered the failure.

PR #1202, opened 2021-08-10
Make the Thing builder stateful

Vinnl:
# New feature description

Make the Thing builder stateful. This means that it is no longer necessary to chain calls after
`buildThing()`. This is useful for conditionally adding data, e.g.
adding a property with someone's last name only if provided.

In other words, you can now do:

    const bookmarkBuilder = buildThing()
      .addUrl("http://www.w3.org/1999/02/22-rdf-syntax-ns#type", "http://www.w3.org/2002/01/bookmark#Bookmark")
      .addUrl("http://www.w3.org/2002/01/bookmark#recalls", form.url);
    if (typeof form.title !== "undefined") {
      bookmarkBuilder.addStringNoLocale("http://purl.org/dc/terms/title", form.title);
    }
    const bookmark = bookmarkBuilder.build();

(Whereas before, the title would not be included.)

That makes it a bit different from our other, mostly-stateless, API's, but given that `buildThing` is mostly used in a local scope, I think it makes sense here. Feel free to reject though.

# Checklist


NSeydoux (approved):
Nice one, this type of features make code much easier to read :)

PR #1205, opened 2021-08-11
Try using Playwright for the browser-based end-to-end test

Vinnl:
This is an experiment in replacing TestCafe with Playwright. The advantages are:

- It's more stable (so far they're reliably succeeding, and enabled on Linux, Windows and MacOS - except for Safari on the latter, which fails for Playwright, but is flaky and requires a weird workaround for TestCafe as well). Main reason for this is probably because it controls the browsers (via the devtools protocols - also see [this article](https://troglodyne.net/posts/1617057517)) rather than via a proxy to the test app.
- It allows setting custom user agents for all browsers, making requests easier to find in server logs.
- For failed tests we don't just have screenshots, but actual screen recordings and test traces that can be inspected using the [Playwright Trace Viewer](https://playwright.dev/docs/trace-viewer/#viewing-the-trace), giving access to console output and all that as well.
- It uses Jest's assertion library, so that syntax is the same as in our unit tests.

The main disadvantage is that other projects still use TestCafe, so it's inconsistent with those.

# Checklist


src/e2e-browser/globalSetup.ts
```diff
@@ -19,29 +19,31 @@
  * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
-import { t, ClientFunction, Selector } from "testcafe";
-import { screen } from "@testing-library/testcafe";
+import { spawn } from "child_process";
```
Vinnl, line 22: `globalSetup.ts` has nothing to do with `gluu.ts` (which I didn't port since Gluu was removed). It basically starts the sandbox server.

NSeydoux:
Superseded by https://github.com/inrupt/solid-client-js/pull/1412

PR #1223, opened 2021-08-24
Get well known solid & add JSON-LD parser

andydavison:
# New feature description

- Adds new `getWellKnownSolid` function to return the contents (as a SolidDataset) of the '.well-known/solid' endpoint for a given resource URL
- Adds JSON-LD parser, and adds support for `application/ld+json` to `responseToSolidDataset` and `getSolidDataset`

# Checklist

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -57,5 +57,11 @@ export default {
       },
     }),
   ],
-  external: ["cross-fetch", "http-link-header", "@rdfjs/dataset", "n3"],
+  external: [
+    "cross-fetch",
+    "http-link-header",
+    "@rdfjs/dataset",
+    "n3",
+    "jsonld-streaming-parser",
+  ],
 };
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -62,6 +62,6 @@ export default {
     "http-link-header",
     "@rdfjs/dataset",
     "n3",
-    "jsonld-streaming-parser",
+    "jsonld",
   ],
 };
--- a/src/formats/jsonLd.ts
+++ b/src/formats/jsonLd.ts
@@ -1,0 +1,70 @@
+/**
+ * Copyright 2021 Inrupt Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to use,
+ * copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the
+ * Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,
+ * INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A
+ * PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+ * HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+ * OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+ * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+import { JsonLdParser } from "jsonld-streaming-parser";
+import { IriString } from "../interfaces";
+import { DataFactory } from "../rdfjs.internal";
+import { getSourceUrl } from "../resource/resource";
+import { Parser } from "../resource/solidDataset";
+
+export const getJsonLdParser = (): Parser => {
+  const onQuadCallbacks: Array<Parameters<Parser["onQuad"]>[0]> = [];
+  const onCompleteCallbacks: Array<Parameters<Parser["onComplete"]>[0]> = [];
+  const onErrorCallbacks: Array<Parameters<Parser["onError"]>[0]> = [];
+
+  return {
+    onQuad: (callback) => {
+      onQuadCallbacks.push(callback);
+    },
+    onError: (callback) => {
+      onErrorCallbacks.push(callback);
+    },
+    onComplete: (callback) => {
+      onCompleteCallbacks.push(callback);
+    },
+    parse: async (source, resourceInfo) => {
+      const parser = await getParser(getSourceUrl(resourceInfo));
+      const parserPromise = new Promise<void>((resolve) => {
+        parser.on("data", (quad) => {
+          onQuadCallbacks.every((callback) => callback(quad));
+        });
+        parser.on("error", (error) => {
+          onErrorCallbacks.forEach((callback) => callback(error));
+        });
+        parser.on("end", () => {
+          onCompleteCallbacks.every((callback) => callback());
+          resolve();
+        });
+        parser.write(source);
+        parser.end();
+      });
+      await parserPromise;
+    },
+  };
+};
+
+async function getParser(baseIri: IriString) {
+  return new JsonLdParser({
+    baseIRI: baseIri,
+    processingMode: "1.1",
+    dataFactory: DataFactory,
+  });
+}
--- a/src/formats/jsonLd.ts
+++ b/src/formats/jsonLd.ts
@@ -19,7 +19,7 @@
  * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
-import { toRDF } from "jsonld";
+import * as jsonld from "jsonld";
 import {
   Quad,
   Quad_Graph,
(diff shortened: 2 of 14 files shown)


src/resource/solidDataset.ts
```diff
@@ -1133,3 +1137,37 @@ function resolveLocalIrisInThing(
...
+  url: UrlString | Url,
+  options: Partial<
+    typeof internal_defaultFetchOptions & ParseOptions
+  > = internal_defaultFetchOptions
+): Promise<SolidDataset & WithServerResourceInfo> {
+  const urlString = internal_toIriString(url);
+  const resourceMetadata = await getResourceInfo(urlString, options);
+  const linkedResources = getLinkedResourceUrlAll(resourceMetadata);
+  const rootResources =
+    linkedResources["http://www.w3.org/ns/pim/space#storage"];
```
NSeydoux, line 1158: IRIs which are part of the spec are typically added to the `src/constants.ts` file. At some point, we should consider using the Artifact Generator for this :)

src/resource/solidDataset.ts
```diff
@@ -1133,3 +1137,37 @@ function resolveLocalIrisInThing(
...
+  const rootResource = rootResources?.length === 1 ? rootResources[0] : null;
+  if (!rootResource) {
+    throw new SolidClientError(
+      `Unable to determine root resource for Resource at [${url}].`
+    );
+  }
+  const wellKnownSolidUrl = `${rootResource}.well-known/solid`;
```
NSeydoux, line 1165: Even though this is valid because it only makes sense for the root resource to be a container, and its canonical IRI should always end with a trailing slash, would it be safer to use `new URL(".well-known/solid", rootResource)` instead of direct concatenation to make sure we get a valid IRI out of this in the case where the rootResource IRI is missing a trailing slash ?

src/resource/solidDataset.ts
```diff
@@ -1133,3 +1137,37 @@ function resolveLocalIrisInThing(
       : thing.url,
   });
 }
+
+/**
+ * Fetch the contents of '.well-known/solid' for a given resource URL.
```
NSeydoux, line 1142: Should we extend the documentation a little bit to mention briefly what the well-known document is ?

src/formats/jsonLd.ts
```diff
@@ -0,0 +1,70 @@
...
+      onErrorCallbacks.push(callback);
+    },
+    onComplete: (callback) => {
+      onCompleteCallbacks.push(callback);
+    },
+    parse: async (source, resourceInfo) => {
+      const parser = await getParser(getSourceUrl(resourceInfo));
+      const parserPromise = new Promise<void>((resolve) => {
+        parser.on("data", (quad) => {
+          onQuadCallbacks.every((callback) => callback(quad));
```
NSeydoux, line 47: Why is this `every`, and not `forEach` ?

NSeydoux (approved):

src/formats/jsonLd.ts
andydavison: …that's a good point actually - I built on/adapted the existing example in `src/formats/turtle.ts` which does the same thing, but I guess it should just be forEach everywhere (we're not doing anything with the `.every` return value)

src/resource/solidDataset.ts
andydavison: good point, done!

src/formats/jsonLd.ts
NSeydoux: Yeah, I did the same thing when reviewing and checked what the turtle parser did, and saw this was the reason, but I wasn't sure why it was like this :)

src/resource/solidDataset.ts
andydavison: Do we have a description of that somewhere already? I've had a little look and not turned anything up.

Happy to expand this a bit, but suspect I might miss some conceptual nuance.

andydavison:
On the topic of whether the json-ld parser is required in `getSolidDataset` - a couple of places imply it should be:
https://github.com/inrupt/solid-client-js/blob/150355ad58a0f47a7f885ca53180ea1b2c71922d/src/resource/resource.internal.ts#L34-L41
and
https://github.com/inrupt/solid-client-js/blob/150355ad58a0f47a7f885ca53180ea1b2c71922d/ARCHITECTURE.md#L70-L74

Happy to either a) update those comments/docs, and switch to passing the parser in manually in `getWellKnownSolid` or b) leave it included as currently? As noted in the issue on `jsonld-streaming-parser`, that library is pretty small, _if_ we can stick with that rather than `jsonld` (which is much larger)

Vinnl:
Those docs are simply a convoluted way of saying that the server understands the RDF data is stored at, otherwise people keep complaining about "well, but you can store RDFA or RDF-XML on a server as well", but the relevant point for that documentation is that solid-client can turn that data into a SolidDataset. And although servers can serialise that data into _both_ Turtle _and_ JSON-LD, only one of those is needed for a given request. Since a Turtle serialiser is already needed for PATCH requests, might as well use the same library for parsing. (It'd be justifiable to replace the Turtle parser by a JSON-LD parser too, but both is unnecessary inflation of bundle sizes.)

src/formats/jsonLd.ts
```diff
@@ -19,8 +19,14 @@
  * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
-import { JsonLdParser } from "jsonld-streaming-parser";
-import { IriString } from "../interfaces";
+import * as jsonld from "jsonld";
```
NSeydoux, line 22: Is there a reason to prefer this syntax over explicitly importing functions from the `jsonld` package ? I have no issue with the current line, I'm just curious :)

src/formats/jsonLd.ts
```diff
@@ -41,30 +47,50 @@ export const getJsonLdParser = (): Parser => {
...
-          onCompleteCallbacks.forEach((callback) => callback());
-          resolve();
-        });
-        parser.write(source);
-        parser.end();
+      let quads = [] as Quad[];
+      try {
+        const plainQuads = (await jsonld.toRDF(JSON.parse(source), {
+          base: getSourceUrl(resourceInfo),
+        })) as [];
```
NSeydoux, line 54: I didn't know this type annotation was possible. Is this equivalent to `any[]` ?

src/resource/solidDataset.test.ts
```diff
@@ -3481,6 +3547,7 @@ describe("getWellKnownSolid", () => {
     );
 
     const wellKnownSolidResponse = await getWellKnownSolid(
+      // "https://pod.inrupt.com/andydavison/public"
```
NSeydoux, line 3550: Is this a test leftover ?

rollup.config.js
```diff
@@ -57,5 +57,11 @@ export default {
       },
     }),
   ],
-  external: ["cross-fetch", "http-link-header", "@rdfjs/dataset", "n3"],
+  external: [
+    "cross-fetch",
+    "http-link-header",
+    "@rdfjs/dataset",
+    "n3",
+    "jsonld-streaming-parser",
```
NSeydoux, line 65:
```suggestion
    "jsonld",
```

src/resource/solidDataset.ts
```diff
@@ -1135,3 +1138,41 @@ function resolveLocalIrisInThing(
...
+  options: Partial<
+    typeof internal_defaultFetchOptions & ParseOptions
+  > = internal_defaultFetchOptions
+): Promise<SolidDataset & WithServerResourceInfo> {
+  const urlString = internal_toIriString(url);
+  const resourceMetadata = await getResourceInfo(urlString, options);
+  const linkedResources = getLinkedResourceUrlAll(resourceMetadata);
+  const rootResources = linkedResources[pim.storage];
+  const rootResource = rootResources?.length === 1 ? rootResources[0] : null;
+  if (!rootResource) {
```
NSeydoux, line 1163:
```suggestion
  if (rootResource === null) {
```
isn't testing for the value we expect more specific that testing for nullish ?

src/resource/solidDataset.ts
```diff
@@ -1135,3 +1138,41 @@ function resolveLocalIrisInThing(
...
+  const wellKnownSolidUrl = new URL(".well-known/solid", rootResource).href;
+
+  const wellKnownSolidDataset = await getSolidDataset(wellKnownSolidUrl, {
+    ...options,
+    parsers: {
+      "application/ld+json": getJsonLdParser(),
+    },
+  });
+
+  return wellKnownSolidDataset;
```
NSeydoux, line 1177:
```suggestion
  return getSolidDataset(wellKnownSolidUrl, {
    ...options,
    parsers: {
      "application/ld+json": getJsonLdParser(),
    },
  });
```

NSeydoux (approved):
Looks good to me ! Ruben mentioned that he had bundle size issues with `jsonld` in the past. Do you think we should add a ticket to follow up on this, and potentially migrate back to `jsonldstreaming-parser` if the dependency on NodeJS buitins is resolved ?

src/resource/solidDataset.test.ts
andydavison: 🤦 yep - removed!

src/formats/jsonLd.ts
andydavison: actually, this probably should be `any[]` (ts isn't complaining at me, but for clarity as much as anything)

src/formats/jsonLd.ts
andydavison: I was playing around when looking into some mocking that wasn't working. But, switched back now to the specific named import and all ok, so have reverted to that

andydavison:
> Looks good to me ! Ruben mentioned that he had bundle size issues with `jsonld` in the past. Do you think we should add a ticket to follow up on this, and potentially migrate back to `jsonldstreaming-parser` if the dependency on NodeJS buitins is resolved ?

@NSeydoux I think this makes sense (especially as all we're using is the parser) - I'll make a ticket

PR #1224, opened 2021-08-24
Add docs on getSolidDataset 401 intercept and retry when no trailing slash for container

### LINKED ISSUE #1216: 401 logged to console despite successful request to solidcommunity.net server
### Search terms you've used

401

### Bug description

I can successfully use `getSolidDataset` to get a directory but it also logs a 401 error to the console.

### To Reproduce

1. Call `getSolidDataset`

**Minimal reproduction**

https://github.com/AJamesPhillips/solid-bugs/tree/1567be3

(I decided not to use [codesandbox](https://codesandbox.io/s/github/inrupt/solid-client-js/tree/main/.codesandbox/sandbox) as there were a lot of error and warning messages to the console).

### Expected result

No error log out.

### Actual result

Solid_Bug

### Environment

```

  Browsers:
    Brave Browser: 92.1.27.109
    Chrome: 92.0.4515.159
  npmPackages:
    @inrupt/solid-client: ^1.1.0 => 1.11.0 
    @inrupt/solid-client-authn-browser: ^1.11.0 => 1.11.0 
    @inrupt/vocab-common-rdf: ^0.6.3 => 0.6.3 
    buffer: ^6.0.2 => 6.0.3 
    clean-webpack-plugin: ^3.0.0 => 3.0.0 
    html-webpack-plugin: ^4.5.0 => 4.5.2 
    lodash: ^4.17.20 => 4.17.21 
    rdf-namespaces: ^1.9.2 => 1.9.2 
    solid-auth-client: ^2.5.3 => 2.5.6 
    tripledoc: ^4.4.0 => 4.4.0 
    webpack: ^5.5.0 => 5.51.1 
    webpack-cli: ^4.2.0 => 4.8.0 
    webpack-dev-server: ^3.11.0 => 3.11.2 
  npmGlobalPackages:
    npm-run: 5.0.1
    npm: 6.14.8
    pnpm: 6.3.0
```

AJamesPhillips:
This PR fixes #1216

### DIFF
--- a/src/resource/solidDataset.ts
+++ b/src/resource/solidDataset.ts
@@ -278,6 +278,8 @@ export async function responseToSolidDataset(
 /**
  * Fetch a SolidDataset from the given URL. Currently requires the SolidDataset to be available as [Turtle](https://www.w3.org/TR/turtle/).
  *
+ * If operating on a container, the URL must end with a trailing slash "/" otherwise a 401 error will be logged ([more information](https://github.com/inrupt/solid-client-js/issues/1216#issuecomment-904703695)).
+ *
  * @param url URL to fetch a [[SolidDataset]] from.
  * @param options Optional parameter `options.fetch`: An alternative `fetch` function to make the HTTP request, compatible with the browser-native [fetch API](https://developer.mozilla.org/docs/Web/API/WindowOrWorkerGlobalScope/fetch#parameters).
  * @returns Promise resolving to a [[SolidDataset]] containing the data at the given Resource, or rejecting if fetching it failed.


NSeydoux:
Although your addition makes sense, it only applies if the user is logged in (if not logged in, no DPoP token is used so the redirections aren't problematic), and using `solid-client-authn-browser` (the same would apply to `solid-client-authn-node`, but the absence of a browser console in node makes the issue invisible), as other libraries may behave differently. I'll modify the comment a bit with this additional info, and as for the last time I'll recreate your PR, since the GH actions still refuse to run on external contributions.

NSeydoux:
On second thought, I'm wondering if this isn't out-of-scope for the TSDoc, or rather if there could be other cases where this could happen in other places too. I'll look for these, and consider if this should be in the TSDoc or if we could have an FAQ with these kind of remarks. Do you think you'd have followed an FAQ link from the repository's README, or would you have assumed that this isn't a behaviour described in an FAQ ?

NSeydoux:
Superseded by https://github.com/inrupt/solid-client-js/pull/1430

PR #1253, opened 2021-09-14
Feature/add get string english

pmcb55:
# New feature description

Added convenience functions `add/get/set/removeStringEnglish()` and  `add/get/set/removeStringEnglishAll()`.   We're still gently expressing the impossible-to-ignore relevance of locales in the underlying RDF (and 'cos trying to 'hide' that critical RDF-ness (e.g., via an implicitly-acting function like `addString()`) would lead to all sorts of confusion later (i.e., would it add an English language tag, or a NoLocale string literal?)).

# Checklist


src/thing/build.ts
```diff
@@ -117,6 +120,10 @@ export type ThingBuilder<T extends Thing> = {
   addDecimal: Adder<number, T>;
   addInteger: Adder<number, T>;
   addStringNoLocale: Adder<string, T>;
+  addStringEnglish: (
+    property: Parameters<typeof addStringWithLocale>[1],
+    value: Parameters<typeof addStringWithLocale>[2]
+  ) => ThingBuilder<T>;
```
NSeydoux, line 126: Couldn't this simply be of type `Adder<string, T>` ?

src/thing/build.ts
```diff
@@ -143,6 +150,10 @@ export type ThingBuilder<T extends Thing> = {
   setDecimal: Setter<number, T>;
   setInteger: Setter<number, T>;
   setStringNoLocale: Setter<string, T>;
+  setStringEnglish: (
+    property: Parameters<typeof setStringWithLocale>[1],
+    value: Parameters<typeof setStringWithLocale>[2]
+  ) => ThingBuilder<T>;
```
NSeydoux, line 156: ditto

src/thing/build.ts
```diff
@@ -170,6 +181,10 @@ export type ThingBuilder<T extends Thing> = {
   removeDecimal: Remover<number, T>;
   removeInteger: Remover<number, T>;
   removeStringNoLocale: Remover<string, T>;
+  removeStringEnglish: (
+    property: Parameters<typeof removeStringWithLocale>[1],
+    value: Parameters<typeof removeStringWithLocale>[2]
+  ) => ThingBuilder<T>;
```
NSeydoux, line 187: ditto

NSeydoux (approved):

src/thing/build.ts
pmcb55: Good question! I don't know (i.e., I don't understand these TypeScript types at all).
I was just cutting-and-pasting from the function directly below it (because my new function simply calls that existing one with one less parameter), i.e.,:
```
  addStringWithLocale: (
    property: Parameters<typeof addStringWithLocale>[1],
    value: Parameters<typeof addStringWithLocale>[2],
    locale: Parameters<typeof addStringWithLocale>[3]
  ) => ThingBuilder<T>;
```
So in effect, I don't know why half these functions have `: Adder<type, T>` and the other half have `:Parameters<typeof 'funcName'>[x]`. In other words, if you're right (i.e., we **_can_** just use `Adder<string, T>` here), then can't we use it on all these functions...?
```
  addTime: Adder<Time, T>;
  addDecimal: Adder<number, T>;
  addInteger: Adder<number, T>;
  addStringNoLocale: Adder<string, T>;
  addStringEnglish: (
    property: Parameters<typeof addStringWithLocale>[1],
    value: Parameters<typeof addStringWithLocale>[2]
  ) => ThingBuilder<T>;
  addStringWithLocale: (
    property: Parameters<typeof addStringWithLocale>[1],
    value: Parameters<typeof addStringWithLocale>[2],
    locale: Parameters<typeof addStringWithLocale>[3]
  ) => ThingBuilder<T>;
  addNamedNode: (
    property: Parameters<typeof addNamedNode>[1],
    value: Parameters<typeof addNamedNode>[2]
  ) => ThingBuilder<T>;
  addLiteral: (
    property: Parameters<typeof addLiteral>[1],
    value: Parameters<typeof addLiteral>[2]
  ) => ThingBuilder<T>;
```
But in fact, now I notice something: is the code I added wrong - i.e., should the function name be _reused_ in the `typeof [funcName]` part - i.e., should this code actually be like the other examples I just pasted above, and so my new function should be this instead:
```suggestion
  addStringEnglish: (
    property: Parameters<typeof addStringEnglish>[1],
    value: Parameters<typeof addStringEnglish>[2]
  ) => ThingBuilder<T>;
```
????

NSeydoux: `addStringWithLocale` and `addStringEnglish` having the same types for the first parameters ( a Thing, a property and a value), the current typing wasn't wrong, but indeed using `addStringEnglish` would be better. However, reusing the `Adder` type definition also works, and looks cleaner in my opinion. See ba5dd0cfdabec531f63efcd351b435d80210549c.

PR #1268, opened 2021-09-28
Setup end-to-end tests against dev-next environment

NSeydoux:
The dev-next environment is the preview server where new ESS features are released for testing before rolling to production. This sets up end-to-end tests in this environment so that issues and/or regressions are detected early.

The node end-to-end tests were already set up to allow for tests against multiple servers, but the testcafé tests required some slight changes.

This also includes calling `logout` for the node tests to prevent jest from hanging on the async token refresh.

Note that the repository secrets have been updated to be in line with the PR.


.github/workflows/ci.yml
```diff
@@ -93,6 +96,10 @@ jobs:
         E2E_TEST_ESS_POD: ${{ secrets.E2E_TEST_ESS_PROD_POD }}
```
matthieubosquet, line 96: Could we fix up the env variable order, so that it is consistent accross envs?

It's no big deal, but makes it easier to check we're consistent and there's no mistake. Possibly, doing an "order alphabetically" on such lists could be a good move.

(here in TEST_ESS, IDP_URL comes before POD)

Potentially, we might want to align the secret names too, that is, rename  E2E_TEST_ESS_PROD_IDP_URL to E2E_TEST_ESS_IDP_URL...

(a bit tedious I imagine, so not necessary right now)

src/e2e-browser/e2e.testcafe.ts
```diff
@@ -73,13 +73,48 @@ fixture("End-to-end tests")
...
+  identityProvider: string;
+  username: string;
+  password: string;
+}[] = [
+  // pod.inrupt.com:
+  {
+    // Cumbersome workaround, but:
+    // Trim `https://` from the start of these URLs,
+    // so that GitHub Actions doesn't replace them with *** in the logs.
+    identityProvider: process.env.E2E_TEST_ESS_IDP_URL!.replace(
```
matthieubosquet, line 86: Can we extract those replace into a utility function?

```javascript
function trimHttps(x: string): string {
  return x.replace(/^https:\/\//, "");
}
```

src/e2e-browser/roles.ts
```diff
@@ -38,15 +39,17 @@ export const essUser = Role("http://localhost:1234", async (t) => {
  * Move this function into `essUser` and replace uses of it with uses of the Role
  * once this issue is resolved: https://github.com/inrupt/solid-client-authn-js/issues/423
  */
-export const essUserLogin = async (_t: TestController) => {
+export const essUserLogin = async (
```
matthieubosquet, line 42:
```suggestion
export const userLogin = async (
```

Potentially, the tests might become more than just ESS, so maybe rename that variable?

matthieubosquet (approved):
Looks good to me.

.github/workflows/ci.yml
NSeydoux: I'll create a task in the backlog to reorder the variables. The naming pattern can also be reconsidered, but the presence of "PROD" is intentional, since `pod.inrupt.com` is our prod environment :). Bet we could consider changing `DEV_NEXT` into `ESS_DEV_NEXT`, even though `DEV_NEXT` is ESS-only, so it may make more sense to change `ESS_PROD` to just `PROD` (that's assuming we'll never have a CSS prod environment)

src/e2e-browser/roles.ts
NSeydoux: I'd approach this the other way around, and rename it if it became more. Right now, this is quite specific to ESS's Cognito page.

.github/workflows/ci.yml
matthieubosquet: I was looking at it from the point of view of the secret naming convention as it stands: **E2E_TEST_ESS_POD** vs secrets.E2E_TEST_ESS_**PROD**_POD.

I think secret and variable names should align.
I also think that you are correct, arguably, keeping the `PROD` part makes more sense.

But on second thoughts, more importantly, it might be a mistake to adapt our code and tests for this.
We should probably have only one series of tests that take standard environment variables and is run with the series of PROD variables, then the series of DEV_NEXT variable and so on...

So really, where we should go is having this in CI:
```bash
E2E_TEST_POD: ${{ secrets.E2E_TEST_ESS_PROD_POD }}
```

And adapt the workflow to run for each environment.

Then in our code, we'll keep it simple and run with parameterizable TEST_POD, TEST_IDP_URL...

matthieubosquet: (probably not for this PR depending on prioritisation and urgency)

src/e2e-browser/roles.ts
matthieubosquet: OK, makes sense.

PR #1310, opened 2021-10-20
Added flag to control diagnostic output on errors

pmcb55:
# New feature description

Added flag to control diagnostic output on errors (as this output can be distracting when errors are expected).

# Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The following changes have been implemented but not released yet:
 
 ## [Unreleased]
 
+### New feature
+
+- Added flag to control diagnostic output on errors.
+
 The following sections document changes that have been released already:
 
 ## [1.14.0] - 2021-10-15
--- a/src/resource/solidDataset.test.ts
+++ b/src/resource/solidDataset.test.ts
@@ -1040,6 +1040,7 @@ describe("saveSolidDatasetAt", () => {
         createSolidDataset(),
         {
           fetch: mockFetch,
+          outputDiagnosticsOnError: true,
         }
       );
 
@@ -1061,6 +1062,7 @@ describe("saveSolidDatasetAt", () => {
         createSolidDataset(),
         {
           fetch: mockFetch,
+          outputDiagnosticsOnError: true,
         }
       );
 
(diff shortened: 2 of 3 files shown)


andydavison (commented):
Couple of thoughts

- What's the use case for suppressing part of the error message (what situations are the errors expected)?
- Should this flag be added to internal_defaultFetchOptions, or is it only relevant for some requests?

pmcb55:
Yep:
"**_What's the use case for suppressing part of the error message (what situations are the errors expected)?_**"
The diagnostic part of the output contains the entire contents of the dataset as Markdown, so it's very noisy (it's useful output, generally, hence the flag is defaulted to 'true'). An expected error case is when doing ETL - we want the Loading-to-a-Pod process to be idempotent, so a resource may already exist (in which case we add any new triples to that resource), or may not exist at all - so a 404 error could be expected, normal behaviour.

"**_Should this flag be added to internal_defaultFetchOptions, or is it only relevant for some requests?_**"
Yep - absolutely. It was just that TypeScript was complaining when I tried to do that - whereas this code 'just worked'. But I'd be very happy to rework to simplify.

pmcb55:
Just marking this PR as draft, as after discussion with Andy today, we can make this cleaner by added the new flag to the `internal_defaultFetchOptions` object, but we should also add a proper TypeScript, and then change all references to `typeof internal_defaultFetchOptions` to use the type instead, and then use spreads to pull defaults, e.g.,:
```
  const config = {
    ...internal_defaultFetchOptions,
    ...options,
  };
```
Some places also use `options: Partial<typeof internal_defaultFetchOptions>`, which means all properties are optional. That means we could use that more widely and don’t necessarily need to create a separate ‘type’ (but we do still need to make sure everywhere handles missing values for fetch with the config).
So given there's a few options, and it needs careful consideration, flipping this to 'draft' (but the SDK could still merge this code to get it working now, and then tidy up everything **_properly_** in a separate ticket).

ThisIsMissEm:
This might be better solved by having a custom error type with a `captureDiagnostics()` method on it, which would then generate the extend output; keeping error messages smaller by default. (That would also put the control in the developers' hands, and would mean potentially sensitive information wouldn't end up by default in error reporting tools like Sentry.

ThisIsMissEm:
Closing due to this being heavily out of date, and before we add anything similar, we first need a Design Document for the feature.

PR #1317, opened 2021-11-01
Exporting type declarations in custom exports for TypeScript 4.5

Maximvdw:
The package.json uses Node.js 12+ exports (https://webpack.js.org/guides/package-exports/)
However, the current implementation does not work when using TypeScript 4.5.0 (https://devblogs.microsoft.com/typescript/announcing-typescript-4-5-beta/#packagejson-exports-imports-and-self-referencing) because it can not find the type definitions.

This PR adds a ```"types"``` entry point per export to add support for TypeScript 4.5.0

**Example from TS 4.5 beta announcement**
```json
// package.json
{
    "name": "my-package",
    "type": "module",
    "exports": {
        ".": {
            // Entry-point for `import "my-package"` in ESM
            "import": "./esm/index.js",

            // Entry-point for `require("my-package") in CJS
            "require": "./commonjs/index.cjs",

            // Entry-point for TypeScript resolution
            "types": "./types/index.d.ts"
        },
    },

    // CJS fall-back for older versions of Node.js
    "main": "./commonjs/index.cjs",

    // Fall-back for older versions of TypeScript
    "types": "./types/index.d.ts"
}
```

This additional property has no side effects for existing ES6 importing

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,33 +30,109 @@
   "exports": {
     ".": {
       "require": "./dist/index.js",
-      "import": "./dist/index.mjs"
+      "import": "./dist/index.mjs",
+      "types": "./dist/index.d.ts"
     },
-    "./interfaces": "./dist/interfaces.mjs",
-    "./resource/resource": "./dist/resource/resource.mjs",
-    "./resource/solidDataset": "./dist/resource/solidDataset.mjs",
-    "./resource/file": "./dist/resource/file.mjs",
-    "./resource/mock": "./dist/resource/mock.mjs",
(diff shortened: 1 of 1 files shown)


ThisIsMissEm (approved):
This looks good to me, but should probably wait on merge until the other API changes are out.

NSeydoux:
Reapplied in https://github.com/inrupt/solid-client-js/pull/1427 so that the CI runs and we can merge this. Thanks for contributing, and really sorry that it took this long to get merged !

PR #1326, opened 2021-11-08
WIP: Improve parser for WAC Allow header

ThisIsMissEm:
This new parser is more thoroughly tested, and also has a small efficiency gain: we only parse the header once, and then make assertions on it, rather than parsing once for each scope group.

There's not yet a ticket for this work, as it was just an improvement I saw that we could make to improve code quality (I wasn't working on anything else more important today whilst I was waiting for feedback on scheduled work)

Related work: https://github.com/solid/wac-allow

# Checklist

- [ ] Relevant documentation, if any, has been written/updated. *not sure if this needs updates, as it's an internal API*
- [ ] The changelog has been updated, if applicable. *not sure if this needs updates, as it's an internal API*
- [ ] New functions/types have been exported in `index.ts`, if applicable. *n/a*
- [ ] New modules (i.e. new `.ts` files) are listed in the `exports` field in `package.json`, if applicable. *n/a*
- [ ] New modules (i.e. new `.ts` files) are listed in the `typedocOptions.entryPoints` field in `tsconfig.json`, if applicable. *n/a*

### DIFF
--- a/src/resource/wacAllow.internal.ts
+++ b/src/resource/wacAllow.internal.ts
@@ -1,0 +1,109 @@
+/**
+ * Copyright 2021 Inrupt Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to use,
+ * copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the
+ * Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,
+ * INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A
+ * PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+ * HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+ * OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+ * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+// TODO: Should we move this type internally, as the Access in this file is not strictly the same as ACL Access
+import { Access } from "../acl/acl";
+
+export type internal_AccessPermissions = {
+  user: Access;
+  public: Access;
+};
+
+interface internal_parsedWacHeader {
+  [group: string]: string[];
+}
+
+function internal_parseWacHeader(header: string): internal_parsedWacHeader {
+  const rawEntries = header.split(",").map((rawEntry) => rawEntry.split("="));
+
+  const entries: internal_parsedWacHeader = {};
+
+  for (const rawEntry of rawEntries) {
+    if (rawEntry.length !== 2) {
+      continue;
+    }
+
+    const scope = rawEntry[0].trim();
+    const statement = rawEntry[1].trim();
+
+    if (!scope || !statement) {
+      continue;
+    }
+
+    if (
+      statement.length === 2 ||
+      statement.charAt(0) !== '"' ||
+      statement.charAt(statement.length - 1) !== '"'
+    ) {
+      continue;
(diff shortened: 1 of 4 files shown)


src/resource/wacAllow.internal.ts
```diff
@@ -0,0 +1,109 @@
...
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,
+ * INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A
+ * PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+ * HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+ * OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+ * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+// TODO: Should we move this type internally, as the Access in this file is not strictly the same as ACL Access
+import { Access } from "../acl/acl";
```
ThisIsMissEm, line 23: @NSeydoux @VirginiaBalseiro what're your thoughts on this one? I think parsing WAC Allow shouldn't be re-using data structures from ACL, as they are different things (the types just happen to currently be compatible)

src/resource/wacAllow.internal.ts
```diff
@@ -0,0 +1,109 @@
...
+      continue;
+    }
+
+    entries[scope] = accessModes.split(" ");
+  }
+
+  return entries;
+}
+
+function internal_getWacPermissions(
```
ThisIsMissEm, line 71: Not sure this is named correctly / strictly belongs here; perhaps `internal_getPermissions` would be better, as it's just checking a parsed WAC Allow for given values

src/resource/wacAllow.internal.ts
```diff
@@ -0,0 +1,109 @@
...
+      read: false,
+      append: false,
+      write: false,
+      control: false,
+    };
+  }
+
+  return {
+    read: permissions.includes("read"),
+    append: permissions.includes("append") || permissions.includes("write"),
```
ThisIsMissEm, line 88: This is the same logic as, but simpler than: https://github.com/inrupt/solid-client-js/blob/main/src/resource/resource.internal.ts#L87-L100

ThisIsMissEm (commented):
Highlighted a few areas in the diff that might be of interest

ThisIsMissEm:
I'm not sure why `CD / cd-end-to-end-tests (14.x) (push) Failing after 11m — cd-end-to-end-tests (14.x)` is failing — the output doesn't make a lot of sense to me.

src/resource/wacAllow.internal.ts
NSeydoux: Well it kind of makes sense that the `Access` definition defined in ACL is used for the WAC-Allow header, given that the WAC-Allow header was initially described in the ACL spec. We have another `Access` type definition from the `access/universal` submodule for universal (i.e. for both ACL-controlled and ACP-controlled Pods), which is somewhat different (because ACP has a different notion of Control than ACL), but could maybe used here, since the WAC-Allow header is now present in both ACL and ACP specs.

src/resource/wacAllow.internal.ts
NSeydoux: Sounds good to me, we can always move it if it becomes relevant elsewhere.

src/resource/wacAllow.internal.ts
NSeydoux: I think the way https://github.com/inrupt/solid-client-js/blob/main/src/resource/resource.internal.ts#L87-L100 is written was motivated by type definitions, if I recall correctly the TS compiler was complaining if the `append`and `write` values were not constants, but this looks definitely simpler.

NSeydoux (approved):

src/resource/wacAllow.internal.ts
matthieubosquet: The Access interface (there is no reason for it to be a type considering its declaration) should in my opinion:

1. Be moved to a type folder under src: `/src/type/`
2. Have all the modes
3. Be used everywhere such Objects are inputs/outputs (we need to remove all the aliasing and other similar definitions)

It would be particularly convenient to have one place where all the standard interfaces are defined.
We really need to remove all the aliasing going on, it is exactly the opposite of what interfaces are for. You don't want 20 names for the same thing, you want 1 name for a specific shape and as few of them as required.

As a sidenote, the ACP specification will adopt the standard Solid access modes shortly, that is, use the `acl:Control` mode just like WAC. It is a design flaw that it currently invents another mechanism. The operations entailed by Access Modes are defined at the Solid protocol level.

PR #1334, opened 2021-11-12
Add optional dataset param to createContainerAt()

pmcb55:
# New feature description

Add optional dataset param to createContainerAt(), which is really useful for adding meta-data to containers, for things like labels or comments, etc.

# Checklist

### DIFF
--- a/src/resource/solidDataset.test.ts
+++ b/src/resource/solidDataset.test.ts
@@ -2017,6 +2017,146 @@ describe("createContainerAt", () => {
     });
   });
 
+  describe("Creating non-empty container", () => {
+    function getMockUpdatedDataset(
+      changeLog: WithChangeLog["internal_changeLog"],
+      fromUrl: IriString
+    ): SolidDataset & WithChangeLog & WithResourceInfo {
+      const mockThing = addUrl(
+        createThing({ url: "https://arbitrary.vocab/subject" }),
+        "https://arbitrary.vocab/predicate",
+        "https://arbitrary.vocab/object"
+      );
+      let mockDataset = setThing(createSolidDataset(), mockThing);
+
+      changeLog.additions.forEach((tripleToAdd) => {
+        let additionThing =
+          getThing(mockDataset, tripleToAdd.subject.value) ??
+          createThing({ url: tripleToAdd.subject.value });
+        additionThing = addUrl(
+          additionThing,
+          tripleToAdd.predicate.value,
+          tripleToAdd.object.value
+        );
+        mockDataset = setThing(mockDataset, additionThing);
+      });
+
+      const resourceInfo: WithResourceInfo["internal_resourceInfo"] = {
+        sourceIri: fromUrl,
+        isRawData: false,
+      };
+
+      return {
+        ...mockDataset,
+        internal_changeLog: changeLog,
+        internal_resourceInfo: resourceInfo,
+      };
+    }
+
+    it("throws if container SolidDataset has updates", async () => {
+      const mockFetch = jest
+        .fn(window.fetch)
+        .mockReturnValue(
+          Promise.resolve(
+            mockResponse(undefined, { url: "https://arbitrary.pod/container/" })
+          )
+        );
+
+      const mockDataset = getMockUpdatedDataset(
+        {
+          additions: [
+            DataFactory.quad(
+              DataFactory.namedNode("https://some.vocab/subject"),
+              DataFactory.namedNode("https://some.vocab/predicate"),
+              DataFactory.namedNode("https://some.vocab/object"),
+              undefined
+            ),
+          ],
+          deletions: [],
+        },
+        "https://arbitrary.pod/container/"
+      );
+
+      await expect(
+        createContainerAt(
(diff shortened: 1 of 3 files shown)


src/resource/solidDataset.ts
```diff
@@ -489,7 +489,7 @@ export async function deleteSolidDataset(
 }
 
 /**
- * Create an empty Container at the given URL.
+ * Create a Container at the given URL (which optionally can have it's only meta-data).
```
NSeydoux, line 492:
```suggestion
 * Create a Container at the given URL. Some content may optionally be specified, e.g. to add metadata describing the container.
```

src/resource/solidDataset.ts
```diff
@@ -500,13 +500,15 @@ export async function deleteSolidDataset(
...
+ * @param solidDataset Optional parameter - if provided we use this dataset as the body of the HTT request, meaning it's data is included in the Container resource.
  * @since 0.2.0
  */
 export async function createContainerAt(
   url: UrlString | Url,
   options: Partial<
     typeof internal_defaultFetchOptions
-  > = internal_defaultFetchOptions
+  > = internal_defaultFetchOptions,
+  solidDataset?: SolidDataset
```
NSeydoux, line 511:
```suggestion
    initialContent: SolidDataset
  > = internal_defaultFetchOptions,
```
Since an `options` object is already passed, the additional optional parameter should be added to this options object, rather than added as an additional parameter. This way, the following would be used: `createContainerAt(someUrl, { initialContent: someDataset})` rather than `createContainerAt(someUrl, undefined, someDataset)`

src/resource/solidDataset.ts
```diff
@@ -515,17 +517,55 @@ export async function createContainerAt(
...
+        "Content-Type": "text/turtle",
+        "If-None-Match": "*",
+        // This header should not be required to create a Container,
+        // but ESS currently expects it:
+        Link: `<${ldp.BasicContainer}>; rel="type"`,
+      },
+    });
+
+    datasetToReturn = solidDataset;
+  }
```
NSeydoux, line 568:
```suggestion
if (config.solidDataset !== undefined && isUpdate(internal_withChangeLog(solidDataset), url)) {
  // If we are creating a new container, then it can't have existed before,
  // so we can't be doing an 'update' to it. So if the provided dataset has
  // updates then we assume there was a programmer error. (This may require
  // changing based on user feedback (e.g., perhaps the dataset was read
  // from a different container, we augmented that, and now want to use it
  // as meta-data for our new container)).
  throw new Error(
    `Attempting to create a new container with a dataset that contains updates!`
  );
}

const response = await config.fetch(url, {
    method: "PUT",
    body: config.solidDataset 
      ? await triplesToTurtle(toRdfJsQuads(solidDataset).map(getNamedNodesForLocalNodes))
      : undefined,
    headers: {
      Accept: "text/turtle",
      "Content-Type": "text/turtle",
      "If-None-Match": "*",
      // This header should not be required to create a Container,
      // but ESS currently expects it:
      Link: `<${ldp.BasicContainer}>; rel="type"`,
    },
  });
```

Throwing the error early allows to reduce code duplication.

src/resource/solidDataset.ts
```diff
@@ -547,7 +588,7 @@ export async function createContainerAt(
   const containerDataset: SolidDataset &
     WithChangeLog &
     WithServerResourceInfo = freeze({
-    ...createSolidDataset(),
+    ...datasetToReturn,
```
NSeydoux, line 591:
```suggestion
    ...options.solidDataset ?? createSolidDataset(),
```

src/resource/solidDataset.test.ts
```diff
@@ -2017,6 +2017,146 @@ describe("createContainerAt", () => {
...
+        "https://arbitrary.pod/container/"
+      );
+
+      await expect(
+        createContainerAt(
+          "https://arbitrary.pod/container/",
+          {
+            fetch: mockFetch,
+          },
+          mockDataset
```
NSeydoux, line 2086:
```suggestion
          {
            fetch: mockFetch,
            initialContent: mockDataset
          }
```
See the suggested signature change below.

src/resource/solidDataset.test.ts
```diff
@@ -2017,6 +2017,146 @@ describe("createContainerAt", () => {
...
+      ).toBe("text/turtle");
+      expect(
+        (mockFetch.mock.calls[0][1]?.headers as Record<string, string>)["Link"]
+      ).toBe('<http://www.w3.org/ns/ldp#BasicContainer>; rel="type"');
+      expect((mockFetch.mock.calls[0][1]?.body as string).trim()).toBe(
+        "<https://arbitrary.vocab/subject> <https://arbitrary.vocab/predicate> <https://arbitrary.vocab/object>."
+      );
+    });
+
+    it("reports HTTP error when an non-empty SolidDataset provided", async () => {
```
NSeydoux, line 2133: I think the suggested refactoring makes this test redundant, as one branch has been removed from the code.

NSeydoux (approved):

PR #1335, opened 2021-11-12
Discover personal profile from WebID profile doc

NSeydoux:
This adds two functions to the API which enable discovering personal profile documents from WebID profile documents. The purpose is to make it easier to link from a non-Solid WebID profile document to a Solid personal profile, so that the WebID profile may not be editable but the personal profile is.

# Checklist


CHANGELOG.md
```diff
@@ -6,6 +6,12 @@ The following changes have been implemented but not released yet:
 
 ## [Unreleased]
 
+### New features
+
+- `getWebIdProfileAll` and `getWebIdProfileAll`: functions to discover one's personal
+  profile (which may be a Solid resource) if discoverable from their WebID profile
+  document (which may NOT be a Solid resource).
```
matthieubosquet, line 13:
```suggestion
- `getProfileAll` and `getProfile`: functions to discover the profile documents
  associated with a WebID profile document. Note that both the WebID profile
  document and its associated profile documents may or may not be a Solid
  resource (said resource(s) would only be editable via the SDK in the former case).
```

src/index.test.ts
```diff
@@ -176,6 +176,8 @@ import {
   addPublicKeyToProfileJwks,
   getProfileJwksIri,
   setProfileJwks,
+  getWebIdProfile,
+  getWebIdProfileAll,
```
matthieubosquet, line 180:
```suggestion
  getProfile,
  getProfileAll,
```

src/index.test.ts
```diff
@@ -357,6 +359,8 @@ it("exports the public API from the entry file", () => {
   expect(addPublicKeyToProfileJwks).toBeDefined();
   expect(getProfileJwksIri).toBeDefined();
   expect(setProfileJwks).toBeDefined();
+  expect(getWebIdProfile).toBeDefined();
+  expect(getWebIdProfileAll).toBeDefined();
```
matthieubosquet, line 363:
```suggestion
  expect(getProfile).toBeDefined();
  expect(getProfileAll).toBeDefined();
```

src/profile/webid.ts
```diff
@@ -0,0 +1,106 @@
...
+ */
+export async function getWebIdProfile(
+  webIdDocument: SolidDataset & WithResourceInfo,
+  options: Partial<
+    typeof internal_defaultFetchOptions
+  > = internal_defaultFetchOptions
+): Promise<SolidDataset> {
+  // Arbitrarily pick a subject of a triple <s, foaf:primaryTopic, o>
+  const [primaryThing] = getThingAll(webIdDocument)
+    .filter((thing) => getIriAll(thing, foaf.primaryTopic).length > 0)
```
matthieubosquet, line 99: We need to catter for the inverse property `foaf:isPrimaryTopicOf`.

matthieubosquet:
```suggestion
  const [primaryThing] = getThingAll(webIdDocument)
    .filter((thing) => getIriAll(thing, foaf.primaryTopic).length > 0 || getIriAll(thing, foaf.isPrimaryTopicOf).length > 0)
```
Something like that but taking predicate object instead of subject predicate. Not sure what that function is in the SDK.

matthieubosquet (changes requested):
Mainly renaming is required (I started to suggest, but I think a find/replace would do better so I stopped).
Also, we need to catter for the inverse property of `foaf:primaryTopic`.

src/profile/webid.ts
```diff
@@ -0,0 +1,104 @@
...
+ *  profile document discoverable from the WebID Profile Document. If several exist,
+ *  one is arbitrarily chosen.
+ *  If none are found, the WebID profile document itself is returned.
+ */
+export async function getProfile(
+  webIdProfile: SolidDataset & WithResourceInfo,
+  options: Partial<
+    typeof internal_defaultFetchOptions
+  > = internal_defaultFetchOptions
+): Promise<SolidDataset> {
```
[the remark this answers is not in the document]
matthieubosquet: So actually, the getSolidDataset function I see returns `SolidDataset & WithServerResourceInfo`.
So, we should probably just return that (and consider adding the extension to not loose info in a refactor).
```suggestion
export async function getProfile(
  webIdProfile: SolidDataset & WithResourceInfo,
  options: Partial<
    typeof internal_defaultFetchOptions
  > = internal_defaultFetchOptions
): Promise<SolidDataset & WithServerResourceInfo> {
```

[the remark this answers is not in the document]
NSeydoux: a38e26bb

src/profile/webid.ts
NSeydoux: This should now be resolved.

CHANGELOG.md
```diff
@@ -6,6 +6,12 @@ The following changes have been implemented but not released yet:
 
 ## [Unreleased]
 
+### New features
+
+- `getProfileAll`: functions to discover one's personal profile (which may be a
+  Solid resource) if discoverable from their WebID profile document (which may NOT
+  be a Solid resource).
```
matthieubosquet, line 13:
```suggestion
- `getProfileAll`: function to discover the WebID Profile Document and its associated
  FOAF Profile Document resources from the WebID URI. Note that Profile documents
  may or not be Solid Resources
```

It is not necessarily "one's" or "personal" (a bot/group/organisation can be identified via a WebID).
All profile documents may or not be a Solid resource (the warning going one way, then the other is confusing to me).

matthieubosquet (approved):
Small comment about the function description and stylistic suggestion (with suggestion about the SDK API design).

But that looks good to me! :)

CHANGELOG.md
matthieubosquet: PS: also, there is only one function, so either way we need to loose the "s".

NSeydoux:
Regarding matching on `<?s, pred, obj>`, we could add filters to `getThingAll` to allow that kind of operations.

matthieubosquet:
> Regarding matching on <?s, pred, obj>, we could add filters to getThingAll to allow that kind of operations.

That sounds really good ^^,

Not sure what the exact final design would be, but I think it would help simplify code and logic in many cases if we get it right.

src/profile/webid.ts
```diff
@@ -0,0 +1,98 @@
...
+  const { fetch, webIdProfile } = options;
+
+  const profileDocument =
+    webIdProfile ?? ((await getSolidDataset(webId, { fetch })) as T);
+
+  const profilesLinkedFrom = getThingAll(profileDocument)
+    .filter((thing) => getIriAll(thing, foaf.primaryTopic).length > 0)
+    // This assumes that getSourceIri returns the IRI where the profile document
+    // has actually been fetched, which may differ from the WebID.
+    .filter((thing) => asIri(thing) !== getSourceIri(profileDocument))
```
ThisIsMissEm, line 78: don't you only need this `getSourceIri` if you've passed a `webIdProfile`? shouldn't `getSourceIri(profileDocument) === webId`?

src/profile/webid.ts
```diff
@@ -0,0 +1,98 @@
...
+        (candidateProfile) => candidateProfile !== getSourceIri(profileDocument)
+      )
+      .map((profile) => getSolidDataset(profile, { fetch }));
+  }
+
+  return {
+    webIdProfile: profileDocument,
+    altProfileAll: await Promise.all([
+      ...profilesLinkedFrom,
+      ...profilesLinkedTo,
```
ThisIsMissEm, line 95: should this list be de-duped?

NSeydoux: Will fix this in a new PR

src/profile/webid.ts
ThisIsMissEm: Right, but we don't need the `getRedirectedIri` or `getActualIri` for now if we don't allow passing in an existing Solid Dataset instance/object, and just accept the URL/IRI of the profile, which more closely meets usage.

```
const profile = getProfileAll(sessionInfo?.webId, { fetch: session.fetch })
```

The above is roughly taken from the react-sdk, as long as the API here is:

```
getProfileAll(webId: WebId, options)
```

Then we can just check does the `foaf:primaryTopic` equal the `webId` argument, as my understanding is that we'd generally be using this API as:

```
getProfileAll("https://some.webid")
```

We only need to worry about the `getSourceIri` being incorrect if `getProfileAll` accepts a profile document and doesn't do the query for it itself. (I'd be inclined to split this in two iterations or have a helper:

```
getLinkedProfiles(profileDocument, profileIri)
```

That way if you really need to go the path of getting the linked profiles from an arbitrary solid dataset that's been preloaded somewhere, you can, but it's on the consumer to track that profileIri that they're looking to match against. Where as `getProfileAll` can just accept a `WebId` IRI, and call it a day.

PR #1336, opened 2021-11-12
Temporarily prevent e2e test failure on dev-next

NSeydoux:
There appears to be an issue with the universal library on ESS 1.2. This prevents the test from failing while we are investigating what is going wrong. while investigating, I also changed the way logging in/out is implemented to something cleaner.

### DIFF
--- a/src/e2e-node/e2e.test.ts
+++ b/src/e2e-node/e2e.test.ts
@@ -19,7 +19,14 @@
  * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
-import { jest, describe, it, expect } from "@jest/globals";
+import {
+  jest,
+  describe,
+  it,
+  expect,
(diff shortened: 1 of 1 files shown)


src/e2e-node/e2e.test.ts
```diff
@@ -373,19 +381,23 @@ describe.each(serversUnderTest)(
...
     }
 
-    async function getSession() {
-      const session = new Session();
+    let session: Session;
+
+    beforeEach(async () => {
+      session = new Session();
       await session.login({
         oidcIssuer: oidcIssuer,
```
matthieubosquet, line 389:
```suggestion
        oidcIssuer,
```

matthieubosquet: (since we addapt the style of assignment below ^^,)

matthieubosquet (approved):
All good!

PR #1342, opened 2021-11-16
De-duplicate values when discovering profiles

NSeydoux:
It is possible that a profile document contains both <webid, foaf:isPromaryTopicOf, someProfile> and <someProfile, foaf:primaryTopic, webid>. In that case, we only want someProfile to be fetched and returned once.


src/profile/webid.ts
```diff
@@ -71,28 +71,38 @@ export async function getProfileAll<T extends SolidDataset & WithResourceInfo>(
...
+      .map((primaryThing) => getSolidDataset(asIri(primaryThing), { fetch }))
+  );
+
+  const fetchedProfilesIri = profilesLinkedFrom.map(getSourceIri);
 
   let profilesLinkedTo: Array<Promise<SolidDataset & WithResourceInfo>> = [];
   const webIdThing = getThing(profileDocument, webId);
   if (webIdThing !== null) {
     profilesLinkedTo = getIriAll(webIdThing, foaf.isPrimaryTopicOf)
       .filter(
```
matthieubosquet, line 89:
```suggestion
  const { fetch, webIdProfile = await getSolidDataset(webId, { fetch }) as T } = options;

  const webIdThing = getThing(profileDocument, webId);

  const altProfilesIri = getThingAll(profileDocument)
    .filter(thing => getIriAll(thing, foaf.primaryTopic).length > 0)
    .map(asIri)
    .concat(
      webIdThing ? getIriAll(webIdThing, foaf.isPrimaryTopicOf) : []        
    )
    .filter(profileIri => profileIri !== getSourceIri(profileDocument));
    
  return {
    webIdProfile,
    altProfileAll: await Promise.all(
      [...new Set(altProfilesIri)]
        .map(uniqueProfileIri => getSolidDataset(asIri(primaryThing), { fetch })
    )
  };
```

matthieubosquet: Maybe a bit less repetition if starting by extracting the IRIs and destructured default assignment.
(I just wrote this here, so needs to be tested ^^,)

matthieubosquet (approved):
LGTM pending stylistic comment.

src/profile/webid.test.ts
```diff
@@ -194,4 +195,34 @@ describe("getProfileAll", () => {
...
+    );
+    // The profile document will have two triples <profile, foaf:primaryTopic, webid>...
+    const profileContent = buildThing({ url: "https://some.profile" })
+      .addIri(foaf.primaryTopic, MOCK_WEBID)
+      .build();
+    // and <webid, foaf:isPrimaryTopicOf, profile>.
+    const webidData = buildThing({ url: MOCK_WEBID })
+      .addIri(foaf.isPrimaryTopicOf, "https://some.profile")
+      .build();
+    const webIdProfile = [profileContent, webidData].reduce(
```
ThisIsMissEm, line 216: I wouldn't recommend using `reduce` here, and instead just do the two (or three) discrete operations

src/profile/webid.ts
ThisIsMissEm: I'd go similar to the above suggestion, but also de-duplicating between the webIdProfile and the alt profiles, essentially an `IrisToFetch` and then de-structure that into the shapes we want to return. I'd also lift the `await Promise.all` up into a constant.

Also, `Array.from(new Set(altProfilesIri))` instead of the spread.

There's also a question of how should this fail? Should it only fail if the webIdProfile fails to load, or should it fail if all the altProfiles fail to load?

NSeydoux: I'm not sure we need de-duplicating between the webid profile and the alt ones, as that would involve the triples in the shape `<webid (foaf:primaryTopic|foaf:isPrimaryTopicOf) webid>`, which doesn't really make sense.

I think it should fail if any profile, the WebID one or the alt ones, fail to load. That's the current behaviour.

src/profile/webid.test.ts
NSeydoux: Done in ac286587. Do you think `reduce` is less readable than the alternative here ?

src/profile/webid.ts
NSeydoux: 920ad00b

matthieubosquet: > Also, Array.from(new Set(altProfilesIri)) instead of the spread.

@ThisIsMissEm I think the spread is the most idiomatic and concise just like destructured assignments of options. Obviously it's a matter of preference, but I think it's just more readable/scanable/visually easy to parse. That's always the angle I come from in my preference of syntax.

So a few that I love to use here represented:

```
// Destructured assignment of options with defaults
const options = { a: 0 };
const { a = 1, b = 1 } = options;
```

```
// Variable assignment in objects (not sure what the precise name is)
const a = 1;
const x = { a };
```

```
// Unique values in array
const a = [ 1, 1, 2, 3 ];
const unique = [ ...new Set(a) ];
```

NSeydoux: Aaah, I didn't realize the destructured assignment had a default value ^^. Fixed that in 581bfe3d.

NSeydoux: However, I think my preference goes to `Array.from()` rather than destructuring too from a readability standpoint, but it's purely a personal opinion. Are there additional elements to consider in this choice ?

PR #1345, opened 2021-11-18
Fix profile type

NSeydoux:
The profiles are returned from the server, and they will have server metadata such as the WAC-allow header. Not adding this to the typing is information loss, and prevents from using useful functions such as `getEffectiveAccess`.


matthieubosquet (approved):
Hear hear!

PR #1356, opened 2021-11-23
WIP: open discussion on typing

NSeydoux:
This commit includes two changes I want to bring up for discussion:
- Defining the constants as raw literals makes the type system more assertive (here it goes from "string" to "http://www.w3.org/ns/solid/acp#allow" | "http://www.w3.org/ns/solid/acp#deny", which I find much more helpful
- Making a function generic without changing its return type should not be necessary, because anything that matches ThingPersisted should be accepted in the proposed form, which means the generic does not add anything.

This is just a partial refactoring I want to use as a basis for conversation, it doesn't change any behaviour.

### DIFF
--- a/src/acp/constants.ts
+++ b/src/acp/constants.ts
@@ -24,25 +24,25 @@ export const ACP_NAMESPACE = "http://www.w3.org/ns/solid/acp#";
 
 /** @hidden */
 export const ACP = {
-  AccessControl: ACP_NAMESPACE.concat("AccessControl"),
-  AccessControlResource: ACP_NAMESPACE.concat("AccessControlResource"),
-  AuthenticatedAgent: ACP_NAMESPACE.concat("AuthenticatedAgent"),
-  CreatorAgent: ACP_NAMESPACE.concat("CreatorAgent"),
-  Matcher: ACP_NAMESPACE.concat("Matcher"),
-  Policy: ACP_NAMESPACE.concat("Policy"),
-  PublicAgent: ACP_NAMESPACE.concat("PublicAgent"),
-  access: ACP_NAMESPACE.concat("access"),
-  accessControl: ACP_NAMESPACE.concat("accessControl"),
-  agent: ACP_NAMESPACE.concat("agent"),
-  allOf: ACP_NAMESPACE.concat("allOf"),
-  allow: ACP_NAMESPACE.concat("allow"),
-  anyOf: ACP_NAMESPACE.concat("anyOf"),
-  apply: ACP_NAMESPACE.concat("apply"),
-  client: ACP_NAMESPACE.concat("client"),
-  deny: ACP_NAMESPACE.concat("deny"),
-  memberAccessControl: ACP_NAMESPACE.concat("memberAccessControl"),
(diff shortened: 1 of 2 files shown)


ThisIsMissEm (approved):
This looks good to me, also far easier to read than the concat + some editors will make the link actually clickable too.

matthieubosquet:
Let's start with generic parameters.

What I see in my development environment is the following:
Screenshot 2021-11-25 at 14 48 33


This illustrates that anything matching the structure of a type cannot be fed into non generic parameters.

Can you not reproduce this type Error?

Code snippet:
```typescript
interface TypeX {
  a: string
}

function A(
  x: TypeX
): string {
  return x.a;
}

function B<T extends TypeX>(
  x: T
): string {
  return x.a;
}

A({ a: "bla", b: "bla" });

B({ a: "bla", b: "bla" });
```

ThisIsMissEm:
> This looks good to me, also far easier to read than the concat + some editors will make the link actually clickable too.

I didn't actually see the change to `getModes` when I reviewed — the string changes look good to me though.

NSeydoux:
Regarding the generic parameters, the behaviour you observe happens when you pass in a "plain" object, and not an explicitly typed variable. The following code snippet is valid too: 

```
interface TypeX {
  a: string;
}

function A(x: TypeX): string {
  return x.a;
}

function B<T extends TypeX>(x: T): string {
  return x.a;
}

const myLiteral = { a: "bla", b: "bla" };
const myTypedVar: TypeX = myLiteral;

A(myTypedVar);
B({ a: "bla", b: "bla" });
```
 Assigning `const myTypedVar: TypeX =  { a: "bla", b: "bla" } ;` directly would fail though, for the same reason you're pointing out. Anything that is a supertype of `TypeX` is valid as an input to function `A` without generics, so I don't think we need them in this case. I think that the following also provides an interesting insight: https://effectivetypescript.com/2020/08/12/generics-golden-rule/

ThisIsMissEm:
```
const myLiteral = { a: "bla", b: "bla" };
const myTypedVar: TypeX = myLiteral;
```

This feels like it should fail and be an error..

NSeydoux:
> `const myLiteral = { a: "bla", b: "bla" };`
> `const myTypedVar: TypeX = myLiteral;`
> This feels like it should fail and be an error..

Well the way I see it, `myLiteral` is of a supertype of `TypeX`, but `TypeX` would be a valid interface for `myLiteral`. I don't think TS types give any guarantee about things that should not be there, they only assert about what you can expect to find in my understanding. That's why it's valid to have `const myVar: TypeA & TypeB` being passed to a function that accepts `TypeA`, or to another function that accepts `TypeB`.

matthieubosquet:
We have the same understanding Nicolas.

The syntax I'm proposing only really absolves you from casting. You can very well write:
- `A({ a: "bla", b: "bla" } as TypeX);`
- `const x = { a: "bla", b: "bla" } as TypeX;`
- `const x = { a: "bla", b: "bla" }; A(x);`

Whether that syntax is more desirable than having a generic type input I'm not sure yet.

NSeydoux:
This is outdated now.

PR #1359, opened 2021-11-24
WIP: fix circular dependencies

ThisIsMissEm:
Just a branch where I started to look into the circular dependencies we currently have in the codebase; They're not particularly hard to solve, just takes time and causes code churn.


NSeydoux:
I'll close this PR, because it's been open for so long that it will be much more work to rebase than to restart from scratch. The underlying issue still exists and should be addressed.

PR #1361, opened 2021-11-25
Implement get and set public access for ACP Draft

matthieubosquet:
## Get and set public access for ACP Draft

### DIFF
--- a/src/acp/util/setPublicAccess.test.ts
+++ b/src/acp/util/setPublicAccess.test.ts
@@ -1,0 +1,68 @@
+/**
+ * Copyright 2021 Inrupt Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to use,
+ * copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the
+ * Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,
+ * INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A
+ * PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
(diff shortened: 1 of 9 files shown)


src/acp/util/setPublicAccess.test.ts
```diff
@@ -0,0 +1,68 @@
...
+import {
+  DEFAULT_ACCESS_CONTROL_RESOURCE_URL,
+  TEST_URL,
+} from "../mock/constants";
+import { createDatasetFromSubjects } from "../mock/dataset";
+import { mockAccessControlledResource } from "../mock/mockAccessControlledResource";
+import { setPublicAccess } from "./setPublicAccess";
+
+describe("setPublicAccess()", () => {
+  it("sets an access mode for the Public", async () => {
```
NSeydoux, line 33: Do we have a tests verifying that `setXAccess` overrides existing access ?

src/e2e-node/acp.test.ts
```diff
@@ -125,6 +127,28 @@ describe("An ACP Solid server", () => {
     });
   });
 
+  it("can get and set read access for the public", async () => {
```
NSeydoux, line 130: Should this test also check that actually doing an unauthenticated fetch succeeds ?

NSeydoux (approved):

src/e2e-node/acp.test.ts
matthieubosquet: Excellent suggestion, I added this in https://github.com/inrupt/solid-client-js/pull/1361/commits/c91b02b43f5b2bdf20ce942b0396e348b3aa8062.

Problem is, I don't know if I'm missing something, but the unauthenticated fetch still fails after setting public access.

Do you see something obvious I'm missing or is it something we need to report?

src/acp/util/setPublicAccess.test.ts
matthieubosquet: The setXAccess functions are adding and removing access only in the default elements.

That is tested.

I think it is an alright (and documented) limitation, that is at least until we properly support external resources consistently in access control. And even if we did that, I am not sure we'd want to automatically update all the conflicting access rather than produce an output hinting at the conflicts to be addressed.

src/e2e-node/acp.test.ts
matthieubosquet: Oh, that’s interesting:
https://github.com/inrupt/solid-client-js/runs/4329049297?check_suite_focus=true#step:7:30

src/acp/util/setPublicAccess.test.ts
NSeydoux: Ah right sorry, I forgot that we were working under the assumption that these access only apply to the default elements. That's alright then for now, as indeed a different approach requires more discussion.

src/e2e-node/acp.test.ts
NSeydoux: Ah damn, I hoped this was an issue caused by the ACP implementation, that's why I suggetsed adding this test.

PR #1369, opened 2021-11-30
Enable public access end-to-end test

NSeydoux:
The ESS bug which prevented this test from passing has been resolved.


matthieubosquet:
To enable public access end-to-end tests, this is all that's required: https://github.com/inrupt/solid-client-js/pull/1375

I think we should work with environments and the bits of code you're deleting which in my opinion make the tests more readable and eliminate boilerplate code enables that. I think we should have a discussion about that point before just deleting this code which works well.

NSeydoux:
I am open to discussing the way tests are setup, but I'd rather have the discussion first, and then introduce the changes. The code which was replaced has less functionalities than the code I introduced, as it only tests against one server, and if we agree to use the alternative you proposed, I'd rather have it implemented once for all end-to-end tests of the repo, and not on some of the tests but not all. 

Note that the end-to-end test is failing in https://github.com/inrupt/solid-client-js/runs/4381303128?check_suite_focus=true: I initially did the same thing as https://github.com/inrupt/solid-client-js/pull/1375, realized things were not working, and tripped on the authentication because it was inconsistent with the rest of the tests, for which my local setup is configured. That's why I removed the setup files.

ThisIsMissEm:
I think we'd be better using a testing matrix here, so we just have the tests use one environment by default, no `.each`, and then in CI we run against all servers

ThisIsMissEm:
> I think we'd be better using a testing matrix here, so we just have the tests use one environment by default, no `.each`, and then in CI we run against all servers

Further, we've currently 3 different sets of env variables used for tests:

```
# For ACP:
E2E_TEST_POD=
E2E_TEST_IDP=
E2E_TEST_CLIENT_ID=
E2E_TEST_CLIENT_SECRET=

# For E2E:
E2E_TEST_ESS_POD=
E2E_TEST_ESS_IDP_URL=
E2E_TEST_ESS_CLIENT_ID=
E2E_TEST_ESS_CLIENT_SECRET=

E2E_TEST_DEV_NEXT_POD=
E2E_TEST_DEV_NEXT_IDP_URL=
E2E_TEST_DEV_NEXT_CLIENT_ID=
E2E_TEST_DEV_NEXT_CLIENT_SECRET=
```

I would be very happy to refactor this to just one set, and then we can just run the node e2e on one environment for dev-next / CSS.

src/e2e-node/acp.test.ts
```diff
@@ -39,201 +38,260 @@ import {
...
-      read: true,
-      append: true,
-      write: false,
-      controlRead: false,
-      controlWrite: false,
+// End-to-end tests against ESS are temporarily disabled while we work out
+// how to authenticate against it now that refresh tokens are rotated after
+// every request:
+// eslint-disable-next-line jest/no-disabled-tests
+describe.each(serversUnderTest)(
```
ThisIsMissEm, line 96: I think this adds unnecessary complexity, let's actually define a clearer e2e testing strategy; this way of defining a testing matrix inside a testsuite is kinda strange, and means we're running many more tests than we really need to be.

ThisIsMissEm (changes requested):

ajacksified:
> I would be very happy to refactor this to just one set, and then we can just run the node e2e on one environment for dev-next / CSS.

I think it's inevitable that if we're testing multiple environments we'll need multiple sets of credentials. That said, we don't need to run every version of node against every os against every server. We should run unit tests against each os and node version we support, and stable node + linux against each server for e2e testing.

matthieubosquet:
OK, so for your consideration, I have enabled and used pretty succinctly and succesfully GitHub environments in this PR: https://github.com/inrupt/solid-client-js/pull/1375/files

See the result in this test run: https://github.com/inrupt/solid-client-js/runs/4401039688

NSeydoux:
Replaced with https://github.com/inrupt/solid-client-js/pull/1375

PR #1374, opened 2021-11-30
Format yml for github workflows according to prettier

ThisIsMissEm:
This was causing conflicts and issues whenever I was dealing with the yml files, as I'd accidentally save with formatting and cause a huge diff instead of just the lines I wanted to change. Functionally there should be no changes (assuming prettier works correctly, which I've never known it not to)

### DIFF
--- a/package.json
+++ b/package.json
@@ -111,7 +111,7 @@
   },
   "lint-staged": {
     "*.{js,jsx,ts,tsx}": "eslint --cache --fix",
-    "*.{ts,tsx,js,jsx,css,md,mdx}": "prettier --write"
+    "*.{ts,tsx,js,jsx,css,md,mdx,yml}": "prettier --write"
   },
   "resolutions": {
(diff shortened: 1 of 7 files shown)


NSeydoux (approved):
Looks good to me. Note that I pushed an additional commit so that prettier explicitly runs on staged YAML files so that we keep these files aligned.

package.json
```diff
@@ -111,7 +111,7 @@
   },
   "lint-staged": {
     "*.{js,jsx,ts,tsx}": "eslint --cache --fix",
-    "*.{ts,tsx,js,jsx,css,md,mdx}": "prettier --write"
+    "*.{ts,tsx,js,jsx,css,md,mdx,yml}": "prettier --write"
```
ThisIsMissEm, line 114: Oh! Good call @NSeydoux!

matthieubosquet (approved):

PR #1375, opened 2021-12-01
Enable e2e test public access

matthieubosquet:


NSeydoux (approved):
Using separate environments makes things nicer indeed !

PR #1376, opened 2021-12-01
Do not use `async` function if not necessary

NSeydoux:
The ACP low-level API had a couple of functions marked as `async`, but not making any asynchronous calls. Fetching and saving the ACR is done outside of these functions, which do regular in-memory data manipulation.

If we want to build a higher-level ACP API eventually, which implicitly fetches the ACR if necessary, we can do it later, but it should be done explicitly and designed independently of the existing API.

Alternatively, we can drop the changes included in this PR, and mark `setVcAccess` and `getVcAccess` as `async` for consistency.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ The following changes have been implemented but not released yet:
 
 ## [Unreleased]
 
+### Bugfixes
+
+- Some of the `acp/acp` functions, namely `getAgentAccess`, `setAgentAccess`,
+  `getPublicAccess` and `setPublicAccess` were marked async, but did not perform
+  asynchronous operations.
+
 The following sections document changes that have been released already:
 
 # [1.17.0] - 2021-11-29
--- a/src/acp/util/getAgentAccess.ts
+++ b/src/acp/util/getAgentAccess.ts
@@ -110,10 +110,10 @@ function reduceModes(
  * @param webId WebID of the Agent you want to get the access for.
  * @since 1.16.0
  */
-export async function getAgentAccess<T extends WithAccessibleAcr>(
+export function getAgentAccess<T extends WithAccessibleAcr>(
   resourceWithAcr: T,
   webId: WebId
-): Promise<AccessModes> {
+): AccessModes {
   // TODO: add support for external resources
   let resourceAccess = {
     read: false,
--- a/src/acp/util/getPublicAccess.test.ts
+++ b/src/acp/util/getPublicAccess.test.ts
@@ -34,7 +34,7 @@ describe("getPublicAccess()", () => {
     const resource = mockAccessControlledResource();
 
     expect(resource.internal_acp.acr.graphs).toStrictEqual({ default: {} });
-    expect(await getPublicAccess(resource)).toStrictEqual({
+    expect(getPublicAccess(resource)).toStrictEqual({
       read: false,
       append: false,
       write: false,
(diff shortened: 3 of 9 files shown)


ThisIsMissEm:
Not sure how I feel about this, wouldn't it be better to make them async even if internally for now they're sync, such that evolving that code won't break downstream consumers in the future?

NSeydoux:
I think this was @matthieubosquet 's initial intent. In the rest of the API, there's a clear boundary between remote interaction (e.g. `(get|save)SolidDataset`) and local manipulation (e.g. `(get|set)Thing`). I think it makes sense to have this boundary for the low-level API too, which is why I'd prefer this not to be `async`. The fetching of the ACR is enforced by the input type of this function, and one can make multiple changes in-memory to a given ACR before saving it back to batch the changes and avoid having multiple round-trips. I would say keeping this `async` takes some control away from users, because it means we may add additional requests here silently, impacting app's performances. We can also always add an async `getResourceAgentAccess` or something, which takes in a resource, silently fetches the ACR, and then calls `getAgentAccess`. I would prefer either both to exist independently, or only a sync `getAgentAccess`, but not just an async `getAgentAccess` which may eventually fetch the ACR.

However, if there are good reasons to think that we may need this to be async one day because of some specificities of the ACP spec for instance, I'm happy to drop these changes and make sure `(get|set)VcAccess` are `async` too.

ThisIsMissEm:
> I think this was @matthieubosquet 's initial intent. In the rest of the API, there's a clear boundary between remote interaction (e.g. `(get|save)SolidDataset`) and local manipulation (e.g. `(get|set)Thing`). I think it makes sense to have this boundary for the low-level API too, which is why I'd prefer this not to be `async`. The fetching of the ACR is enforced by the input type of this function, and one can make multiple changes in-memory to a given ACR before saving it back to batch the changes and avoid having multiple round-trips. I would say keeping this `async` takes some control away from users, because it means we may add additional requests here silently, impacting app's performances. We can also always add an async `getResourceAgentAccess` or something, which takes in a resource, silently fetches the ACR, and then calls `getAgentAccess`. I would prefer either both to exist independently, or only a sync `getAgentAccess`, but not just an async `getAgentAccess` which may eventually fetch the ACR.
> 
> However, if there are good reasons to think that we may need this to be async one day because of some specificities of the ACP spec for instance, I'm happy to drop these changes and make sure `(get|set)VcAccess` are `async` too.

Okay, this all makes good sense to me; I don't know if we'd need async because of something specific to the ACP spec.

matthieubosquet:
I think this should really be async as per @ThisIsMissEm's comment.

First, I think the low level API functions signatures should be aligned with the upper level API functions and in that sense, if we're gonna write to the dataset, this should be done in the low level, allowing the users to use either universal or the ACP function with the same result. I just didn't refactor because of time and ease of testing and because I wanted to have this conversation before doing so.

Second, I think we should pass to all such functions a `Resource` which has a URL, and possibly is augmented with its representation, its info and maybe its ACL. Maybe it isn't even a resource to which the currently authenticated user has access or a `SolidResource`. But I think that would be the right level of abstraction to use such APIs conveniently (we don't want users to have to write too much logic about what's already there). I do think we should let users write the dataset themselves though (not have it saved as part of the universal function).

Third, to get the full representation or meaning of a representation, we might have to fetch more resources (to get the full graph). I don't think that's a problem endemic to ACP, but we will have to deal with that here anyway.

NSeydoux:
Okay, I think this is an important conversation with deep implications on the API design, let's park this PR for now and have it properly.

PR #1386, opened 2021-12-14
Simplify unit tests config

matthieubosquet:
This is a simplification of the test config that also allows to remove the coverage directory from the .gitignore (as test coverage just gets written to the console output via the "text" coverage reporter).

Note: I think we might want to consider removing running lint as part of the test command.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -22,8 +22,10 @@
 module.exports = {
   preset: "ts-jest",
   testEnvironment: "jsdom",
+  testRegex: "/src/.*\\.test\\.ts$",
   clearMocks: true,
   collectCoverage: true,
+  coverageReporters: ["text"],
   coverageThreshold: {
     global: {
       branches: 100,
@@ -32,12 +34,5 @@ module.exports = {
       statements: 100,
     },
   },
-  coveragePathIgnorePatterns: ["/node_modules/", "<rootDir>/dist"],
-  testPathIgnorePatterns: [
-    "/node_modules/",
-    // By default we only run unit tests:
-    "/e2e/node/",
-    "/e2e/browser/",
-  ],
   injectGlobals: false,
 };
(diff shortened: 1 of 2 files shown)


jest.config.js
```diff
@@ -32,12 +34,5 @@ module.exports = {
       statements: 100,
     },
   },
-  coveragePathIgnorePatterns: ["/node_modules/", "<rootDir>/dist"],
-  testPathIgnorePatterns: [
-    "/node_modules/",
-    // By default we only run unit tests:
-    "/e2e/node/",
-    "/e2e/browser/",
```
NSeydoux, line 40: This depends on the `e2e` tests moving out of `src` right ?

jest.config.js
```diff
@@ -32,12 +34,5 @@ module.exports = {
       statements: 100,
     },
   },
-  coveragePathIgnorePatterns: ["/node_modules/", "<rootDir>/dist"],
```
NSeydoux, line 35: I'm not sure why specifying a specific coverage reporter changes the paths that should be ignored measuring coverage. Does this config has an impact in the current setting ? And if so can you elaborate on why it isn't useful anymore ?

jest.config.js
matthieubosquet: By default, jest would pick up anything that has a `.test` suffix, but having the e2e in their own directory makes it easier to have a simple matcher, yes.

jest.config.js
matthieubosquet: The paths `node_modules`is ignored by default, the `dist` folder would not be picked up by unit tests anyway, so that was unnecessary anyway.

As you can see, the test coverage is accurately reported: https://github.com/inrupt/solid-client-js/runs/4521157986?check_suite_focus=true#step:7:70

NSeydoux (approved):

PR #1391, opened 2021-12-21
Remove husky and lint-staged from project

ThisIsMissEm:
We'll be replacing these with an as yet to be determined tool. Standard check-list does not apply for this PR.

If you find after this change that you get errors about husky, run: 
```
git config --unset core.hookspath
```

As Husky has an annoying habit of messing with core git settings.


matthieubosquet (approved):
Happy days ;)

ajacksified (approved):
I'll be back

PR #1399, opened 2022-01-05
Skip end-to-end tests for dependabot PRs

NSeydoux:
Dependabot does not have access to the environment secrets that are needed to run the end-to-end tests (namely, client ID and client secret for authentication).

Merging this should unblock some of the pending dependabot PRs.


matthieubosquet (approved):
Brilliant ;)

PR #1401, opened 2022-01-06
Add convenience function to discover a user's Pods from their profile

NSeydoux:
This adds `getPodUrlAll` and `getPodUrlAllFrom`, two functions aiming at discovering a user's Pods based on their profile. The former function dereferences the WebID, and fetches additional alternative resources if applicable, while the latter takes resources as an input. This design intends at exposing a higher-level function for convenience, while also enabling users to avoid re-fetching profiles if they already have them locally.

# Checklist

### DIFF
--- a/src/profile/webid.ts
+++ b/src/profile/webid.ts
@@ -96,15 +96,17 @@ export async function getProfileAll<
   };
 }
 
-export async function getPodUrlAll<
-  T extends SolidDataset & WithServerResourceInfo
->(
+export async function getPodUrlAll(
   webId: WebId,
   options: Partial<
-    typeof internal_defaultFetchOptions & {
-      webIdProfile: ProfileAll<T>;
-    }
+    typeof internal_defaultFetchOptions
   > = internal_defaultFetchOptions
 ): Promise<UrlString[]> {
-  throw new Error("Unimplemented");
+  throw new Error("unimplemented");
+}
+
+export function getPodUrlAllFrom<
+  T extends SolidDataset & WithServerResourceInfo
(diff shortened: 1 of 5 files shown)


src/profile/webid.ts
```diff
@@ -94,3 +95,60 @@ export async function getProfileAll<
...
+ * should appear in the profile resources.
+ *
+ * @param profiles The profile resources in which the Pods should be discovered
+ * @param webId The WebID of the agent whose Pods should be discovered
+ * @returns An array containing the URLs of all the Pods linked from the agent's
+ * profile resource using the `pim:storage` predicate.
+ */
+export function getPodUrlAllFrom<
+  T extends SolidDataset & WithServerResourceInfo
+>(profiles: ProfileAll<T>, webId: WebId): UrlString[] {
```
matthieubosquet, line 136: Did we not establish that Generics should at least appear twice?
https://github.com/inrupt/solid-client-js/pull/1356
https://microsoft.github.io/TypeScript-New-Handbook/everything/#type-parameters-should-appear-twice
```suggestion
export function getPodUrlAllFrom(
  profiles: ProfileAll<SolidDataset & WithServerResourceInfo>,
  webId: WebId
): UrlString[] {
```

matthieubosquet (approved):
Wondering about the generic type, but otherwise LGTM!

src/profile/webid.ts
NSeydoux: That's a very good point ^^

NSeydoux: Fixed in d4272cf5b710e0c1191383affa6e42fada7349e0

PR #1402, opened 2022-01-06
Update licence header year to 2022

NSeydoux:
Update the license header to 2022. This is done by updating `resource/license-header.js`, and having the linter fix all the in-file headers, so there's little room for human error.


matthieubosquet (approved):
LGTM!

PR #1412, opened 2022-01-11
Switch browser tests from TestCafé to Playright

NSeydoux:
This reapplies https://github.com/inrupt/solid-client-js/pull/1205, with some minor changes. In particular, the GH workflows have been broken down for simplicity: the browser-based tests are no longer part of the `CI` workflow, and run as their own. This enables simplifying/factoring the conditions and matrix.


matthieubosquet:
Can we close https://github.com/inrupt/solid-client-js/pull/1205 if this one supersedes?

playwright.config.ts
```diff
@@ -0,0 +1,78 @@
...
+    },
+    {
+      name: "Chromium",
+      use: {
+        browserName: "chromium",
+        userAgent: `Browser-based solid-client end-to-end tests running ${
+          process.env.CI === "true" ? "in CI" : "locally"
+        }. Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/51.0.2704.103 Safari/537.36`,
+      },
+    },
```
matthieubosquet, line 53:
```suggestion
    },
    {
    name: "WebKit",
    use: {
      browserName: "webkit",
      userAgent: `Browser-based solid-client end-to-end tests running ${
        process.env.CI === "true" ? "in CI" : "locally"
      }. Mozilla/5.0 (iPhone; CPU iPhone OS 13_5_1 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/13.1.1 Mobile/15E148 Safari/604.1`,
    },
```
(see comment below and GitHub action's matrix)

.eslintrc.js
```diff
@@ -30,6 +30,13 @@ module.exports = {
...
     "@typescript-eslint/no-floating-promises": "error",
-    "license-header/header": [process.env.CI ? "error" : "warn", "./resources/license-header.js"],
+    "license-header/header": [
+      process.env.CI ? "error" : "warn",
+      "./resources/license-header.js",
+    ],
   },
+  // ESLint will try to parse these files using TypeScript,
+  // but since they're not part of the project, it will complain
+  // about them not being listed in tsconfig.json:
```
matthieubosquet, line 40: This is because our eslint config is wrong, it should only apply the ts linting when the file extension is correct.
We should discuss at some point fixing our configs.

There is an `overrides` directive in eslint that can be file extensions based and where you can specify `parserOptions` and `plugins`! :)

I'd say let's do this separately though.

.github/workflows/e2e-browser.yml
```diff
@@ -0,0 +1,94 @@
...
+jobs:
+  test:
+    # Running end-to-end tests requires accessing secrets which aren't available to dependabot.
+    # https://github.blog/changelog/2021-02-19-github-actions-workflows-triggered-by-dependabot-prs-will-run-with-read-only-permissions/
+    if: github.actor != 'dependabot[bot]'
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        # Available OS's: https://help.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners
+        os: [ubuntu-20.04, windows-2019, macos-10.15]
```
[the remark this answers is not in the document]
NSeydoux: As it happens, the error the config switch intended to fix has been resolved ever since, so this no longer applies.

playwright.config.ts
NSeydoux: As mentioned above, the error this intended to work around has been resolved.

PR #1424, opened 2022-01-17
Re-enable blocking browser-based tests

NSeydoux:
The browser-based tests started failing at some point for an unknown reason, and they were no longer gating. Now that the tests have moved to Playwright, this tries to make them gating again and see if the failure is still happening.


matthieubosquet (approved):
Looks like it’s passing!
Rebase and merge?

PR #1426, opened 2022-01-18
Use GitHub test environments in e2e tests

matthieubosquet:
Refactor e2e to use GitHub Environment secrets, summary.

The end to end tests:
1. are [run on schedule every 5m](https://github.com/inrupt/solid-client-js/blob/39e6d5d1d0908de914a59f85b892ead6d2d8ac8f/.github/workflows/e2e-node.yml#L10) (no change);
2. are separated from other tasks which means there is [no more complex logic](https://github.com/inrupt/solid-client-js/blob/3aa6992d68a67063f6751b62c38aafdb2dc7568b/.github/workflows/ci.yml#L51) in the workflow, it is instead a [simple and focused task](https://github.com/inrupt/solid-client-js/blob/39e6d5d1d0908de914a59f85b892ead6d2d8ac8f/.github/workflows/e2e-node.yml#L33);
3. their separation from other tasks such as [auditing](https://github.com/inrupt/solid-client-js/blob/39e6d5d1d0908de914a59f85b892ead6d2d8ac8f/.github/workflows/audit.yml#L1) means a [schedule](https://github.com/inrupt/solid-client-js/blob/39e6d5d1d0908de914a59f85b892ead6d2d8ac8f/.github/workflows/audit.yml#L9) and accompanying [alert setup](https://github.com/inrupt/solid-client-js/blob/39e6d5d1d0908de914a59f85b892ead6d2d8ac8f/.github/workflows/audit.yml#L23) can be optimised for timeliness and readability;
4. their separation from other tasks such as [linting](https://github.com/inrupt/solid-client-js/blob/39e6d5d1d0908de914a59f85b892ead6d2d8ac8f/.github/workflows/lint.yml#L1) means workflow failures have clearly identifiable cause;
5. ultimately, more focused workflows mean that tasks can run in parallel and have a more easily understandable outcome;
6. they are also much faster to fail or succeed, for example, a typical full CI run that [could previously span to over 20m](https://github.com/inrupt/solid-client-js/actions/runs/1724000973) would now [typically succeed in under 2m](https://github.com/inrupt/solid-client-js/actions/runs/1724745348), which is great for feedback and not bad for something run on a 5m schedule;
7. using GitHub environment secrets also means alerts pinpoint exactly [which environment](https://github.com/inrupt/solid-client-js/blob/39e6d5d1d0908de914a59f85b892ead6d2d8ac8f/.github/workflows/e2e-node.yml#L45) failed e2e;
8. it is also a separation of concern between [tests](https://github.com/inrupt/solid-client-js/blob/74b47695a01c5e07177c6acae377929dfc4fe955/e2e/node/e2e.test.ts#L61) and the [environments](https://github.com/inrupt/solid-client-js/tree/39e6d5d1d0908de914a59f85b892ead6d2d8ac8f/e2e/env) they're run against and simplifies the [previous code](https://github.com/inrupt/solid-client-js/blob/3aa6992d68a67063f6751b62c38aafdb2dc7568b/e2e/node/e2e.test.ts#L335-L360) by using a standard [environment setup](https://github.com/inrupt/solid-client-js/blob/74b47695a01c5e07177c6acae377929dfc4fe955/e2e/util/setupEnv.ts#L22-L28) in the [jest e2e config](https://github.com/inrupt/solid-client-js/blob/74b47695a01c5e07177c6acae377929dfc4fe955/jest.e2e.config.js#L32);
9. it becomes easy to add a new environment as they contain a [set of standard variables](https://github.com/inrupt/solid-client-js/blob/39e6d5d1d0908de914a59f85b892ead6d2d8ac8f/e2e/util/getTestingEnvironment.ts#L62-L70);
10. those environments and the separation of the tests make it easy to do [feature](https://github.com/inrupt/solid-client-js/blob/74b47695a01c5e07177c6acae377929dfc4fe955/e2e/node/wac.test.ts#L59-L62) or even [environment](https://github.com/inrupt/solid-client-js/blob/74b47695a01c5e07177c6acae377929dfc4fe955/e2e/node/nss-unauth.test.ts#L66-L69) based testing.


e2e/node/wac.test.ts
```diff
@@ -0,0 +1,179 @@
...
+  if (!supportsWac(env.pod)) {
+    return;
+  }
+
+  let options: any;
+  let session: Session;
+  let sessionResource: string;
+
+  beforeEach(async () => {
+    session = await getSession(env);
```
NSeydoux, line 81: Could  `getSession` be renamed `getAuthenticatedSession`for clarity ?

e2e/node/acp-v3.test.ts
```diff
@@ -0,0 +1,271 @@
...
+
+let env: TestingEnvironment;
+
+beforeAll(() => {
+  config({
+    path: __dirname,
+    // Disable warning messages in CI
+    silent: process.env.CI === "true",
+  });
+  env = getTestingEnvironment();
```
NSeydoux, line 59: Looking at `getTestingEnvironment`, whether the env variables or not are defined is not checked, and non-null assertions are used instead. This should be changed to detect when the environment is not properly set.

e2e/node/util/supportsAcp.ts
```diff
@@ -0,0 +1,24 @@
...
+* THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,
+* INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A
+* PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+* HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+* OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+* SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+*/
+
+export function supportsAcp(url: string) {
+  return url.includes("pod.inrupt.com");
```
NSeydoux, line 23: Wouldn't this be better explicitly defined as part of the environment ? E.g. adding `E2E_TEST_ACP` and `E2E_TEST_WAC` to the environments, instead of relying on IRI matching ?

e2e/node/acp-v3.test.ts
```diff
@@ -0,0 +1,271 @@
...
+});
+
+describe(`Authenticated end-to-end ACP V3 tests against environment ${env.environment}`, () => {
+  if (!supportsAcp(env.pod)) {
+    return;
+  }
+
+  let options: any;
```
NSeydoux, line 67:
```suggestion
  let options: { fetch: typeof global.fetch };
```

e2e/node/acp-v3.test.ts
```diff
@@ -0,0 +1,271 @@
...
+  let session: Session;
+  let sessionResource: string;
+
+  beforeEach(async () => {
+    session = await getSession(env);
+    sessionResource = `${env.pod}solid-client-tests/node/acp-v3-test-dataset-${session.info.sessionId}`;
```
NSeydoux, line 73: If possible, could the URL constructor be used here instead of plain string templates ?

e2e/node/acp-v3.test.ts
```diff
@@ -0,0 +1,271 @@
...
+    let publicRule = acp.createRule(policyResourceUrl + "#rule-public");
+    publicRule = acp.setPublic(publicRule);
+
+    let publicReadPolicy = acp.createPolicy(
+      policyResourceUrl + "#policy-publicRead"
+    );
+    // Note: we should think of a better name for "optional", as this isn't really optional.
+    //       At least one "optional" rule should apply, and since this is the only rule for this
+    //       policy, it will in practice be required.
+    publicReadPolicy = acp.addAnyOfRuleUrl(publicReadPolicy, publicRule);
```
NSeydoux, line 96: I'm not sure to what this comment applies.

e2e/node/acp-v3.test.ts
```diff
@@ -0,0 +1,271 @@
...
+    publicReadPolicy = acp.setAllowModes(publicReadPolicy, {
+      read: true,
+      append: false,
+      write: false,
+    });
+
+    let selfRule = acp.createRule(policyResourceUrl + "#rule-self");
+    selfRule = acp.addAgent(selfRule, session.info.webId!);
```
NSeydoux, line 104: Checking that `session.info.webId` is not null could be added to the `beforeEach` block, and the non-null assertion could be replaced with a type assertion `as string`.

e2e/node/acp-v3.test.ts
```diff
@@ -0,0 +1,271 @@
...
+      read: true,
+      append: false,
+      write: false,
+    });
+
+    let policyResource = createSolidDataset();
+    policyResource = setThing(policyResource, publicRule);
+    policyResource = setThing(policyResource, publicReadPolicy);
+    policyResource = setThing(policyResource, selfRule);
+    policyResource = setThing(policyResource, selfWriteNoReadPolicy);
```
NSeydoux, line 129: Should we encourage the pattern
```
const policyResource = [publicRule, publicReadPolicy, selfRule, selfWriteNoReadPolicy].reduce(setThing, createSolidDataset());
```
? It's something I have seen before, and it allows declaring the `policyResource` as a `const`, but I don't know readability-wise how it compares to having multiple assignments.

We could even consider adding a `setThingAll(dataset, things)` function to the API if this becomes recurrent enough.

e2e/node/acp-v3.test.ts
```diff
@@ -0,0 +1,271 @@
...
+    await expect(
+      getSolidDataset(sessionResource, options)
+    ).rejects.toThrow(
+      // Forbidden:
+      // @ts-ignore-next
+      expect.objectContaining({ statusCode: 403 }) as FetchError
+    );
+
+    // Clean up:
+    await deleteSolidDataset(sessionResource, options);
```
NSeydoux, line 184: I think this is redundant with the `afterEach` block.

e2e/node/acp-v3.test.ts
```diff
@@ -0,0 +1,271 @@
...
+      sessionResource + "#policy-selfWriteNoRead",
+      session
+    );
+
+    // Verify that indeed, the current user can no longer read it:
+    await expect(
+      getSolidDataset(sessionResource, options)
+    ).rejects.toThrow(
+      // Forbidden:
+      // @ts-ignore-next
```
NSeydoux, line 179: Can you elaborate on why this is needed ?

e2e/node/acp-v3.test.ts
```diff
@@ -0,0 +1,271 @@
...
+
+  it("can allow public Read access", async () => {
+    // Create a Resource containing Access Policies and Rules:
+    await initialisePolicyResource(sessionResource, session);
+
+    // Verify that we cannot fetch the Resource before adding public Read access
+    // when not logged in (i.e. not passing the session's fetch):
+    await expect(getSolidDataset(sessionResource)).rejects.toThrow(
+      // Unauthorised:
+      // @ts-ignore-next
```
NSeydoux, line 195: ditto

e2e/node/acp-v3.test.ts
```diff
@@ -0,0 +1,271 @@
...
+    // Verify that indeed, an unauthenticated user can now read it:
+    await expect(
+      getSolidDataset(sessionResource)
+    ).resolves.not.toBeNull();
+
+    // Clean up:
+    await deleteSolidDataset(sessionResource, options);
+  });
+
+  it("can set Access from a Resource's ACR", async () => {
```
NSeydoux, line 217: I'm not sure the description matches the actual test.

e2e/node/acp.test.ts
```diff
@@ -142,7 +142,7 @@ describe("An ACP Solid server", () => {
         ? latest_getPublicAccess
         : legacy_getPublicAccess;
     await expect(
-      getSolidDataset(sessionDataset, { fetch: session.fetch })
+      getSolidDataset(sessionDataset, )
```
NSeydoux, line 145:
```suggestion
      getSolidDataset(sessionDataset)
```

e2e/node/unauth.test.ts
```diff
@@ -0,0 +1,287 @@
...
+  deleteFile,
+  createContainerAt,
+  createContainerInContainer,
+  getBoolean,
+  setBoolean,
+} from "../../src/index";
+
+// This block of end-to-end tests should be removed once solid-client-authn-node works against NSS,
+// and the e2e tests have an NSS environment setup.
+describe("End-to-end tests with pre-existing data against resources in an NSS server", () => {
```
NSeydoux, line 66: Seeing what these tests intend to do, wouldn't it be more accurate to name the file `nss-unauth` or something ?

NSeydoux (approved):

e2e/node/wac.test.ts
matthieubosquet: Fixed in https://github.com/inrupt/solid-client-js/commit/7b57210bb326483ec569ba19673520a658d0cd19

e2e/node/acp-v3.test.ts
matthieubosquet: Fixed in https://github.com/inrupt/solid-client-js/commit/7b57210bb326483ec569ba19673520a658d0cd19

e2e/node/util/supportsAcp.ts
matthieubosquet: Fixed in https://github.com/inrupt/solid-client-js/pull/1426/commits/3f12c1a4758d580af685f89dd2c47b56ed52e841.

e2e/node/unauth.test.ts
matthieubosquet: Renamed in https://github.com/inrupt/solid-client-js/pull/1426/commits/956ff574950f6b57711d88249b8d2bcfa7a2276e.

e2e/node/acp-v3.test.ts
matthieubosquet: Fixed in https://github.com/inrupt/solid-client-js/pull/1426/commits/86a099bb7480faf126a5b5ea6e8133e5846f2656.

e2e/node/acp.test.ts
matthieubosquet: That was a typo, the session is supposed to be used here ^^,
Thanks for spotting it!

NSeydoux: Only half-spotted the issue then ^^

e2e/node/acp-v3.test.ts
matthieubosquet: Not sure, I removed it in https://github.com/inrupt/solid-client-js/pull/1426/commits/b63d01405ca610982d8680df69e4b6e3d8341386.

e2e/node/acp-v3.test.ts
matthieubosquet: I don't think it's needed, removed in https://github.com/inrupt/solid-client-js/pull/1426/commits/b63d01405ca610982d8680df69e4b6e3d8341386.

e2e/node/acp-v3.test.ts
matthieubosquet: Removed all redundant cleanup in https://github.com/inrupt/solid-client-js/pull/1426/commits/39b56eba3a4e1260fe98ba5d99d87eba83d4c99a.

e2e/node/acp-v3.test.ts
matthieubosquet: I am not against it, but this is not a re-write, just a separation of ACP V3 tests which will eventually by phased out, I'll make sure I keep it in mind when adding new tests or refactoring the existing ones.

e2e/node/acp-v3.test.ts
matthieubosquet: I am not sure either, this is ACP V3 tests that were written before I joined the team. I'll leave those as is for now.

e2e/node/acp-v3.test.ts
matthieubosquet: Sure, but as for previous 2 comments, these are ACP V3 tests, so I will not refactor them as part of this PR.

e2e/node/acp-v3.test.ts
NSeydoux: Could you add a FIXME note so that we think about revisiting ^^ ?

PR #1230, opened 2021-08-26
Update saveSolidDatasetAt to correct misleading overwrite statement

AJamesPhillips:
Suggested improvement to documentation.

On solidcommunity.net, `saveSolidDatasetAt` allows you to overwrite the data.  On inrupt.com you have to delete the data first (or fetch first -- I have not tested this yet) and save to overwrite the data.

I think the wording could be improved.

PR #1429, opened 2022-01-19
Doc/save solid dataset

NSeydoux:
Supersedes https://github.com/inrupt/solid-client-js/pull/1230 (late after the battle ^^).

The proposed change has been modified, because I'm afraid it describes something which is already described in the common issues page, which risks getting out of sync at some point and confusing users. Linking to that page though is a good idea.


matthieubosquet (approved):
LGTM

PR #1432, opened 2022-01-20
add Code of Conduct

VirginiaBalseiro:
This PR adds a Code of Conduct to `solid-client-js`. It is adapted from [Contributor Covenant](https://www.contributor-covenant.org/).

I am particularly interested in feedback regarding how to further adapt this to include behaviours and standards specific to our community, but of course all feedback is very much welcome.

### DIFF
--- a/CODE-OF-CONDUCT.md
+++ b/CODE-OF-CONDUCT.md
@@ -1,0 +1,132 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+We as members, contributors, and leaders pledge to make participation in our project and community 
+a harassment-free experience for everyone, regardless of age, body
+size, visible or invisible disability, ethnicity, sex characteristics, gender
+identity and expression, level of experience, education, socio-economic status,
+nationality, personal appearance, race, caste, color, religion, geographic location, or sexual
+identity and orientation.
+
+We pledge to act and interact in ways that contribute to an open, welcoming,
+diverse, inclusive, and healthy community.
+
+## Our Standards
+
+Examples of behavior that contributes to a positive environment for our
+community include:
+
+- Demonstrating empathy and kindness toward other people
+- Being respectful of differing opinions, viewpoints, and experiences
+- Giving and gracefully accepting [constructive feedback](https://www.virginiabalseiro.com/blog/feedback.html)
+- Accepting responsibility and apologizing to those affected by our mistakes,
+  and learning from the experience
+- Focusing on what is best not just for us as individuals, but for the overall
+  community
+
+Examples of unacceptable behavior include:
+
+- The use of sexualized language or imagery, and sexual attention or advances of
+  any kind
+- Trolling, insulting or derogatory comments, and personal or political attacks
+- Public or private harassment
+- Publishing others' private information, such as a physical or email address,
+  without their explicit permission
+- Other conduct which could reasonably be considered inappropriate in a
+  professional setting
+
+## Enforcement Responsibilities
+
+Project members are responsible for clarifying and enforcing our standards of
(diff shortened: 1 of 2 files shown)


CODE-OF-CONDUCT.md
```diff
@@ -0,0 +1,132 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+We as members, contributors, and leaders pledge to make participation in our project and community 
+a harassment-free experience for everyone, regardless of age, body
+size, visible or invisible disability, ethnicity, sex characteristics, gender
+identity and expression, level of experience, education, socio-economic status,
+nationality, personal appearance, race, caste, color, religion, geographic location, or sexual
+identity and orientation.
```
NSeydoux, line 10: Should "political orientation" be mentioned in the list as well ?

CODE-OF-CONDUCT.md
```diff
@@ -0,0 +1,132 @@
...
+
+We pledge to act and interact in ways that contribute to an open, welcoming,
+diverse, inclusive, and healthy community.
+
```
NSeydoux, line 14:
```suggestion
As members of the Solid Community, we pledge to follow its [code of conduct](https://github.com/solid/process/blob/main/code-of-conduct.md), which values are aligned with our own.

```
I think it's nice to reference the Solid Community code of conduct, but it may go in the attributions too.

NSeydoux (approved):

CODE-OF-CONDUCT.md
ThisIsMissEm: @NSeydoux usually not, as it's used as an escape hatch by the far-right when they harass people

CODE-OF-CONDUCT.md
ThisIsMissEm: Maybe: 
```suggestion
We pledge to act and interact in ways that contribute to an open, welcoming,
diverse, inclusive, and healthy community. These values align with those of the 
[Solid Community's Code of Conduct]
(https://github.com/solid/process/blob/main/code-of-conduct.md).
```

ThisIsMissEm:
In general we should really try to avoid making per-repo changes to the contributor covenant — it's been very well thought out and battle tested by a lot of teams, projects & organisations, so it's probably pretty spot-on.

CODE-OF-CONDUCT.md
ajacksified: I wonder if the Solid org would be interested in adopting this CoC instead of a handwritten one.

matthieubosquet:
I agree 100% with Emelia, I think the code of conduct should be centralised at the Inrupt level and referenced by our open source projects (potentially together with a central security policy).
Maybe in a `@inrupt/open-source-projects` or simply `@inrupt/open-source` repo?

CODE-OF-CONDUCT.md
NSeydoux: Oh I see. I figured it could be an issue, but my impression was that let's say banning a nazi for being a nazi is not harassement, but I see how that could be instrumentalized.

NSeydoux:
We do have a repo for public-facing documentations (https://github.com/inrupt/public-documentation), that could indeed be a place to put this, and then individual projects READMEs could point there.

VirginiaBalseiro:
> I agree 100% with Emelia, I think the code of conduct should be centralised at the Inrupt level and referenced by our open source projects (potentially together with a central security policy). Maybe in a `@inrupt/open-source-projects` or simply `@inrupt/open-source` repo?

@ajacksified and I discussed this and we agreed that having one file in each repo is more visible/obvious when cloning/forking. We can have a discussion about this but it does make sense to have something in place until we do have that discussion, unless someone feels very strongly about this.

PR #1441, opened 2022-01-26
Deprecate ACP *asMarkdown functions

matthieubosquet:


NSeydoux (approved):
Can you make sure the commit message captures the rationale for the deprecation, and not only the fact that these are deprecated ? Also, this should appear in the CHANGELOG.

PR #1447, opened 2022-02-01
Remove deprecated group functions from universal API

matthieubosquet:
The `getAccessFor`, `getAccessForAll` and `setAccessFor` functions are aliases for managing group-based access.

This should be removed as part of the deprecation of groups in ACP.

### DIFF
--- a/package.json
+++ b/package.json
@@ -143,6 +143,10 @@
     "./formats": {
       "import": "./dist/formats/index.mjs",
       "types": "./dist/formats/index.d.ts"
+    },
+    "./universal": {
+      "import": "./dist/universal/index.mjs",
+      "types": "./dist/universal/index.d.ts"
     }
   },
   "files": [
--- a/src/access/for.ts
+++ b/src/access/for.ts
@@ -64,6 +64,7 @@ export type Actor = "agent" | "group" | "public";
  * @param options Optional parameter `options.fetch`: An alternative `fetch` function to make the HTTP request, compatible with the browser-native [fetch API](https://developer.mozilla.org/docs/Web/API/WindowOrWorkerGlobalScope/fetch#parameters).
  * @returns What access the given Agent or Group has.
  * @since 1.5.0
+ * @deprecated Access Control Policies will no longer support vcard:Group. Use the mechanism-specific access API's if you want to define access for groups of people.
  */
 export async function getAccessFor(
   resourceUrl: UrlString,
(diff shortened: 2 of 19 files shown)


NSeydoux:
Would this be a breaking change for the current "main" ACP entrypoint, or does it only affect `acp_v4` ?

matthieubosquet:
@NSeydoux the `getAccessFor`, `getAccessForAll` and `setAccessFor` functions are available through the universal API.

```typescript
import { getAccessFor, getAccessForAll, setAccessFor } from "@inrupt/solid-client/src/access/universal";
// OR
import { access } from "@inrupt/solid-client";
```

So, once the 1.2 update gets through, the group based permissions aren't a thing anymore. You'd still be able to import them through universal_v1 directly.

matthieubosquet:
@NSeydoux I actually added the deprecation markers in the related module. I think it was just an oversight.

But really this should be removed altogether from the default universal API.

See that commit for deprecation: https://github.com/inrupt/solid-client-js/pull/1447/commits/0535cbcaf8154f4bea41c836c3c6053604a6f58c

ThisIsMissEm (approved):
I think this is good; If I'm understanding correctly universal_v2 isn't really being used by anyone at present, and those methods shouldn't have been part of it's API in the first place as they were deprecated beforehand

ajacksified (changes requested):
As discussed in slack - because it _is_ public we should not break the interface, regardless of whether it was previously deprecated, without a major version change. Major version bumps are required for the removal of deprecated functions.

ThisIsMissEm:
@ajacksified I don't think this does break the public interface, as you currently have to explicitly import from `"@inrupt/solid-client/src/access/universal_v2";` or use `access_v2`, both of which aren't currently documented: https://docs.inrupt.com/developer-tools/api/javascript/solid-client/modules/access_universal.html

Though, switching `access/universal` from `access/universal_v1` to `access/universal_v2` WOULD be a breaking change.

ThisIsMissEm:
Or would we actually need an `access/universal_v3` now, which omits those functions?

ajacksified:
They're not documented but they are exported by both our index file and by direct import. The fact that they're exported was probably a mistake at the time, but because they are, regardless of the status of documentation, they are part of the public interface and should not change.

access/universal_v3 makes sense - that's the conclusion @matthieubosquet and I came to in Slack.

ajacksified (approved):

PR #1465, opened 2022-02-14
Update TypeScript

matthieubosquet:
The TypeScript version used on this package was wildly out of date ^^,

Just a little version bump.


NSeydoux (approved):
I'm curious why dependabot didn't say anything about it ?

PR #1467, opened 2022-02-14
Fix the universal module doc generation

matthieubosquet:
The docs generation doesn't support having multiple comments at the begining of a file.


kay-kim (approved):
👍

NSeydoux (approved):

PR #1472, opened 2022-02-16
Fix/getpodurlall 401

NSeydoux:
Instead of throwing an error when looking up an alternative profile fails, getProfileAll now catches the thrown errors and returns the successully fetched profiles (if any). In complement to this change, a function has been added to the public API to retrieve the alternative profiles IRIs from a given WebID profiles. This makes it possible for the client to see that they don't have access to some alternative profile, and to request access to said profile.


matthieubosquet (approved):
LGTM!

PR #1477, opened 2022-02-17
Auto-merge Dependabot

matthieubosquet:
This enables Dependabot auto-merging for anything that is a minor or patch dependency update.

### DIFF
--- a/.github/workflows/dependabot.yml
+++ b/.github/workflows/dependabot.yml
@@ -1,0 +1,27 @@
+name: Dependabot
+
+on:
+  pull_request:
+
+permissions:
+  pull-requests: write
+  contents: write
+
+jobs:
+  dependabot:
+    runs-on: ubuntu-latest
+    if: ${{ github.actor == 'dependabot[bot]' }}
+    steps:
+      - uses: dependabot/fetch-metadata@v1.1.1
+        id: metadata
+        with:
+          github-token: "${{ secrets.GITHUB_TOKEN }}"
+      - run: gh pr review --approve "$PR_URL"
+        env:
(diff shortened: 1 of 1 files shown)


.github/workflows/dependabot.yml
```diff
@@ -0,0 +1,27 @@
...
+jobs:
+  dependabot:
+    runs-on: ubuntu-latest
+    if: ${{ github.actor == 'dependabot[bot]' }}
+    steps:
+      - uses: dependabot/fetch-metadata@v1.1.1
+        id: metadata
+        with:
+          github-token: "${{ secrets.GITHUB_TOKEN }}"
+      - run: gh pr review --approve "$PR_URL"
```
NSeydoux, line 19: Just for my understanding, is this based on the GitHub CLI client ?

matthieubosquet: yes

matthieubosquet: See also: https://docs.github.com/en/actions/using-workflows/using-github-cli-in-workflows

NSeydoux: And is this compatible with the codeowners feature ? Which user approves the PR ? I assume that would be the one who generated the token, is that correct ? I don't see any way of setting this up with a shared account, because the shared account would have admin access to our repos (being a code owner), which would be a pretty bad design.

matthieubosquet: Good question, the user approving the request would be the github-actions bot. I'm gonna test on my repo what happens when a codeowner review is required. 🤔
Not sure the best way is with that...

PR #1483, opened 2022-02-21
Add turtle serialisation of Solid Datasets

matthieubosquet:


src/formats/solidDatasetAsTurtle.ts
```diff
@@ -0,0 +1,60 @@
...
+  options?: {
+    prefixes?: Record<string, string>,
+    thing?: string,
+  }
+): Promise<string> {
+  const { prefixes = defaultPrefixes, thing } = { ...options };
+  const writer = new Writer({ format: "application/Turtle", prefixes });
+  const subject = thing ? new NamedNode(thing) : undefined;
+
+  for (const quad of toRdfJsDataset(dataset).match(subject)) {
```
NSeydoux, line 47:
```suggestion
  // If the subject is undefined, all the triples match.
  for (const quad of toRdfJsDataset(dataset).match(subject)) {
```

NSeydoux: Just because I had to double check ^^

src/formats/solidDatasetAsTurtle.ts
```diff
@@ -0,0 +1,60 @@
...
+ */
+export async function solidDatasetAsTurtle(
+  dataset: SolidDataset,
+  options?: {
+    prefixes?: Record<string, string>,
+    thing?: string,
+  }
+): Promise<string> {
+  const { prefixes = defaultPrefixes, thing } = { ...options };
+  const writer = new Writer({ format: "application/Turtle", prefixes });
```
NSeydoux, line 44:
```suggestion
  const writer = new Writer({ format: "application/turtle", prefixes });
```

src/formats/solidDatasetAsTurtle.ts
```diff
@@ -0,0 +1,60 @@
...
+  const subject = thing ? new NamedNode(thing) : undefined;
+
+  for (const quad of toRdfJsDataset(dataset).match(subject)) {
+    writer.addQuad(quad);
+  }
+
+  return new Promise<string>((resolve, reject) => {
+    writer.end((error, result) => {
+      /* istanbul ignore next */
```
NSeydoux, line 53: Could you add a short comment explaining why this cannot be covered ? I assume it's related to N3's permissiveness ^^

src/formats/solidDatasetAsTurtle.ts
```diff
@@ -0,0 +1,60 @@
...
+import { prefixes as defaultPrefixes } from "./prefixes";
+import { NamedNode, Writer } from "n3";
+import { toRdfJsDataset } from "../rdfjs";
+
+/**
+ * A function to serialise a Solid Dataset as Turtle
+ * 
+ * @param dataset The Dataset to serialize as Turtle
+ * @param options.prefixes The Prefixes to use for Turtle serialisation (defaulting to a set of well known prefixes)
+ * @param options.thing Restricts serialisation to the part of a dataset related to the thing
```
NSeydoux, line 32: This is going to look like two different params in the docs. Maybe use a bullet list but a single `@param` tag ?

NSeydoux (approved):

src/formats/solidDatasetAsTurtle.ts
```diff
@@ -0,0 +1,60 @@
...
+import type { SolidDataset } from "../interfaces";
+import { prefixes as defaultPrefixes } from "./prefixes";
+import { NamedNode, Writer } from "n3";
+import { toRdfJsDataset } from "../rdfjs";
+
+/**
+ * A function to serialise a Solid Dataset as Turtle
+ * 
+ * @param dataset The Dataset to serialize as Turtle
+ * @param options.prefixes The Prefixes to use for Turtle serialisation (defaulting to a set of well known prefixes)
```
kay-kim, line 31: Quick question -- can we link to the well-known prefixes we're using or something?

src/formats/solidDatasetAsTurtle.ts
matthieubosquet: The related comment is in a skipped test in the test file.

src/formats/solidDatasetAsTurtle.ts
matthieubosquet: q1: The prefixes are linkable, but it's more of a TypeDoc question I guess? It's just a constant.

q2: I'd say number two, but no strong opinion.

src/formats/solidDatasetAsTurtle.ts
matthieubosquet: I still can't really find the relevant bits in [TypeDoc's documentation](https://typedoc.org/guides/doccomments/#%40param-%3Cparam-name%3E) (how ironic ^^,)... but it looks like this pattern is what's expected if you want to explain options parameters without creating an interface: https://github.com/TypeStrong/typedoc/issues/497#issuecomment-315555494

PR #1490, opened 2022-02-24
Fix e2e browser tests

ThisIsMissEm:
Previously we were trying to start the webserver from playwright's globalSetup, which could have been causing our tests to hang. Now playwright will attempt to start the server, and timeout if that fails, rather than leaving the test in a running state.

Further fixes include:
- prevent dotenv being loading in CI environments, and being loaded multiple times in local environments (by moving it to globalSetup)
- use npm link for testing against the current library build in CI
- added a maximum timeout for e2e tests of 20 minutes (down from the default 360 minutes or 6 hours previously used) in CI
- use parcel from .codesandbox/sandbox instead of npx when starting the e2e test server

This PR fixes #<issue ID>.


.github/workflows/e2e-node.yml
```diff
@@ -22,14 +22,27 @@ jobs:
...
+          - "NSS"
     steps:
       - uses: actions/checkout@v2
+
       - name: Runs on ${{ matrix.os }} with Node.js ${{ matrix.node-version }} against ${{ matrix.environment-name }}
-        uses: actions/setup-node@v2.5.1
+        uses: actions/setup-node@v2
+        id: setup-node
         with:
           node-version: ${{ matrix.node-version }}
```
matthieubosquet, line 37:
```suggestion
          node-version: "16"
```

NSeydoux:
Fixtures could be used to set up test resources as it is done in the solid-client-authn browser-based tests, but I'd rather add this in a separate PR, this one has already grown in scope quite much.

NSeydoux (approved):

.github/workflows/e2e-browser.yml
```diff
@@ -5,81 +5,108 @@ on:
...
-        continue-on-error: true
-        if: github.event_name != 'schedule'
-        with:
-          name: dist
-          path: dist
+          path: e2e/browser/test/test-results/
+        # Ensure test recordings are always uploaded if the exist:
+        if: ${{ always() }}
+
       - name: Send a notification that the test has failed
```
ThisIsMissEm, line 110: You can remove this as this no longer runs on a schedule

PR #1491, opened 2022-02-24
Improve CI configuration

ThisIsMissEm:
# New feature description

This will be an attempt at improving our overall build pipeline, but for now it just speeds up our unit tests by running a build once and testing in each environment (jest doesn't actually depend on the build output)

# Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,12 +6,12 @@ env:
   CI: true
 jobs:
   build:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
         with:
-          node-version: 16
+          node-version: "16"
           cache: "npm"
           cache-dependency-path: "**/package-lock.json"
       - run: npm ci --ignore-scripts
@@ -22,8 +22,8 @@ jobs:
     strategy:
       matrix:
         # Available OS's: https://help.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners
-        os: [ubuntu-20.04, windows-2019, macos-10.15]
-        node-version: [16.x, 14.x, 12.x]
+        os: [ubuntu-latest, windows-latest, macos-latest]
+        node-version: ["16", "14", "12"]
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,9 +4,20 @@ on: [push]
 
 env:
   CI: true
-  DEFAULT_NODE_VERSION: 16.x
 jobs:
   build:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 16
+          cache: "npm"
+          cache-dependency-path: "**/package-lock.json"
+      - run: npm ci --ignore-scripts
+      - run: npm run build
+  test:
+    needs: build
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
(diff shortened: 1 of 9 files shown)


.github/workflows/ci.yml
```diff
@@ -4,42 +4,34 @@ on: [push]
...
-          key: ${{ runner.os }}-node${{ runner.node-version }}-build-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
-      # Retry `npm ci` if it fails (No, GitHub Actions at this point can retry)
-      - run: npm ci || npm ci || npm ci
-      - run: npm run build
+          cache: "npm"
+          cache-dependency-path: "**/package-lock.json"
+      - run: npm ci --ignore-scripts
       - run: npm test
-      - run: npx package-check
-      - name: Archive code coverage results
```
ThisIsMissEm, line 34: We don't actually generate a code coverage artifact as a file anymore, as of #1386 — instead the coverage is just reported to the console output.

.github/workflows/ci.yml
```diff
@@ -4,42 +4,34 @@ on: [push]
...
-          path: node_modules
-          key: ${{ runner.os }}-node${{ runner.node-version }}-build-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
-      # Retry `npm ci` if it fails (No, GitHub Actions at this point can retry)
-      - run: npm ci || npm ci || npm ci
-      - run: npm run build
+          cache: "npm"
+          cache-dependency-path: "**/package-lock.json"
+      - run: npm ci --ignore-scripts
       - run: npm test
-      - run: npx package-check
```
ThisIsMissEm, line 33: I have also dropped this, as it's really a linting thing, not an CI thing.

.github/workflows/ci.yml
```diff
@@ -4,42 +4,34 @@ on: [push]
...
   build:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v2
+        with:
+          node-version: 16
+          cache: "npm"
+          cache-dependency-path: "**/package-lock.json"
```
ThisIsMissEm, line 17: Turns out this doesn't actually work correctly in v2 from what I can tell

.github/workflows/ci.yml
```diff
@@ -4,42 +4,34 @@ on: [push]
...
-        env:
-          cache-name: cache-node-modules
-        with:
-          path: node_modules
-          key: ${{ runner.os }}-node${{ runner.node-version }}-build-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
-      # Retry `npm ci` if it fails (No, GitHub Actions at this point can retry)
-      - run: npm ci || npm ci || npm ci
-      - run: npm run build
+          cache: "npm"
+          cache-dependency-path: "**/package-lock.json"
```
matthieubosquet, line 35:
```suggestion
```

.github/workflows/ci.yml
```diff
@@ -4,42 +4,34 @@ on: [push]
...
 env:
   CI: true
-  DEFAULT_NODE_VERSION: 16.x
 jobs:
   build:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v2
```
matthieubosquet, line 13:
```suggestion
      - uses: actions/setup-node@v2
```

ThisIsMissEm: Why this change? Does actions/setup-node automatically print out the version number? also, I think there's now setup-node@v3

matthieubosquet: There is indeed a v3 and sure, we should use that.

I think we should in most cases (not sure if many justify the opposite) leave the standard action name as default instead of customising the messages, because:

- Yes the node version gets printed as well as all the parameters to the action
- More code/prose means more maintenance 
-  The denormalisation is actually bad because I want to see at a glance which standard action is being used
- In this specific case, why do we even improvise to say “Use” when the action says “Setup”?


In any case, I think we’re better off without this kind of prose and denormalisation.

.github/workflows/ci.yml
ThisIsMissEm: I'm going to keep this for now, as that way we're consistent across all our repositories.

matthieubosquet (approved):

PR #1519, opened 2022-03-16
Export Actor type in universal access v1

ianconsolata:
Export the `Actor` type as part of the universal `access` export.


NSeydoux (approved):
Thanks for contributing and fixing something that had been overlooked !

NSeydoux:
Re-applied in https://github.com/inrupt/solid-client-js/pull/1522 because of CI constraints

PR #1524, opened 2022-03-18
Documentation Improvements

ThisIsMissEm:
This PR started out as fixing just a few documentation issues where I'd accidentally broken some configuration (I only noticed this whilst doing access-grants), and then ended up expanding the documentation to include information from the  readme and have a section on importing the universal access APIs.


ThisIsMissEm:
You can see the `universalAccess` changes here: https://solid-client-js-git-fix-minor-docs-issues-inrupt.vercel.app/modules/universalAccess.html

The readme changes can be seen here: https://solid-client-js-git-fix-minor-docs-issues-inrupt.vercel.app/index.html

docs/api/source/index.rst
```diff
@@ -1,12 +1,79 @@
 :orphan:
 
-================
-solid-client API
-================
+@inrupt/solid-client API Documentation
+======================================
 
+`@inrupt/solid-client <https://docs.inrupt.com/developer-tools/javascript/client-libraries/>`__
```
ThisIsMissEm, line 6: We may wish to change this URL to point to either https://www.npmjs.com/package/@inrupt/solid-client or the GitHub repo.

matthieubosquet: I think we should point to the npm package.
Also, we may want to standardise our READMEs a bit more.
Some simple improvements include badges pointing to the npm package and showing tests passing.

src/universal/index.ts
```diff
@@ -19,6 +19,48 @@
...
+ *
+ * This API is still experimental, and subject to change. It builds on top of
+ * both ACP and ACL, aiming at being adaptable to any Access Control system that
+ * may be implemented in Solid. That is why it is purely Resource-centric: the
+ * library discovers metadata associated with the Resource itself, and calls the
+ * appropriate underlying API to deal with the Access Control in place for the
+ * target Resource.
+ *
+ * As it is still under development, the following export is *only* intended for
+ * experimentation by early adopters, and is not recommended yet for production
```
matthieubosquet, line 34:
```suggestion
 * experimentation by early adopters, and is not recommended for production
```

src/universal/index.ts
```diff
@@ -19,6 +19,48 @@
...
+/**
+ * :::{admonition} Experimental API
+ * :class: important
+ *
+ * This API is still experimental, and subject to change. It builds on top of
+ * both ACP and ACL, aiming at being adaptable to any Access Control system that
+ * may be implemented in Solid. That is why it is purely Resource-centric: the
+ * library discovers metadata associated with the Resource itself, and calls the
+ * appropriate underlying API to deal with the Access Control in place for the
+ * target Resource.
```
matthieubosquet, line 31:
```suggestion
 * This API is still experimental, and subject to change. It builds on top of
 * both ACP and WAC, aiming at being adaptable to any Access Control system that
 * may be implemented in Solid. That is why it is purely Resource-centric: the
 * library discovers metadata associated with the Resource itself, and calls the
 * appropriate underlying API to deal with the Access Control in place for the
 * target Resource.
```

tsconfig.json
```diff
@@ -1,18 +1,21 @@
 {
   "compilerOptions": {
     /* Basic Options */
-    "target": "es2017",                          /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017','ES2018' or 'ESNEXT'. */
-    "module": "commonjs",                     /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', or 'ESNext'. */
-    "lib": ["es2015", "dom"],                 /* Specify library files to be included in the compilation. */
+    "target": "es2017" /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017','ES2018' or 'ESNEXT'. */,
```
matthieubosquet, line 4: We really should just get rid of all those comments in code.
There is documentation for tsconfig.

We should not have all those comments (including commented out values) but rather just the actual config.

Don't you think?

matthieubosquet (approved):
Two minor comments and something more general about our projects.

But all looks good to me, a serious improvement. Thanks @ThisIsMissEm !

src/universal/index.ts
```diff
@@ -19,6 +19,48 @@
...
+ * ```typescript
+ * import * as universalAccess from "@inrupt/solid-client/universal";
+ * ```
+ *
+ * If you're using Typescript, and receive errors about type definitions not
+ * being found, please see this
+ * [documentation](https://www.typescriptlang.org/docs/handbook/esm-node.html)
+ *
+ * @packageDocumentation
+ * @module universalAccess
```
matthieubosquet, line 61: Nice trick :)

src/universal/index.ts
ThisIsMissEm: Heads up: this comes from https://github.com/inrupt/solid-client-js/blob/4badfc72b390e35e06454c6e550bc67c79f8c491/src/index.ts#L371-L389

tsconfig.json
ThisIsMissEm: I don't mind either way, for now, this is just tackling the formatting.

src/universal/index.ts
ThisIsMissEm: We may later wanna rename `src/universal/index` to something different, but let's save that for another day

docs/api/source/index.rst
ThisIsMissEm: I've pointed to the npm package, but the badges and standardisation is totally out of scope here.

PR #1527, opened 2022-03-18
Make unauthenticated fetches for WebID

NSeydoux:
The WebID is not necessarily the IRI of a Solid Resource, and as such it isn't correct to make an authenticated request to it. `getPodUrlAll` and `getProfileAll` now both use an explicitly unauthenticated fetch to initially get the WebID profile, and only then either use the provided authenticated fetch or use our default session heuristics.

### DIFF
--- a/src/profile/webid.test.ts
+++ b/src/profile/webid.test.ts
@@ -192,19 +192,24 @@ describe("getProfileAll", () => {
     await getProfileAll(MOCK_WEBID);
     // The embedded fetch should have been used to fetch the alt profile.
     expect(mockedAuthFetcher.fetch).toHaveBeenCalledTimes(1);
-    expect(mockedAuthFetcher.fetch).toHaveBeenCalledWith("https://some.profile", expect.anything());
+    expect(mockedAuthFetcher.fetch).toHaveBeenCalledWith(
+      "https://some.profile",
+      expect.anything()
+    );
     // The unauthenticated fetch should have been used to fetch the WebID.
     expect(mockedUnauthFetch.fetch).toHaveBeenCalledTimes(1);
-    expect(mockedUnauthFetch.fetch).toHaveBeenCalledWith(MOCK_WEBID, expect.anything());
+    expect(mockedUnauthFetch.fetch).toHaveBeenCalledWith(
+      MOCK_WEBID,
+      expect.anything()
+    );
   });
 
   it("uses the provided fetch to fetch alt profiles, but not the WebID", async () => {
-    
     // Mock the alt profile authenticated fetch
     const mockedAuthFetcher = jest.fn(fetch) as jest.Mock<
-        ReturnType<typeof window.fetch>,
-        [RequestInfo, RequestInit?]
-      >;
+      ReturnType<typeof window.fetch>,
+      [RequestInfo, RequestInit?]
+    >;
     mockedAuthFetcher.mockResolvedValueOnce(
       new Response(await triplesToTurtle(toRdfJsQuads(MOCK_PROFILE)), {
         headers: {
@@ -498,9 +503,9 @@ describe("getPodUrlAll", () => {
 
   it("uses the provided fetch to fetch alt profiles, but not the WebID", async () => {
     const mockedAuthFetch = jest.fn(fetch) as jest.Mock<
-        ReturnType<typeof window.fetch>,
-        [RequestInfo, RequestInit?]
-      >;
+      ReturnType<typeof window.fetch>,
+      [RequestInfo, RequestInit?]
+    >;
 
     mockedAuthFetch.mockResolvedValueOnce(
       new Response(await triplesToTurtle(toRdfJsQuads(MOCK_PROFILE)), {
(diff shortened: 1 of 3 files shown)


ThisIsMissEm (commented):
I'm gonna need to review this more on Monday, but the tests look maybe not correct at a first pass, but I could be misreading (looking on mobile)

src/profile/webid.ts
```diff
@@ -91,21 +93,20 @@ export async function getProfileAll<
...
-
+  const authFetch = options?.fetch ?? defaultFetch;
+  const webIdProfile =
+    options?.webIdProfile ??
+    ((await getSolidDataset(webId, { fetch: unauthenticatedFetch })) as T);
   const altProfileAll = (
     await Promise.allSettled(
       getAltProfileUrlAllFrom(webId, webIdProfile).map((uniqueProfileIri) =>
-        getSolidDataset(uniqueProfileIri, { fetch })
+        getSolidDataset(uniqueProfileIri, { fetch: authFetch })
```
matthieubosquet, line 109: This looks good.

src/profile/webid.test.ts
```diff
@@ -459,24 +565,30 @@ describe("getPodUrlAll", () => {
...
+    );
+
+    const mockedFetch = jest.fn(fetch).mockResolvedValueOnce(
+      new Response(await triplesToTurtle(toRdfJsQuads(altProfileAll[0])), {
+        headers: {
+          "Content-Type": "text/turtle",
+        },
+        url: "https://some.alt-profile",
+      } as ResponseInit)
+    );
```
matthieubosquet, line 591: Maybe in the expect section, you could add something like:

```js
expect(mockedFetch).toHaveBeenCalledTimes(1);
expect(mockedUnauthFetch).toHaveBeenCalledTimes(1);
```

Just to be more explicit about testing what was called?

matthieubosquet (approved):
LGTM.

Small suggestion about the test.

src/profile/webid.test.ts
```diff
@@ -146,22 +157,62 @@ describe("getProfileAll", () => {
...
+        headers: {
+          "Content-Type": "text/turtle",
+        },
+        url: MOCK_WEBID,
+      } as ResponseInit)
+    );
+
     await getProfileAll(MOCK_WEBID);
+    // The embedded fetch should have been used.
     expect(mockedFetcher.fetch).toHaveBeenCalled();
```
ThisIsMissEm, line 194: I think here we actually need to test that we're using an unauthenticated fetcher for the first call, and then an authenticated fetcher for the rest, and actually assert that, or at least check that `mockedFetcher.fetch` has been called N number of times (2 if I'm reading this right)

src/profile/webid.test.ts
```diff
@@ -146,22 +157,62 @@ describe("getProfileAll", () => {
...
         },
-      })
+        url: "https://some.profile",
+      } as ResponseInit)
     );
+    const mockedFetch = jest.fn(fetch);
     await getProfileAll(MOCK_WEBID, { fetch: mockedFetch });
-    expect(mockedFetch).toHaveBeenCalled();
+    expect(mockedFetch).not.toHaveBeenCalled();
+    expect(mockedFetcher.fetch).toHaveBeenCalled();
```
ThisIsMissEm, line 215: Ditto here, `mockedFetch` should've been called once, the built-in fetch should've been called once too, and we can assert the URLs that are being fetched align with our expectations.

src/profile/webid.ts
```diff
@@ -91,21 +93,20 @@ export async function getProfileAll<
...
 ): Promise<ProfileAll<T>> {
-  const {
-    fetch,
-    webIdProfile = (await getSolidDataset(webId, { fetch })) as T,
-  } = options;
-
+  const authFetch = options?.fetch ?? defaultFetch;
+  const webIdProfile =
+    options?.webIdProfile ??
+    ((await getSolidDataset(webId, { fetch: unauthenticatedFetch })) as T);
```
ThisIsMissEm, line 105: Aside, it might be nice to change `getSolidDataset` to accept `T` as a generic, such that we could just do:
```suggestion
    (await getSolidDataset<T>(webId, { fetch: unauthenticatedFetch }));
```

ThisIsMissEm (commented):
The tests do read a bit strange to me, and aren't particularly clear, I think we can definitely add more clarity to them, whether now or in future work.

src/profile/webid.ts
NSeydoux: Actually, I'd argue that this is an issue with the function signature rather than `getSolidDataset`: in the case when `getSolidDataset` is needed to fetch the WebID, the generic type is unbound, and the return type should be `SolidDataset & WithServerResourceInfo`, which is the return type of `getSolidDataset` in the first place. I'll see if this can be cleaned up with a little overload.

src/profile/webid.test.ts
NSeydoux: Should be fixed in fe6c5ea8f2b4419726d34b731398ab13e40a91cf

src/profile/webid.ts
NSeydoux: That was easy enough: https://github.com/inrupt/solid-client-js/pull/1530

src/profile/webid.test.ts
```diff
@@ -371,28 +467,118 @@ describe("getPodUrlAll", () => {
...
+      .addIri(foaf.isPrimaryTopicOf, "https://some.profile")
+      .build();
+
+    const webIdProfile = setThing(
+      mockSolidDatasetFrom(MOCK_WEBID),
+      profileContent
+    );
+
+    // The WebID is explicitly fetched using the unauthenticated fetch.
+    const mockedUnauthFetcher = jest.requireMock("cross-fetch") as {
```
ThisIsMissEm, line 485: Can these be setup in a beforeEach?

ThisIsMissEm (approved):
Looking good, left a few small comments of things you might want to change, but otherwise we're good

PR #1530, opened 2022-03-21
Remove unnecessary type assertion in getProfileAll

NSeydoux:
This introduces an overload to cleanly define the return type of getProfileAll, namely it the WebID profile is NOT provided, the return type includes that of `getSolidDataset`, and if it IS provided then the return type includes the generics instanciated by the provided WebID profile. Making this distinction allows to remove a type assertion from the implementation.

### DIFF
--- a/src/profile/webid.ts
+++ b/src/profile/webid.ts
@@ -89,20 +89,36 @@ export function getAltProfileUrlAllFrom(
  *  If none are found, the WebID profile document itself is returned.
  * @since 1.16.0
  */
+ export async function getProfileAll<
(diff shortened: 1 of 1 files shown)


ThisIsMissEm (approved):
LGTM, but do double check how this looks in the documentation, as to whether it's clear there.

NSeydoux:
Not too bad actually: https://solid-client-js-ni3xdinux-inrupt.vercel.app/modules/profile_webid.html ! I just removed the type parameter from the overload where it isn't necessary.

PR #1543, opened 2022-03-25
Refactor end-to-end test setup and teardown

NSeydoux:
This is a small refactoring of the end-to-end tests setup and teardown: 
- The `fetch` function used thorough the tests is added a custom User-Agent so that the requests issued by the tests are easier to track through the logs.
- Resources IRIs are no longer manually built. Instead, the test container and resource are created by the server, which indicates their IRI to the client. The client still gets to suggest a slug they want in the resource IRI. Note that if this setup happens in CI, the slug is not specified to avoid name collision: a totally random name will be provided instead. Running locally, the slug is sent to the server to make it easier to manually inspect the results on the target Pod if needed.
- Setting up and tearing down test is now done by a shared function.

The `e2e.test.ts` file has been renamed to `resource.test.ts` to reflect more accurately what it tests, and the ACP-related test in this file has been removed: it was redundant with `acp.test.ts`, and also it wasn't making a distinction between legacy and latest universal modules, which resulted in bad IRIs being constructed by the legacy module against the latest authorization server.


e2e/node/acp-v3.test.ts
```diff
@@ -48,6 +48,9 @@ import {
   TestingEnvironment,
 } from "../util/getTestingEnvironment";
 import { getAuthenticatedSession } from "../util/getAuthenticatedSession";
+import { setupTestResources, teardownTestResources } from "./test-setup";
```
ThisIsMissEm, line 51: Can this be done via Jest?

e2e/node/resource.test.ts
```diff
@@ -137,12 +131,12 @@ describe("Authenticated end-to-end", () => {
   it("can create, delete, and differentiate between RDF and non-RDF Resources", async () => {
     const fileUrl = `${sessionResource}.txt`;
```
matthieubosquet, line 132: Do we actually need this?

Or is the "overwriteFile" function name misleading?

I'm wondering if we could just overwrite the resource (even though it's a solidDataset) as a text file.

e2e/node/test-setup.ts
```diff
@@ -0,0 +1,72 @@
...
+}
+
+export async function teardownTestResources(
+  session: Session,
+  containerUrl: string,
+  resourceUrl: string,
+  userAgentFetch: typeof fetch
+) {
+  await deleteSolidDataset(resourceUrl, { fetch: userAgentFetch });
+  await deleteSolidDataset(containerUrl, { fetch: userAgentFetch });
```
matthieubosquet, line 70: We should really have a deleteSolidResource function that works with anything (it doesn't matter if you delete a resource/container/file in this case).

And we could maybe provide a utility function to delete Solid Containers that recursively deletes everything they contain to allow that operation.

matthieubosquet (approved):
It prompted a few reflections that are not directly linked to the PR, sorry ^^,

On the whole it's a nice refactor :)

e2e/node/acp.test.ts
```diff
@@ -62,28 +70,41 @@ describe("An ACP Solid server", () => {
...
   beforeEach(async () => {
     session = await getAuthenticatedSession(env);
-    sessionResource = `${env.pod}${sessionResourcePrefix}${session.info.sessionId}`;
-    options = { fetch: session.fetch };
-    await saveSolidDatasetAt(sessionResource, createSolidDataset(), options);
+    const { containerUrl, resourceUrl, fetchWithAgent } =
+      await setupTestResources(session, TEST_SLUG, env.pod);
+    sessionResource = resourceUrl;
+    sessionContainer = containerUrl;
+    options = { fetch: fetchWithAgent };
```
ThisIsMissEm, line 81: might be better to call this `fetchOptions` as that's what the type is called, might be clearer than just `options`

ThisIsMissEm (commented):
Couple of comments, but looking good.

e2e/node/acp-v3.test.ts
NSeydoux: I'm not sure what you mean: the actual setup/teardown is done via jest's `beforeEach` and `afterEach`, the `test-setup` module just exports convenience functions to create resources and clear them up. It's just that it is shared across multiple test suites.

ThisIsMissEm: Perhaps we'd be better calling it `test-helpers` instead of `test-setup` which is commonly used to refer to the jest setup?

e2e/node/acp-v3.test.ts
```diff
@@ -60,15 +63,24 @@ describe("Authenticated end-to-end ACP V3", () => {
   let options: { fetch: typeof global.fetch };
   let session: Session;
   let sessionResource: string;
+  let sessionContainer: string;
 
   beforeEach(async () => {
     session = await getAuthenticatedSession(env);
-    sessionResource = `${env.pod}${sessionResourcePrefix}${session.info.sessionId}`;
-    options = { fetch: session.fetch };
+    const { containerUrl, resourceUrl, fetchWithAgent } =
```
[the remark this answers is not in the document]
NSeydoux: 8cbd5790aec49131d6305f29bf9d8ba13b6a7668

PR #1548, opened 2022-03-28
Run scheduled E2E tests only for the main branch

matthieubosquet:
This restricts the branch for scheduled workflow runs to the main branch.


matthieubosquet:
This fix is not required, spotting of scheduled workflow runs was a mistake (not sure where a misleading run was spotted).

See the GitHub docs:
> Scheduled workflows run on the latest commit on the default or base branch.
> -- https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onschedule

PR #1564, opened 2022-04-06
Export acp_ess_ versions

matthieubosquet:
This PR creates default exports for acp_ess version 1 and 2.

It allows having a non-breaking release of solid-client compatible with the low level ACP API in both ESS 1.x and ESS 2.x series.


ajacksified (approved):

src/acp/ess1.ts
```diff
@@ -0,0 +1,156 @@
...
+ * Thus, the following export is *only* intended for experimentation by early adopters, and is not
+ * recommended for production applications. Because of this, all ACP-related API's are exported on a
+ * single object, which does not facilitate tree-shaking: if you use one ACP-related API, all of
+ * them will be included in your bundle.
+ *
+ * For more information see: [Tutorial: Managing
+ * Access](https://docs.inrupt.com/developer-tools/javascript/client-libraries/tutorial/manage-access/)
+ * :::
+ *
+ * This module can be imported as an object from the main package, which results
```
kay-kim, line 40: Is this to support ess 1.1 and earlier?  If so, probably should mention it?  otherwise, it's just acp_ess_1 and acp_ess_2 and people will have to guess at which. And maybe state this replaces acp_v3 or whatever it used to be (for people using PodSpaces and who won't care about ESS versions directly)?

src/acp/ess2.ts
```diff
@@ -0,0 +1,146 @@
...
+/**
+ * :::{admonition} Experimental API
+ * :class: important
+ *
+ * The Access Control Policies proposal has not yet been reviewed for inclusion in the Solid spec.
+ * To enable early experimentation, solid-client exposes a low-level API. However, this API can and
+ * will include breaking changes in non-major releases. Additionally, for most applications, a
+ * higher-level API that is planned will be more applicable.
+ *
+ * Thus, the following export is *only* intended for experimentation by early adopters, and is not
```
kay-kim, line 31: same comment about just a sentence about what acp_ess_2 is for, to support ESS v2.

Is it possible to have acp_ess_2 have the descriptions in its page rather than as references to the acp_ess_1?  https://solid-client-js-git-acp-ess-export-inrupt.vercel.app/modules/acp_ess_2.html#references

It makes it seem like acp_ess_1 is the more definitive one.  So, if only one can have the definition, I'd almost go for having it in acp_ess_2 and have acp_ess_1 reference the acp_ess_2 doc.

kay-kim (approved):
lgtm - mod couple of minor comments.  Feel free to ignore the comments.

src/acp/ess1.ts
matthieubosquet: I made an explicit statement about ESS 1.1 and ESS 2.0 support in https://github.com/inrupt/solid-client-js/commit/6c6385d4f7ddd376053d3825011ab124b2ec1a3f.

There is already a deprecation notice in acp_v3 pointing to this module (in "[src/acp/v3.ts](https://github.com/inrupt/solid-client-js/pull/1564/commits/c4d5957d98993aefb45ade380a09aae53fc87175#diff-ddf00bafa60da66f28fd8d0b033aca4913fc95a3abeaed01e22c8ca2b3ac011d)".

src/acp/ess2.ts
matthieubosquet: Fixed as per previous comment (now includes an explicit statement about ESS 2.0.

Interestingly, it was enough to ammend the TypeDoc entrypoints ordering to get all functions descriptions in acp_ess_2 (and references in acp_ess_1): https://github.com/inrupt/solid-client-js/pull/1564/commits/143f52d90ee7ca2df9f4943ce57d660fa9f0de5c.

I'm not sure how to have them in both places I'm afraid, hope it's good enough.

PR #1567, opened 2022-04-08
Allow setting client IDs that are not URLs

matthieubosquet:
This PR allows using client IDs as strings.

I have deprecated the setClient function that used to just override all clients that were set in a matcher with just one (I don't think it makes much sense...). I could re-enable it and fix it if it's a problem.


src/acp/matcher.ts
```diff
@@ -823,8 +824,9 @@ export function getClientAll(matcher: Matcher): WebId[] {
  * @param matcher The Matcher for which clients are set.
  * @param client The Client the Matcher should apply to.
  * @returns A copy of the input Matcher, applying to a different set of Clients.
- * @since Not released yet.
+ * @deprecated.
```
NSeydoux, line 827: Why deprecate this? I think we have similar functions to set a user on a matcher too. I don't think it's problematic.

src/acp/matcher.ts
```diff
@@ -850,6 +852,9 @@ export function setClient(matcher: Matcher, client: WebId): Matcher {
  * @since Not released yet.
  */
 export function addClient(matcher: Matcher, client: WebId): Matcher {
```
NSeydoux, line 854: Ditto WebId

src/acp/matcher.ts
```diff
@@ -866,6 +871,9 @@ export function addClient(matcher: Matcher, client: WebId): Matcher {
  * @since Not released yet.
  */
 export function removeClient(matcher: Matcher, client: WebId): Matcher {
```
NSeydoux, line 873: Ditto WebId

src/acp/matcher.ts
```diff
@@ -808,9 +809,9 @@ export function removeCreator(matcher: Matcher): Matcher {
  * @since Not released yet.
  */
 export function getClientAll(matcher: Matcher): WebId[] {
-  return getIriAll(matcher, acp.client).filter(
-    (client: WebId) => client !== solid.PublicOidcClient
-  );
+  return getIriAll(matcher, acp.client)
+    .concat(getStringNoLocaleAll(matcher, acp.client))
+    .filter((client: WebId) => client !== solid.PublicOidcClient);
```
NSeydoux, line 814:
```suggestion
  return getIriAll(matcher, acp.client)
    .filter((client: WebId) => client !== solid.PublicOidcClient)
    .concat(getStringNoLocaleAll(matcher, acp.client));
```
The non-IRI clients would never match the filter condition.

src/acp/matcher.test.ts
```diff
@@ -141,14 +140,20 @@ const MOCK_CLIENT_WEBID_1 = DataFactory.namedNode(
...
   predicate: NamedNode,
   objects: Url[]
 ): ThingPersisted => {
   return objects.reduce((thingAcc, object) => {
-    return addUrl(thingAcc, predicate, object);
+    try {
+      return addUrl(thingAcc, predicate, object);
+    } catch (e) {
+      return addStringNoLocale(thingAcc, predicate, object.value);
+    }
```
NSeydoux, line 156: Wouldn't it be cleaner to use `internal_isValidUrl` ? Even if this works, I find it a bit surprising ^^.

src/acp/matcher.test.ts
```diff
@@ -141,14 +140,20 @@ const MOCK_CLIENT_WEBID_1 = DataFactory.namedNode(
 const MOCK_CLIENT_WEBID_2 = DataFactory.namedNode(
   "https://your.app/registration#it"
 );
+const MOCK_CLIENT_WEBID_3 = DataFactory.namedNode("test_string_client_id");
```
NSeydoux, line 143: It would be more accurate to rename all these `MOCK_CLIENT_IDENTIFIER_X`

src/acp/matcher.test.ts
```diff
@@ -141,14 +140,20 @@ const MOCK_CLIENT_WEBID_1 = DataFactory.namedNode(
...
+const MOCK_CLIENT_WEBID_4 = DataFactory.namedNode("other_string_client_id");
```
NSeydoux, line 144: Also, isn't a `namedNode` always an IRI? Wouldn't it be possible to this to be just a plain string?

src/acp/matcher.test.ts
```diff
@@ -1320,51 +1326,21 @@ describe("getClientAll", () => {
   });
 });
 
-describe("setClient", () => {
```
NSeydoux, line 1323: I would not remove the tests until the deprecated function is removed from the codebase, instead of removing the implementation from the code coverage.

NSeydoux (approved):
A couple of comments, but all superficial really.

src/acp/matcher.ts
```diff
@@ -823,8 +824,9 @@ export function getClientAll(matcher: Matcher): WebId[] {
  * @param matcher The Matcher for which clients are set.
  * @param client The Client the Matcher should apply to.
  * @returns A copy of the input Matcher, applying to a different set of Clients.
- * @since Not released yet.
+ * @deprecated.
  */
+/* istanbul ignore next */
 export function setClient(matcher: Matcher, client: WebId): Matcher {
```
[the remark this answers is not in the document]
matthieubosquet: I think we should just change the type to string or URL.
(As part of a separate PR)

src/acp/matcher.ts
matthieubosquet: (Same)

src/acp/matcher.ts
matthieubosquet: (Same)

src/acp/matcher.test.ts
matthieubosquet: Oh absolutely ^^,

src/acp/matcher.ts
ThisIsMissEm: If the since says it's not yet released, couldn't we just wholesale delete it? Or did someone forget to mark it as released?

src/acp/matcher.test.ts
NSeydoux: To be fair, that's *exactly* how `internal_isValidUrl` is implemented, but still :p

src/acp/matcher.ts
NSeydoux: It is part of the `acp_v4` export, so it is part of our public API. I wasn't sure about this (as the commit history shows ^^), but I expect it may be a common use case to want to enforce only the current client has access to a given resource, mostly for config/indexing stuff. Even if it isn't something we want to encourage overall, having a consistent way of doing so, and to interact with triples in general, with `get`, `set`, `add` and `remove` for the `<s> <p> ?o` operations.

src/acp/matcher.ts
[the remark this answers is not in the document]
NSeydoux: While I was at it... 9f72e60491a734496d4cae704522a8c1ce234a2c

src/acp/matcher.ts
NSeydoux: d7f7f2e0

src/acp/matcher.test.ts
NSeydoux: f0abb7b3

PR #1584, opened 2022-04-20
Use ThingPersisted for asUrl argument type instead of IsNotThingLocal

ianconsolata:
This is a simple bugfix for an issue I was facing with the typescript types of `asUrl`. Felt it was easier to fix than submit an issue, but happy to submit the issue as well. 

The following Typescript code wouldn't compile and gave very confusing errors, despite being correct according to the documentation and (I think) the spirit of this function:
```
export function getUUID(thing: Thing): UUIDString | null {
  if (!isThingLocal(thing)) {
    const thingPersisted: ThingPersisted = thing as ThingPersisted;
    const url = asUrl(thingPersisted);
    if (isUUID(url)) {
      return url;
    }
  }

  return getUrlAll(thing, OWL.sameAs).find(isUUID) || null;
}
```
I don't know why the original type was failing -- it seems to be the guard should not have reset it to `never` given my type assertion, but it was easier to change the signature to something simpler than debug precisely why it was failing.

I looked through the code, and `asUrl` is the only place the type guard IsNotThingLocal is being used. Since ThingLocal and ThingPersisted are the only two Thing subtypes right now, this logic is equivalent to the logic as currently implemented. This could potentially cause issues for any users of the library who have custom ThingPersisted extensions, but they should be able to use type assertions (as I do above) to pass the compile check even with a custom subtype.

In general, I also think it's a lot less confusing if the library is permissive (i.e. `ThingPersisted | SomeOtherNotThingLocal`), rather that using a guard (i.e. `IsNotThingLocal<T>`). It took me a long time to figure out where the `never` type was coming from in the error message, whereas now the error will clearly state that the passed in Thing does not match a ThingPersisted if I give it the wrong argument.

### DIFF
--- a/src/thing/thing.ts
+++ b/src/thing/thing.ts
@@ -323,7 +323,6 @@ export function isThing<X>(input: X | Thing): input is Thing {
   );
 }
 
-type IsNotThingLocal<T extends Thing> = T extends ThingLocal ? never : T;
 /**
  * Get the URL to a given [[Thing]].
  *
@@ -331,9 +330,7 @@ type IsNotThingLocal<T extends Thing> = T extends ThingLocal ? never : T;
  * @param baseUrl If `thing` is not persisted yet, the base URL that should be used to construct this [[Thing]]'s URL.
  */
 export function asUrl(thing: ThingLocal, baseUrl: UrlString): UrlString;
-export function asUrl<T extends ThingPersisted>(
-  thing: T & IsNotThingLocal<T>
-): UrlString;
+export function asUrl(thing: ThingPersisted): UrlString;
 export function asUrl(thing: Thing, baseUrl: UrlString): UrlString;
 export function asUrl(thing: Thing, baseUrl?: UrlString): UrlString {
   if (isThingLocal(thing)) {


ThisIsMissEm:
Hey @ianconsolata, thanks for the PR, I've raised it internal for review but it may take a few days at minimum to get this merged.

ThisIsMissEm:
@ianconsolata if you rebase this with main then vercel will fix itself.

ianconsolata:
I noticed the tests weren't passing on my branch, and then noticed they weren't passing on `main` either. Is there a reasons you're not running `npm run test` in CI?

ThisIsMissEm:
> 

Yeah, third-party PRs do not have access to secrets, so we can't run them

ThisIsMissEm:
It's currently choking on `// @ts-expect-error` though, but I'm not sure where that is

NSeydoux:
As @ThisIsMissEm pointed out, this fix would introduce a regression, caught by the following test: 
```
  it("triggers a TypeScript error when passed a ThingLocal without a base IRI", () => {
    const localThing: ThingLocal = createThing();

    // @ts-expect-error
    expect(() => asUrl(localThing)).toThrow();
  });
```
That means that we'd expect Typescript to complain about `asUrl(localThing)`, but it does not.

Note that a `ThingLocal` and a `ThingPersisted` are very similar from a type system standpoint, the only difference being the shape of the IRI of the thing. `ThingLocal` uses a well-known IRI which is erased when the dataset is saved at a given IRI, and the `ThingPersisted` uses any IRI. This is why the type guard was used: we can't just check that the `Thing` is a `ThingPersisted`, because a `ThingLocal` would pass that test, we actually have to test that the `Thing` is **not** a `ThingLocal`, because we know the shape of the IRI to be matched.

ianconsolata:
Huh, that does work. I was doing something similar before with a ternary, and I couldn't get it to work so I refactored it to an if and added the assertion. Weirdly enough, trying to build it with the type assertion fails with the following error, even though your version without the type assertion succeeds.
```
(typescript) Error: /Users/i/workspace/mysilio-co/garden-kit/src/utils.ts(52,23): semantic error TS2345: Argument of type 'ThingPersisted' is not assignable to parameter of type 'never'.
```

Gonna close for now, since I go things working, but here's a branch with the failing type assertion: https://github.com/mysilio-co/garden-kit/tree/failing-asurl-type. No clue why it breaks when I assert it's a ThingPersisted.

NSeydoux:
Thanks, I'll look into it to see if there is anything we need to change for things to be cleaner

PR #1585, opened 2022-04-20
Improve documentation for getWellKnownSolid

ThisIsMissEm:
# New feature description

This expands the documentation in code for `getWellKnownSolid` but due to the unstable and unspecific results of this method, we've decided to hide it from the public API docs.

# Checklist

### DIFF
--- a/src/resource/solidDataset.ts
+++ b/src/resource/solidDataset.ts
@@ -1156,6 +1156,8 @@ function resolveLocalIrisInThing(
 }
 
 /**
+ * @hidden
+ *
  * Fetch a SolidDataset containing information about the capabilities of the
  * storage server that hosts the given resource URL. For more information,
  * please see the [ESS
--- a/src/resource/solidDataset.ts
+++ b/src/resource/solidDataset.ts
@@ -1156,14 +1156,41 @@ function resolveLocalIrisInThing(
 }
 
 /**
- * Fetch the contents of '.well-known/solid' for a given resource URL.
+ * Fetch a SolidDataset containing information about the capabilities of the
+ * storage server that hosts the given resource URL. For more information,
+ * please see the [ESS
+ * Documentation](https://docs.inrupt.com/ess/latest/services/discovery-endpoint/#well-known-solid).
  *
- * The contents of the '.well-known/solid' endpoint define the capabilities of the server, and provide their associated endpoints/locations.
- * This behaves similarly to the use of '.well-known' endpoints in e.g. (OIDC servers)[https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig]
+ * **Note:** The data contained in this dataset has changed between ESS 1.1 and
+ * ESS 2.0, as such you will need to check for multiple predicates to support
+ * both versions.
+ *
+ * ```typescript
+ * const wellKnown = await getWellKnownSolid(resource);
+ *
+ * // The wellKnown dataset uses a blank node for the subject all of it’s predicates,
+ * // such that we need to call getThingAll with acceptBlankNodes set to true to
+ * // retrieve back predicates contained within the dataset
+ * const wellKnownSubjects = getThingAll(wellKnown, {
+ *   acceptBlankNodes: true,
+ * });
+ * const wellKnownSubject = wellKnownSubjects[0];
+ *
+ * // Retrieve a value from the wellKnown dataset:
+ * let notificationGateway = getIri(
+ *   wellKnownSubject,
+ *   "http://www.w3.org/ns/solid/terms#notificationGateway"
+ * );
+ * ```
+ *
  *
  * @param url URL of a Resource.
- * @param options Optional parameter `options.fetch`: An alternative `fetch` function to make the HTTP request, compatible with the browser-native [fetch API](https://developer.mozilla.org/docs/Web/API/WindowOrWorkerGlobalScope/fetch#parameters).
- * @returns Promise resolving to a [[SolidDataset]] containing the data at '.well-known/solid' for the given Resource, or rejecting if fetching it failed.
+ * @param options Optional parameter `options.fetch`: An alternative `fetch`
+ * function to make the HTTP request, compatible with the browser-native [fetch
+ * API](https://developer.mozilla.org/docs/Web/API/WindowOrWorkerGlobalScope/fetch#parameters).
+ * @returns Promise resolving to a [[SolidDataset]] containing the data at
+ * '.well-known/solid' for the given Resource, or rejecting if fetching it
+ * failed.
  * @since 1.12.0
  */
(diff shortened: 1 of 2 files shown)


src/resource/solidDataset.ts
```diff
@@ -1156,14 +1156,41 @@ function resolveLocalIrisInThing(
...
+ * storage server that hosts the given resource URL. For more information,
+ * please see the [ESS
+ * Documentation](https://docs.inrupt.com/ess/latest/services/discovery-endpoint/#well-known-solid).
+ *
+ * **Note:** The data contained in this dataset has changed between ESS 1.1 and
+ * ESS 2.0, as such you will need to check for multiple predicates to support
+ * both versions.
+ *
+ * ```typescript
+ * const wellKnown = await getWellKnownSolid(resource);
```
RubenVerborgh, line 1169: First of all: the .well-known document is broken IMHO. So no perfect solution exists.

Here are the assumptions that we can seemingly make about [the](https://docs.inrupt.com/ess/1.2/services/discovery-endpoint/#well-known-solid):
1. there will be a `solid:DiscoveryDocument`
2. it will have a `solid:notificationGateway` triple attached to it

I'm not sure if assumption 1 will always be true (not specified); but 2 would or it would blow up. So let's take it from there.

src/resource/solidDataset.ts
```diff
@@ -1156,14 +1156,41 @@ function resolveLocalIrisInThing(
...
+ *
+ * // The wellKnown dataset uses a blank node for the subject all of it’s predicates,
+ * // such that we need to call getThingAll with acceptBlankNodes set to true to
+ * // retrieve back predicates contained within the dataset
+ * const wellKnownSubjects = getThingAll(wellKnown, {
+ *   acceptBlankNodes: true,
+ * });
+ * const wellKnownSubject = wellKnownSubjects[0];
```
RubenVerborgh, line 1177: What we're doing here is getting all subjects in the document, and getting the first. But the document might contain anything, and RDF defines no fixed order. So `[0]` could point to absolutely anything.

src/resource/solidDataset.ts
```diff
@@ -1156,14 +1156,41 @@ function resolveLocalIrisInThing(
...
+ *
+ * // Retrieve a value from the wellKnown dataset:
+ * let notificationGateway = getIri(
+ *   wellKnownSubject,
+ *   "http://www.w3.org/ns/solid/terms#notificationGateway"
+ * );
```
RubenVerborgh, line 1183: So what we _should_ be doing is going through the dataset and look for all quads in the default graph with `http://www.w3.org/ns/solid/terms#notificationGateway` as predicate, and hope there is only one.

I am not sure whether the `Thing` API supports it, but I fear it does not. Does `SolidDataset` somehow support this? https://docs.inrupt.com/developer-tools/api/javascript/solid-client/modules/resource_solidDataset.html#getsoliddataset is unclear about that.

If it does not, then we need to loop over all things, and over each of their predicates.

And we should stop depending on Solid Thing because it does not make sense to approach RDF this way, especially not at the low level we are parsing at here.

RubenVerborgh: Nah it does not.

src/resource/solidDataset.ts
ThisIsMissEm: I don't think 1 holds try for 1.1, https://docs.inrupt.com/ess/latest/services/discovery-endpoint/#well-known-solid

There it's context is `<Pod Host>/solid/v1` and I'm not even sure if this can be retrieved as text/turtle.

RubenVerborgh: Do we have a live one somewhere? Then I can verify.

src/resource/solidDataset.ts
ThisIsMissEm: We're pretty much expecting this to only have one subject in the document — if we fail to get a value, then we just disable that functionality

src/resource/solidDataset.ts
ThisIsMissEm: I don't believe there is any way to iterate over all predicates using the Thing APIs.. I guess we could write:

```
let notificationGateway = wellKnownSubjects.reduce((subject) => {
  const modern = getIri(subject, "http://www.w3.org/ns/solid/terms#notificationGateway");
  const legacy = getIri(subject, "http://inrupt.com/ns/ess#notificationGatewayEndpoint");

  if (modern) {
    return modern;
  }

  if (legacy) {
    return legacy;
  }

  return null;
}, null);
```

Or something..

ThisIsMissEm: But is that really better?

src/resource/solidDataset.ts
RubenVerborgh: That's a very strong assumption, which breaka for instanstance when any value in the JSON structure becomes composite.

(Note that all my remarks are about longterm sustainable coding with RDF data; if that's not a requirement then we're good, but we might be better off reading the JSON.)

src/resource/solidDataset.ts
RubenVerborgh: Nah, and it's a bug in the Thing design.

src/resource/solidDataset.ts
ThisIsMissEm: See the docs link; But regardless, I've now hidden `getWellKnownSolid` from the API docs — it's still documented in code though.

src/resource/solidDataset.ts
ThisIsMissEm: Yup, let's address this by getting spec to cover the reason we have `getWellKnownSolid` — e.g., a `getStorageMetadata(iri: string | URL): Record<string, string | string[]>` or something.

kay-kim (approved):

src/resource/solidDataset.ts
```diff
@@ -1156,14 +1156,43 @@ function resolveLocalIrisInThing(
 }
 
 /**
- * Fetch the contents of '.well-known/solid' for a given resource URL.
+ * @hidden
+ *
+ * Fetch a SolidDataset containing information about the capabilities of the
+ * storage server that hosts the given resource URL. For more information,
+ * please see the [ESS
+ * Documentation](https://docs.inrupt.com/ess/latest/services/discovery-endpoint/#well-known-solid).
```
NSeydoux, line 1164: I think I would just add a note to specify that this is an ESS-specific feature, which is implied by linking to the ESS documentation but may not be explicit enough.

NSeydoux (approved):

src/resource/solidDataset.ts
ThisIsMissEm: It's actually not though — but this method is hidden from the docs, so these docs are really only for our benefit.

PR #1596, opened 2022-04-27
Fix the CD E2E tests link

matthieubosquet:
The CD E2E tests have been broken in solid-client since commit https://github.com/inrupt/solid-client-js/commit/e957cc3eff56c87888fc1f7569fa6e0dedac6b29.

See CD run: https://github.com/inrupt/solid-client-js/runs/6178263859 (triggered on release).

And corresponding GitHub action:  https://github.com/inrupt/solid-client-js/blob/84461647c9af6ba2e6102ab2fb46215a9088be07/.github/workflows/release.yml#L193


NSeydoux (approved):

ThisIsMissEm:
Not sure how we managed to miss this, but glad you caught it. 

@NSeydoux @ajacksified I think this adds weight to having a release process that publishes canaries continuously, but versions manually; thoughts?

PR #1597, opened 2022-04-28
Fix node test instructions

edwardsph:
There is no `.env.example` so this was a little misleading.

It might also be worth explaining why the tests only run on 4 environments:
* Inrupt Dev-Next
* Inrupt Production
* Inrupt 1.1
* NSS

1. What difference does this make to the tests?
2. If you want to run any other environment, you have to use one of these preset names and then provide credentials for the environment you are testing - why is the list restricted?

### DIFF
--- a/e2e/node/README.md
+++ b/e2e/node/README.md
@@ -5,9 +5,9 @@
 To run the tests locally:
 
 1. At the root, run `npm install`.
-2. Copy `.env.example` to `.env.test.local`.
-3. Enter the client credentials of the Pod you want the test to run against, and
-   add them to `.env.test.local`.
+2. Copy the relevant `.env.*` file for your environment to `.env.test.local`.
+3. Add the client credentials of the Pod you want the test to run against
+   into `.env.test.local`.
    For example, for pod.inrupt.com, you can obtain them via
    https://broker.pod.inrupt.com/catalog.html.
 4. You can now run `npm run test:e2e:node` from the root.


edwardsph:
I also discovered that if a previous test run fails and leaves files behind, then many of the tests will fail. This either needs clearly documenting or the tests making more robust so that this doesn't happen.

ThisIsMissEm:
> I also discovered that if a previous test run fails and leaves files behind, then many of the tests will fail. This either needs clearly documenting or the tests making more robust so that this doesn't happen.

This shouldn't be the case? Or, at the very least, these files should be ignored by git.

edwardsph:
>  Or, at the very least, these files should be ignored by git.

I'm referring to test resources/containers created in the pod with names such as `solid-client-test-e2e-acp`, `solid-client-tests`

ThisIsMissEm:
Looks like this relates to #1461

edwardsph:
Yes, I used the env templates added in https://github.com/inrupt/solid-client-js/pull/1461 - it was just that the instructions weren't in sync with that change.
I moved the other question to it's own issue: https://github.com/inrupt/solid-client-js/issues/1598

ThisIsMissEm:
> > Or, at the very least, these files should be ignored by git.
> 
> I'm referring to test resources/containers created in the pod with names such as `solid-client-test-e2e-acp`, `solid-client-tests`

Ah, these should be cleaned up, though that might not be functioning correctly. There is code in test-helpers.ts to try and ensure this

matthieubosquet (approved):

ThisIsMissEm:
On:

> What difference does this make to the tests?

The different environments have potentially different functionality (either ESS 1.1, ESS 2.0, NSS, or CSS) and we need to keep our SDK compatible with all.

> If you want to run any other environment, you have to use one of these preset names and then provide credentials for the environment you are testing - why is the list restricted?

It's currently restricted to make it less likely to be misconfigured — once again, these tests aren't designed to be run against other environments than the ones we develop against.

ThisIsMissEm:
This'll need a rebase and also we should keep in mind that the current environment names are going to change soon, to align with the names we use on the ESS team. (it's just a matter of creating the new environments and updating the code)

ThisIsMissEm:
Have cherry-picked the commit from this PR into #1667 so that we can finally call this done for you.

PR #1604, opened 2022-05-10
Fetch well known solid from root

matthieubosquet:
This makes fetching the well known solid from the server's root by default.

### DIFF
--- a/src/resource/solidDataset.test.ts
+++ b/src/resource/solidDataset.test.ts
@@ -74,7 +74,6 @@ import { removeStringNoLocale } from "../thing/remove";
 import { ldp, rdf } from "../constants";
 import { getUrl } from "../thing/get";
 import { getLocalNodeIri } from "../rdf.internal";
-import { JsonObject } from "fast-check/lib/types/arbitrary/_internals/helpers/JsonConstraintsBuilder";
 
 jest.mock("jsonld", () => ({
   toRDF: jest.fn().mockImplementation(() =>
--- a/src/resource/solidDataset.test.ts
+++ b/src/resource/solidDataset.test.ts
@@ -19,7 +19,7 @@
  * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
-import { jest, describe, it, expect } from "@jest/globals";
+import { beforeAll, jest, describe, it, expect } from "@jest/globals";
 import type { Mock } from "jest-mock";
 
 jest.mock("../fetcher.ts", () => ({
@@ -74,6 +74,7 @@ import { removeStringNoLocale } from "../thing/remove";
 import { ldp, rdf } from "../constants";
 import { getUrl } from "../thing/get";
 import { getLocalNodeIri } from "../rdf.internal";
+import { JsonObject } from "fast-check/lib/types/arbitrary/_internals/helpers/JsonConstraintsBuilder";
 
 jest.mock("jsonld", () => ({
   toRDF: jest.fn().mockImplementation(() =>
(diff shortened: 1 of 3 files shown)


src/resource/solidDataset.test.ts
```diff
@@ -3534,103 +3534,205 @@ describe("getWellKnownSolid", () => {
...
+        url: resourceUrl,
+        headers: {
+          "Content-Type": "text/turtle",
+          link: `</username>; rel="http://www.w3.org/ns/pim/space#storage"`,
+        },
+      })
+    );
+    // Fetches Pod root well known
+    mockedFetcher.fetch.mockResolvedValueOnce(
+      new Response(
```
ThisIsMissEm, line 3673: Could we store these responses in variables ahead of time? That or in a fixtures directory?

src/resource/solidDataset.ts
```diff
@@ -1202,6 +1202,19 @@ export async function getWellKnownSolid(
...
+
+  // Try to fetch the well-known solid dataset from the pod's root
+  try {
+    const wellKnownSolidUrl = new URL(
+      "/.well-known/solid",
+      new URL(urlString).origin
+    ).href;
+    return await getSolidDataset(wellKnownSolidUrl);
+  } catch (e) {
+    // In case of error, do nothing and try to discover the .well-known
```
ThisIsMissEm, line 1214: Would it be worth storing this error in case the rootResource isn't found?

ThisIsMissEm (commented):
Quick initial review, two questions/concerns

src/resource/solidDataset.test.ts
matthieubosquet: I would rather be able to read the tests easily ^^,

src/resource/solidDataset.ts
matthieubosquet: I fixed up the commnent of the outer function.
All we know here is the server's root well known solid is not accessible, then we can try the other mechanism, when we're done with it, all we know is neither worked... not which one should have worked... I guess the resource might even not be hosted on a solid server ^^,
So, no, I don't think we need to store the error.

For what it's worth, the behaviour was the same prior (the error was catched and the other mechanism tried out).

src/resource/solidDataset.ts
```diff
@@ -1202,6 +1202,19 @@ export async function getWellKnownSolid(
   > = internal_defaultFetchOptions
 ): Promise<SolidDataset & WithServerResourceInfo> {
   const urlString = internal_toIriString(url);
+
+  // Try to fetch the well-known solid dataset from the server's root (ESS 2.0 behaviour)
```
NSeydoux, line 1206: This is a comment out of scope for the current PR, but do you think it would be worth documenting systematically, for each function of the public API, if it is a function built against the Solid spec or a particular implementation ? Something like a custom TSDoc tag for instance.

src/resource/solidDataset.ts
NSeydoux: I think I agree with @matthieubosquet: the error that matters is that none of the `.well-known` discovery mechanism worked as a whole, rather than detailing each failed attempt. This whole thing is implementation-specific anyway, once the discovery mechanism is set at the spec level this will be completely rewritten.

NSeydoux (approved):

src/resource/solidDataset.ts
matthieubosquet: Custom TSDoc tag sounds interesting, but I'm not sure it would be very appropriate inside of functions or how standard it is (a quick search [yielded an issue](https://github.com/microsoft/tsdoc/issues/21), but nothing on [the docs](https://tsdoc.org)).

I see however that there is an [experimental tag](https://tsdoc.org/pages/tags/experimental/) that I would recommend using for everything that is not standard. I also think it would be good to [tag as alpha](https://tsdoc.org/pages/tags/alpha/) everything that is not in a versionned standard ([at this point in time everything but Solid Protocol](https://solidproject.org/TR/)).

Furthermore, the [see tag](https://tsdoc.org/pages/tags/see/) seems appropriate for spec links and the [label tag](https://tsdoc.org/pages/tags/label/) is probably the best fit for implementation specific functions.

matthieubosquet: (I removed this version pointer for now)

PR #1608, opened 2022-05-12
Return a more accurate source URL on container creation

matthieubosquet:
If a server applies some URL normalisation when creating a container, then the SDK would not reflect that in the source URL returned from container creation requests.

We can safely assume that if URL normalisation happens, a server will return an absolute URL as Location Header.

Therefore, the function used to create a container now tries to parse the location header as an absolute header to return a more accurate source URL.

### DIFF
--- a/src/resource/solidDataset.test.ts
+++ b/src/resource/solidDataset.test.ts
@@ -177,57 +177,59 @@ describe("responseToSolidDataset", () => {
       .mockReturnValue("https://some.pod/resource");
     const solidDataset = await responseToSolidDataset(response);
 
-    expect(solidDataset).toEqual(expect.objectContaining({
-      graphs: {
-        default: {
-          "https://some.pod/resource": {
-            predicates: {
-              "http://www.w3.org/1999/02/22-rdf-syntax-ns#type": {
-                namedNodes: [
-                  "http://xmlns.com/foaf/0.1/PersonalProfileDocument",
-                ],
-              },
-              "http://xmlns.com/foaf/0.1/maker": {
-                namedNodes: ["https://some.pod/resource#me"],
-              },
-              "http://xmlns.com/foaf/0.1/primaryTopic": {
-                namedNodes: ["https://some.pod/resource#me"],
-              },
-            },
-            type: "Subject",
-            url: "https://some.pod/resource",
-          },
-          "https://some.pod/resource#me": {
-            predicates: {
-              "http://www.w3.org/1999/02/22-rdf-syntax-ns#type": {
-                namedNodes: ["http://xmlns.com/foaf/0.1/Person"],
-              },
-              "http://www.w3.org/2006/vcard/ns#fn": {
-                blankNodes: [
-                  {
-                    "https://some.pod/resource#predicate": {
-                      namedNodes: ["for://a.blank/node"],
-                    },
-                  },
-                ],
-                literals: {
-                  "http://www.w3.org/2001/XMLSchema#string": ["Vincent"],
+    expect(solidDataset).toEqual(
+      expect.objectContaining({
+        graphs: {
+          default: {
+            "https://some.pod/resource": {
+              predicates: {
+                "http://www.w3.org/1999/02/22-rdf-syntax-ns#type": {
+                  namedNodes: [
+                    "http://xmlns.com/foaf/0.1/PersonalProfileDocument",
+                  ],
+                },
+                "http://xmlns.com/foaf/0.1/maker": {
+                  namedNodes: ["https://some.pod/resource#me"],
+                },
+                "http://xmlns.com/foaf/0.1/primaryTopic": {
+                  namedNodes: ["https://some.pod/resource#me"],
                 },
               },
+              type: "Subject",
+              url: "https://some.pod/resource",
             },
-            type: "Subject",
-            url: "https://some.pod/resource#me",
+            "https://some.pod/resource#me": {
+              predicates: {
+                "http://www.w3.org/1999/02/22-rdf-syntax-ns#type": {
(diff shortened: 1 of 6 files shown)


src/resource/solidDataset.test.ts
```diff
@@ -2859,6 +2863,52 @@ describe("createContainerInContainer", () => {
...
+
+    const savedSolidDataset = await createContainerInContainer(
+      "https://some.pod/parent-container//a/b",
+      {
+        fetch: mockFetch,
+      }
+    );
+
+    expect(savedSolidDataset!.internal_resourceInfo.sourceIri).toBe(
+      "https://some.pod/parent-container//a/x/y/"
```
ThisIsMissEm, line 2908: isn't the goal here to strip the additional `/` in `//` ?

src/resource/solidDataset.ts
```diff
@@ -816,27 +816,35 @@ export async function createContainerInContainer(
     );
   }
 
-  const locationHeader = response.headers.get("Location");
-  if (locationHeader === null) {
+  const internalResourceInfo = internal_parseResourceInfo(response);
```
ThisIsMissEm, line 819: in `resource.internal.ts` we've added `contentLocation` but none of the code below uses it, and will in fact throw if the Location header is not present..?

ThisIsMissEm (commented):
I'm not sure if the code here is correct, and would definitely be more confident with an e2e test being added to check things really work.

src/resource/solidDataset.test.ts
matthieubosquet: If the location header is relative, we can't presume which URL normalisation happens on the server.

src/resource/solidDataset.ts
matthieubosquet: Throwing if the location header is not present is the behaviour that was previously adopted ^^,

I'm not necessarily saying it's what we want, but it's what was decided upon prior this PR which I would rather not change unless it's reported as a bug.

We could remove content-location. I added it because prior my research into the subject, it was evaluated as a potential candidate for this feature (my understanding is that location is the correct header). I left the header parsing in the internal function because it might come in handy in other situations (content negotiated resources). Happy to remove if need be.

acoburn: In the context of creating a container via POST where the response is 201, both `Location` and `Content-Location` headers are possible responses under [RFC 7231](https://datatracker.ietf.org/doc/html/rfc7231). The Solid Protocol does not say anything about this.

The difference between `Location` and `Content-Location` (in this context) entirely depends on whether the response contains a payload. Again, the Solid Protocol is silent on this matter.

For instance, _if_ the response is 201 to a `POST` request _and_ the response contains a payload with the representation of the new resource, then a `Content-Location` response header will reference the URI that corresponds to that representation.

If, however, the response is 201 to a `POST` request but the response does _not_ contain a payload, then the appropriate response header would be `Location`, which will indicate the location of the newly created resource.

In other words, a client library should consider both cases possible and act accordingly, which is to say that (from my interpretation of the Solid Protocol specification) a conforming Solid Server could return either `Location` or `Content-Location` as part of a successful `POST` operation.

acoburn: In the context of creating a container via POST where the response is 201, both `Location` and `Content-Location` headers are possible responses under [RFC 7231](https://datatracker.ietf.org/doc/html/rfc7231). The Solid Protocol does not say anything about this.

The difference between `Location` and `Content-Location` (in this context) entirely depends on whether the response contains a payload. Again, the Solid Protocol is silent on this matter.

For instance, _if_ the response is 201 to a `POST` request _and_ the response contains a payload with the representation of the new resource, then a `Content-Location` response header will reference the URI that corresponds to that representation.

If, however, the response is 201 to a `POST` request but the response does _not_ contain a payload, then the appropriate response header would be `Location`, which will indicate the location of the newly created resource.

In other words, a client library should consider both cases possible and act accordingly, which is to say that (from my interpretation of the Solid Protocol specification) a conforming Solid Server could return either `Location` or `Content-Location` as part of a successful `POST` operation.

matthieubosquet: My interpretation of [RFC 7231](https://datatracker.ietf.org/doc/html/rfc7231#section-7.1.2) is that even if the request to create a resource contains a payload, the `Location` header should be a reliable indicator of that new resource's URI.

>  For 201 (Created) responses, the Location value refers to the primary
>  resource created by the request.

It is possible however to have two headers:
- the `Content-Location` header referring to the resource [matching the representation type](https://datatracker.ietf.org/doc/html/rfc7231#section-3.1) sent ("specific"); and,
- the `Location` header still relevant and present, but referring to the more generic URI of the created resource ("primary").
 
>  Note: The Content-Location header field ([Section 3.1.4.2](https://datatracker.ietf.org/doc/html/rfc7231#section-3.1.4.2)) differs
>  from Location in that the Content-Location refers to the most
>  specific resource corresponding to the enclosed representation.
>  It is therefore possible for a response to contain both the
>  Location and Content-Location header fields.

In other words, my interpretation of the `Content-Location` header applies mostly to content negotiable resources. That is, you send RDF over the wire serialised as turtle, the `Content-Location` header will point you to the URI of the resource serialised as Turtle (if content-negotiation is reflected in resource URIs) instead of the primary, content negotiable URI of that RDF resource... the same could apply to images and so on, but in the case of Solid servers makes most sense talking about content negotiable RDF.

I don't think the `Location` header should be missing from responses to requests when indicating the URI of a newly created resource. I don't think that the `Content-Location` header is a valid substitute to answer the created resource URI question.

ThisIsMissEm (approved):
Discussed concerns with @matthieubosquet and decided to approve for merge, fixing other issues later.

PR #1621, opened 2022-05-27
Fix getThingAll API description

matthieubosquet:
Fixes consistency issue in the description of the `getThingAll` function.

### DIFF
--- a/src/thing/thing.ts
+++ b/src/thing/thing.ts
@@ -115,7 +115,7 @@ export function getThing(
 }
 
 /**
- * Get all [[Thing]]s about which a [[SolidDataset]] contains Quads.
+ * Get all [[Thing]]s in a [[SolidDataset]].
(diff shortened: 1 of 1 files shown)


src/thing/thing.ts
```diff
@@ -115,7 +115,7 @@ export function getThing(
 }
 
 /**
- * Get all [[Thing]]s about which a [[SolidDataset]] contains Quads.
+ * Get all [[Thing]]s in a [[SolidDataset]].
```
ThisIsMissEm, line 118: Perhaps we should also link to https://docs.inrupt.com/developer-tools/javascript/client-libraries/structured-data/ ?

ThisIsMissEm (approved):

solid-akb:
Leaving a comment to see if I can :)

PR #1625, opened 2022-05-30
Fix getWellKnownSolid with e2e tests and custom jsonld loader

ThisIsMissEm:
This is just to see if it's working correctly or not across different environments, as I suspect we're stubbing too much in the unit tests.

### DIFF
--- a/src/formats/jsonLd.test.ts
+++ b/src/formats/jsonLd.test.ts
@@ -25,6 +25,12 @@ import { DataFactory } from "n3";
 import jsonld from "jsonld";
 import { getJsonLdParser } from "./jsonLd";
 
+jest.mock("../fetcher.ts", () => ({
+  fetch: jest
+    .fn()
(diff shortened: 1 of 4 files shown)


matthieubosquet (approved):
I checked the implementation of jsonld.

This looks good.

src/formats/jsonLd.test.ts
```diff
@@ -144,4 +150,94 @@ describe("The Parser", () => {
...
+      }
+    `;
+
+      await parser.parse(jsonLd, {
+        internal_resourceInfo: {
+          sourceIri: "https://some.pod/resource",
+          isRawData: false,
+          linkedResources: {},
+        },
+      });
```
matthieubosquet, line 198: Nice :)

matthieubosquet (approved):
It looks good.

I was not familiar with the jsonld module which I checked, well done figuring out how to overload the document loader!

PR #1633, opened 2022-06-03
ci: fix configuration where we've moved away from matrixes

ThisIsMissEm:
This fixes the issue with the latest release's CI failing.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -262,15 +262,12 @@ jobs:
     runs-on: ubuntu-latest
     environment:
       name: "Inrupt Production"
-    strategy:
-      matrix:
-        node-version: ["16"]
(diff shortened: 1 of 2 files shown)


.github/workflows/cd.yml
```diff
@@ -262,15 +262,12 @@ jobs:
...
-    strategy:
-      matrix:
-        node-version: ["16"]
     needs: [prepare-deployment, publish-npm]
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
         with:
-          node-version: ${{ matrix.node-version }}
+          node-version: "16"
```
ThisIsMissEm, line 270: There's no real need to use a matrix here when it's a single value anyway.

ajacksified (approved):

PR #1639, opened 2022-06-09
Run Sonar analysis with test coverage on PRs

edwardsph:
This adds a CI workflow to run a SonarCloud Scan on each PR, including unit tests and the coverage report. The results are added to SonarCloud.

I created this as a separate workflow but it could be integrated into the normal CI if we wanted to avoid running tests twice.

Once this is tested, I would look at adding coverage for the E2E tests as well.


ajacksified:
I think it makes sense to test as-is, then look at how to merge the sonar checks in with the CI process (without it running for each node version) as we look into refactoring our CI/CD pipeline.

ajacksified (approved):

jest.config.js
```diff
@@ -25,7 +25,7 @@ module.exports = {
   testRegex: "/src/.*\\.test\\.ts$",
   clearMocks: true,
   collectCoverage: true,
-  coverageReporters: ["text"],
+  coverageReporters: ["text", "lcov"],
```
ThisIsMissEm, line 28: Would it be possible to only enable this if `CI == true` ? That way our local systems don't get full of reports we don't look at because we look at the terminal output

PR #1644, opened 2022-06-16
Integrate Sonar scan into main CI

edwardsph:
Relies on the jest config to get the coverage run in a CI build but only runs the scan once


ThisIsMissEm (approved):
Nice one!

ThisIsMissEm:
Suggestion for the merge message: `ci: integrate SonarQube Scanner into CI workflow`

edwardsph:
I finally have a configuration that only does the deep fetch when running the Sonar scan so the warning in the scan results has gone. I think this is ready to merge now. Who does that?

ThisIsMissEm (approved):

PR #1660, opened 2022-07-13
chore: adopt shared eslint style guides

ThisIsMissEm:


tsconfig.eslint.json
```diff
@@ -1,6 +1,6 @@
 {
   "extends": "./tsconfig.json",
-  "include": ["src/**/*.ts", "e2e/**/*", "./*.js"],
+  "include": ["src/**/*.ts", "e2e/**/*.ts"],
```
ThisIsMissEm, line 3: The `./*.js` seemed to be pulling in way too much for linting, but I'll try re-enabling it.

tsconfig.json
```diff
@@ -1,72 +1,19 @@
 {
   "compilerOptions": {
-    /* Basic Options */
```
ThisIsMissEm, line 3: This config is the same, largely, sans the comments; The only real change is `lib` of `es2015` to `es6`

tsconfig.json
```diff
@@ -101,7 +48,6 @@
     ],
     "exclude": [
       "node_modules/**",
-      ".codesandbox/**",
```
ThisIsMissEm, line 104: todo: remove codesandbox entirely

src/formats/jsonLd.ts
```diff
@@ -105,7 +112,7 @@ function fixQuads(plainQuads: any[]): Quad[] {
   return fixedQuads;
 }
 
-function term(term: any) {
+function term(term: Quad_Object | DefaultGraph) {
```
jeswr, line 115: Why not just `RDF.Term` (https://github.com/rdfjs/types/blob/183bda795f57a9464ddf95deac45a0c4a48879cf/data-model.d.ts#L13)? The only difference in the typings is that `RDF.Term` allows the slightly looser `RDF.BaseQuad` rather than enforcing `RDF.Quad` (https://github.com/rdfjs/types/blob/183bda795f57a9464ddf95deac45a0c4a48879cf/data-model.d.ts#L152); but I don't see the problem anyways since there is an error thrown in either case.

src/resource/resource.test.ts
```diff
@@ -52,6 +43,16 @@ import {
...
+jest.mock("../fetcher.ts", () => ({
+  fetch: jest.fn().mockImplementation(() =>
+    Promise.resolve(
+      new Response(undefined, {
+        headers: { Location: "https://arbitrary.pod/resource" },
+      })
+    )
+  ),
+}));
+
```
jeswr, line 55: Depending on how soon fetch mocking is updated it might be worth factoring this into a factory function or running it in the global jest environment setup

jeswr (approved):

src/acp/policy.ts
```diff
@@ -755,15 +751,15 @@ export function policyAsMarkdown(policy: Policy | ResourcePolicy): string {
   }
   if (allOfRules.length > 0) {
     markdown += "\nAll of these rules should match:\n";
-    markdown += "- " + allOfRules.join("\n- ") + "\n";
+    markdown += `- ${allOfRules.join("\n- ")}\n`;
```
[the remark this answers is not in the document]
ThisIsMissEm: what would the signature of that function be?

src/acp/v4.ts
```diff
@@ -32,23 +32,23 @@ import {
...
-  //getPolicyUrlAll,
+  // acrAsMarkdown,
+  // addAcrPolicyUrl,
+  // addMemberAcrPolicyUrl,
+  // addMemberPolicyUrl,
+  // addPolicyUrl,
+  // getAcrPolicyUrlAll,
+  // getMemberAcrPolicyUrlAll,
+  // getMemberPolicyUrlAll,
+  // getPolicyUrlAll,
```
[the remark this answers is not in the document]
ThisIsMissEm: Good question, I'm going to defer to @matthieubosquet who originally wrote this, iirc; I suspect maybe there's a reason there here, but I don't know.

src/formats/jsonLd.ts
ThisIsMissEm: If RDF.Term works, I'm happy to have a PR to improve on it; the types in that file are all a little weird. I'd probably be in favour of a standalone `@inrupt/solid-parsers` project that pulls in jsonld and turtle and makes them work fully.

src/resource/resource.test.ts
ThisIsMissEm: We still need to plan it, and it's quite a big undertaking. But I do wanna see it resolved. (just not in this PR)

src/formats/jsonLd.ts
jeswr: > the types in that file are all a little weird

How so, I've never had problems working with these type interfaces in the past?

> @inrupt/solid-parsers project

I would note that https://github.com/rubensworks/rdf-parse.js/blob/master/package.json achieves that though it is a little more heavy weight as it supports a number of other serializations.

Otherwise would recommend switching to https://github.com/rubensworks/jsonld-streaming-parser.js which wont require the Term conversion seen here.

tsconfig.json
ThisIsMissEm: deferring for later.

src/formats/jsonLd.ts
ThisIsMissEm: Have created a ticket for this, and will continue without it.

ThisIsMissEm:
@NSeydoux For merging as E2E has started failing due to migration (405 method not allowed)

PR #1680, opened 2022-08-17
fix: always use options.fetch in getWellKnownSolid

ThisIsMissEm:
This PR fixes the bug SDK-2780 partially, as in some situations `fetcher.ts` was failing to load cross-fetch, so the easy work-around is to always use the provided fetch implementation, that way we know that the method will behave correctly in `@inrupt/solid-client-notifications`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,21 @@ The following changes have been implemented but not released yet:
 
 ## [Unreleased]
 
+### Bugfixes
+
+- Modify the internal `getWellKnownSolid` method used by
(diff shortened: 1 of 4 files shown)


jeswr (approved):
The functional changes are fine - but lots of unecessary formatting changes (changing indents, adding lines etc.).

ThisIsMissEm:
> The functional changes are fine - but lots of unnecessary formatting changes (changing indents, adding lines etc.).

Yeah, in the tests it's mostly just trying to simplify them and make them consistent in implementation of fetch mocking.

PR #1709, opened 2022-09-13
Add note for WebID authenticated fetch

NSeydoux:
WebID must be public resources, and may not be Solid resources. As a result, by default they should be fetched unauthenticated. This commit simply adds a note describing this already implemented behavior.

### DIFF
--- a/src/profile/webid.ts
+++ b/src/profile/webid.ts
@@ -76,7 +76,9 @@ export function getAltProfileUrlAllFrom(
  * A WebID Profile may be any RDF resource on the Web, it doesn't have
  * to be a Solid resource. That is why, in order to expose a Solid-enabled part
  * of their profile, some WebID profiles link to a Profile Resource, which may
- * be a Solid resource.
+ * be a Solid resource. A WebID resource should be public, so `getProfileAll` will
+ * issue an unauthenticated request to the WebID, and only use the provided
+ * authenticated `fetch` (if any) to access extended profile documents.
  *
  * @param webId WebID of the agent you want the profile of.
  * @param options Optional parameter
--- a/src/profile/webid.ts
+++ b/src/profile/webid.ts
@@ -75,8 +75,8 @@ export function getAltProfileUrlAllFrom(
  *
  * A WebID Profile may be any RDF resource on the Web, it doesn't have
  * to be a Solid resource. That is why, in order to expose a Solid-enabled part
- * of their profile, some WebID profiles link to a Profile Resource, which may
- * be a Solid resource. A WebID resource should be public, so `getProfileAll` will
+ * of their profile, some WebID profiles link to one or more Extended Profile Resources,
+ * which may are Solid resources. A WebID resource should be public, so `getProfileAll` will
  * issue an unauthenticated request to the WebID, and only use the provided
  * authenticated `fetch` (if any) to access extended profile documents.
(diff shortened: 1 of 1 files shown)


src/profile/webid.ts
```diff
@@ -76,7 +76,9 @@ export function getAltProfileUrlAllFrom(
  * A WebID Profile may be any RDF resource on the Web, it doesn't have
  * to be a Solid resource. That is why, in order to expose a Solid-enabled part
  * of their profile, some WebID profiles link to a Profile Resource, which may
- * be a Solid resource.
+ * be a Solid resource. A WebID resource should be public, so `getProfileAll` will
```
kay-kim, line 79: So ... since we state "to expose a Solid-enabled part ..." above, should  "a Profile Resource, which may be a Solid resource"  be just  "a Profile Resource, which is ..."?  Or  even  "an Extended Profile, which is a Solid Resource"?  

And, with "WebID resource should be public" -- is it a should be or is it a must?  If a must, I'd just say "A WebID profile is publicly readable"

kay-kim (approved):
lgtm - left some comments.

src/profile/webid.ts
kay-kim: Oh ... also, in case your "should be public" stands -- such that, some people make their profiles not publicly readable (mistakenly or not),  just wondering, for the WebID profile part of getProfileAll, do we first do an unauthed fetch, and then, if that fails, we do an authe'd fetch?

ThisIsMissEm:
Why don't we just add a "getWebId" method?

NSeydoux:
We can also add a `getWebID`, but that's a separate issue I think, we'll still want to be able to get both the WebID and all available extended profiles in one call.

ThisIsMissEm:
@NSeydoux hmm, okay, yeah, agreed.

src/profile/webid.ts
ThisIsMissEm: "Do we first do a...", yes, that's the implementation: https://github.com/inrupt/solid-client-js/pull/1709/files#diff-df051b184aa88ad98045d31b7def5698e9b5c75b17fd85b15bc8a8790cf82798R122

keep in mind this isn't strictly specified by a spec yet.

ThisIsMissEm, line 79:
```suggestion
 * be a Solid resource. A WebID resource MUST be public, so `getProfileAll` will
```

I think this is what the spec says, but I might be wrong

ThisIsMissEm: Spec doesn't say "MUST" nor "should", in fact, it no longer says anything about WebID's other than they're resources — i.e., it reintroduced as of September 10th the ability for WebID's to require authentication to read, by way of omitting to specify that they must be publicly readable.

NSeydoux: Arguably, https://www.w3.org/2005/Incubator/webid/spec/identity/ can be interpreted to mandate the WebID to be public (that's the worst sentence to say about a specification), so unless the Solid-WebID spec adds additional constraints, it sounds like a fair assumption that WebID profiles are public documents.

ThisIsMissEm (approved):

ThisIsMissEm:
Let's merge after #1765 is merged and the new environments roll out, that'll fix the CI issues here.

PR #1761, opened 2022-11-18
feat: Add getWebIdDataset method

VirginiaBalseiro:
# Add getWebIdDataset function

# Checklist

### DIFF
--- a/src/profile/webid.test.ts
+++ b/src/profile/webid.test.ts
@@ -34,7 +34,7 @@ import {
   SolidDataset,
   WithServerResourceInfo,
 } from "..";
-import { foaf, pim } from "../constants";
+import { foaf, pim, rdf } from "../constants";
 import { triplesToTurtle } from "../formats/turtle";
 import { toRdfJsQuads } from "../rdfjs.internal";
 import {
@@ -42,6 +42,7 @@ import {
   getPodUrlAll,
   getPodUrlAllFrom,
   getProfileAll,
+  getWebIdDataset,
 } from "./webid";
 
 jest.mock("cross-fetch", () => {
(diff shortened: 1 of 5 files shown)


src/profile/webid.test.ts
```diff
@@ -746,3 +750,34 @@ describe("getPodUrlAllFrom", () => {
...
+  });
+  it("throws an error if fetching fails", async () => {
+    const { fetch: mockedUnauthFetch } = jest.requireMock("cross-fetch") as {
+      fetch: jest.Mock<
+        ReturnType<typeof window.fetch>,
+        [RequestInfo | URL, RequestInit?]
+      >;
+    };
+    mockedUnauthFetch.mockRejectedValueOnce(new Error("error"));
+    await expect(getWebIdDataset(MOCK_WEBID)).rejects.toThrow("error");
```
ThisIsMissEm, line 781: Might be good to make the error message something like `Fetch Failed 500 internal server error` just so the code can't be satisfied by just doing `new Error("error")` or something — i.e., swallowing the error

src/profile/webid.ts
```diff
@@ -195,3 +195,16 @@ export function getPodUrlAllFrom(
...
+/**
+ * Get the WebID Profile document dataset.
+ *
+ * @param webId The WebID of the agent whose WebID Profile dataset is to be fetched.
+ * @returns a SolidDataset for the WebID Profile document.
+ * @since unreleased
+ */
+export async function getWebIdDataset(
+  webId: WebId
+): Promise<SolidDataset | null> {
```
ThisIsMissEm, line 208:
```suggestion
): Promise<SolidDataset> {
```

ThisIsMissEm (approved):
A few small changes, but otherwise, great work!

src/profile/webid.ts
ThisIsMissEm: You may alternatively wish to use 
```suggestion
): Promise<ReturnType<getSolidDataset>> {
```

As `getSolidDataset` returns a `SolidDataset & WithResourceInfo`

NSeydoux: In this case I think it's better to only treat this as a a `SolidDataset`, as that's already a misnomer, but the `WithResourceInfo` type is really relevant only for Solid resources, which isn't an assumption we can make about WebID profiles.

NSeydoux: Actually my bad, you're right: since other functions of this module (e.g. `getAltProfileUrlAllFrom`) expect a `SolidDataset & WithResourceInfo`, it's probably better to remain aligned.

PR #1765, opened 2022-11-21
Set up CI matrix experimental flag and concurrency

NSeydoux:


.github/workflows/audit.yml
```diff
@@ -7,9 +7,11 @@ on:
...
   schedule:
     - cron: "40 10 * * *"
-
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
 jobs:
   audit:
-    uses: inrupt/javascript-style-configs/.github/workflows/reusable-audit.yml@v1.1.0
+    uses: inrupt/typescript-sdk-tools/.github/workflows/reusable-lint.yml@v1.4.2
```
ThisIsMissEm, line 15: Copy and paste error: 
```suggestion
    uses: inrupt/typescript-sdk-tools/.github/workflows/reusable-audit.yml@v1.4.2
```

.github/workflows/e2e-browser.yml
```diff
@@ -7,90 +7,49 @@ env:
   CI: true
   DEBUG: "pw:webserver"
   ACTIONS_STEP_DEBUG: true
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
 jobs:
-  build:
```
ThisIsMissEm, line 11: @NSeydoux we actually should move towards this, as otherwise we are building N times when we only need to build once. I know it's inconsistent though across repos. Haven't had a chance to roll out "prebuilding the SDK" across to all repos — this is the whole "one pipeline" dream.

ThisIsMissEm (changes requested):
Just needs to fix the audit action due to a copy and paste error.

.github/workflows/audit.yml
NSeydoux: Oh good catch thanks

.github/workflows/e2e-browser.yml
NSeydoux: Oh I see, I missed that. I've added it back with a short comment explaining this.

.github/workflows/e2e-browser.yml
```diff
@@ -37,60 +36,47 @@ jobs:
...
-        # FIXME: setup-node's caching seems to not restore correctly: https://github.com/actions/setup-node/pull/323/files
+      - # FIXME: setup-node's caching seems to not restore correctly: https://github.com/actions/setup-node/pull/323/files
         # if: steps.setup-node.outputs.cache-hit != 'true'
         run: npm ci --ignore-scripts
-      - name: Build the local code
-        run: npm run build
-
-      - name: Setup Playwright
-        run: npx playwright install --with-deps
+      - run: npm run build
```
ThisIsMissEm, line 74: You don't need this in a "build once, test everywhere" setup.

.github/workflows/e2e-browser.yml
```diff
@@ -37,60 +36,47 @@ jobs:
...
+      - run: npx playwright install --with-deps
 
-      - name: Prepare browser-based end-to-end tests
-        run: npm run test:e2e:browser:build
+      - run: npm run test:e2e:browser:build
```
ThisIsMissEm, line 77: We could also likely do this once for the codebase, and then reuse across tests (FUTURE)

.github/workflows/e2e-node.yml
```diff
@@ -7,26 +7,31 @@ on:
...
+    continue-on-error: ${{ matrix.experimental }}
     strategy:
       matrix:
         os: [ubuntu-latest]
-        node-version: [16.x]
-        # Note: "Inrupt Production" and "Inrupt 1.1" have been disabled due to the migration
-        # "Inrupt Dev-Next" is a 2.x environment.
-        environment-name: ["ESS Dev-Next", "NSS"]
+        node-version: [16.x, 14.x]
+        environment-name: ["ESS PodSpaces", "NSS"]
```
ThisIsMissEm, line 25:
```suggestion
#        environment-name: ["ESS PodSpaces", "NSS"]
        environment-name: ["ESS PodSpaces"]
```

Let's de-provision NSS for now, and focus on getting ESS PodSpaces and ESS Dev-Next rolled out as testing environments. We can add this back in across all the SDKs in the future, if we decide to continue testing against NSS directly from here — we need to actually come up with our formal strategy for testing other servers than ESS.

ThisIsMissEm (changes requested):

.github/workflows/e2e-node.yml
NSeydoux: I agree that we need an overarching strategy, but why remove the NSS tests that are currently implemented and working?

ThisIsMissEm: Because the NSS environment is not managed by our new e2e-environments.

ThisIsMissEm: Though it is incidentally supported here: https://github.com/inrupt/typescript-sdk-tools/blob/main/packages/internal-test-env/index.ts#L34-L39

ThisIsMissEm: Let's get a plan in place for rolling out e2e-environments for NSS and CSS in general, then we can add it back in.

NSeydoux: My concern is that we already have a lot of things to get done before testing on CSS/NSS, so if we disable this now we'll be blind for a while, even when we roll out stuff like the authn refactor, which I'd prefer to be aware of it it breaks NSS compatibility.

ThisIsMissEm (approved):

PR #1800, opened 2022-12-21
Removed test coverage for e2e tests

solid-akb:
This PR is part of a wider clean up effort to remove test coverage for our E2E node tests.
It is part of [ticket SDK-2973](https://inrupt.atlassian.net/browse/SDK-2973).

Details:
The E2E Node.js tests now run via Jest, but we have Jest usually configured to collect code coverage metrics. For now, these metrics do not make sense to collect for our E2E tests, which have traditionally been much lighter (as we rely more on unit tests with mocking), and we’d be better collecting not metrics of the amount of code executed as part of an E2E test run, but instead the number of features/scenarios that were exercised and able to be tested consistently.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -41,7 +41,7 @@
         "ts-jest": "^27.0.3",
         "typedoc": "^0.23.7",
         "typedoc-plugin-markdown": "^3.13.3",
-        "typescript": "^4.9.4"
+        "typescript": "^4.7.4"
       },
       "engines": {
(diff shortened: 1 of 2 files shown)


NSeydoux (approved):
Looks good to me! It's not yet required on this particular repository, as the root cause of the issue (the jest config being consolidated in a single file) hasn't been implemented here, but it's only a matter of time that it is :)

PR #1802, opened 2022-12-26
Update jest to v29

NSeydoux:
This implements a couple of changes to upgrade to jest v29, namely:
- Fix the `Mocked<...>` typing signature
- Regenerate snapshots

This also changes some of the ACP unit tests. These tests are heavily geared towards testing for implementation rather than for behavior, so the checks removed don't significantly reduce the confidence we can have in the codebase.


VirginiaBalseiro (approved):
looks good! needs rebasing though.

PR #1803, opened 2022-12-26
Upgrade typescript to 4.9.0

solid-akb:
Updates the library to use typescript 4.9.0.

### DIFF
--- a/package.json
+++ b/package.json
@@ -189,7 +189,7 @@
     "ts-jest": "^27.0.3",
     "typedoc": "^0.23.7",
     "typedoc-plugin-markdown": "^3.13.3",
-    "typescript": "^4.7.4"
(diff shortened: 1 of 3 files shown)


NSeydoux (approved):

package.json
```diff
@@ -189,7 +189,7 @@
     "ts-jest": "^27.0.3",
     "typedoc": "^0.23.7",
```
ThisIsMissEm, line 190: Do these need an update as well? I think in another pull request you updated them?

ThisIsMissEm (approved):

package.json
NSeydoux: There is a separate PR to update jest: https://github.com/inrupt/solid-client-js/pull/1802

PR #1817, opened 2023-01-09
Change dev dependencies dependabot updates frequency

NSeydoux:


ThisIsMissEm:
I've revoked the token used, see our internal chat for details.

.github/workflows/dependabot-automation.yml
```diff
@@ -0,0 +1,31 @@
...
+jobs:
+  dependabot:
+    runs-on: ubuntu-latest
+    if: ${{ github.actor == 'dependabot[bot]' }}
+    steps:
+      - name: Dependabot metadata
+        id: metadata
+        uses: dependabot/fetch-metadata@v1
+        with:
+          github-token: "${{ secrets.GITHUB_TOKEN }}"
```
ThisIsMissEm, line 16: I think you can use the github actions access token here

NSeydoux:
See https://github.com/inrupt/temp-test-dependabot-automation for experimentations showing the config looks correct.

.github/dependabot.yml
```diff
@@ -1,23 +1,49 @@
...
-      interval: "daily"
+      interval: "weekly"
+    labels:
+      - "dependencies"
+      - "pip"
   # Enable version updates for our CI tooling
   - package-ecosystem: "github-actions"
     # Look for `package.json` and `lock` files in the `root` directory
     directory: "/"
     # Check the npm registry for updates every day (weekdays)
```
chelseapinka, line 44: Should we update this comment to reflect that is is now checking weekly? Also we could add which day of the week we believe the schedule will be triggered on

.github/dependabot.yml
```diff
@@ -1,23 +1,49 @@
...
+      - "npm"
+      - "development"
+    allow:
+      - dependency-type: development
+
   # Enable version updates for the website tooling
   - package-ecosystem: "pip"
     # Look for `package.json` and `lock` files in the `root` directory
     directory: "/docs"
     # Check the npm registry for updates every day (weekdays)
```
chelseapinka, line 34: Should we update the comment to reflect the weekly checks?

.github/dependabot.yml
```diff
@@ -1,23 +1,49 @@
...
+    labels:
+      - "dependencies"
+      - "npm"
+      - "development"
+    allow:
+      - dependency-type: development
+
   # Enable version updates for the website tooling
   - package-ecosystem: "pip"
     # Look for `package.json` and `lock` files in the `root` directory
```
chelseapinka, line 32: update `root` in comment or remove comment entirely?

NSeydoux: Removed in 7b2df5c22ca51381750df223637b8bd1285ae8aa

.github/dependabot.yml
NSeydoux: Fixed in d15f9ee31a2cb09927a87efab780ec6b0a96e478

ThisIsMissEm (commented):
The changes here look reasonable, and inline with what we would want to adopt. However, the pull request title is incorrect, and we probably need to discuss in parking lot next steps on this PR / Ticket.

ThisIsMissEm (approved):

ThisIsMissEm:
Looks like this needs to be rebased against `main` though :(

PR #1821, opened 2023-01-12
Fix: package subexports & move json-ld to being dynamically loading at call-time instead of run-time

ThisIsMissEm:
The other day I tried out that new tool [publint](https://github.com/bluwy/publint), and it identified a [whole bunch of issues](https://publint.dev/@inrupt/solid-client@1.24.0) in our `package.json`, especially relating to compatibility with typescript and how our package subexports work. This aims to resolve these issues.

Whilst I was working on that, I realised that with fairly minimal overhead, we could actually move `jsonld` to be being imported at call time of the asynchronous `getJsonLdParser` method `parse`, which hopefully will allow us to mitigate our ESM woes with `jsonld` just a little bit longer. This incurs minimal overhead, but the module will get cached on the first call by the runtime / module system, so it's only a potential extra cost on first-call of the parse method.

I've also marked `@rdfjs/data-mdoel` as an external dependency, this was missing in rollup's configuration and causing some weirdness. We still need to rectify the global module names in the UMD builds, and the circular dependencies, but hopefully this helps just a little bit.

Finally, there's a few minor test changes where after doing other changes on this branch completely unrelated tests started spewing errors around fetch and Response, and it seems those were just not being imported in those specific tests, or something _weird_ was going on with the fetcher autoloading thing that doesn't actually really work well anyway, but this should put us in a better position to some degree.

At a future point in time, we should look to making publint part of our CI/CD processes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -144,6 +144,10 @@
       "types": "./dist/profile/webid.d.ts",
       "import": "./dist/profile/webid.mjs"
     },
+    "./profile/webid": {
+      "types": "./dist/profile/webid.d.ts",
+      "import": "./dist/profile/webid.mjs"
+    },
     "./formats": {
       "types": "./dist/formats/index.d.ts",
       "import": "./dist/formats/index.mjs"
--- a/package.json
+++ b/package.json
@@ -36,121 +36,121 @@
   "types": "dist/index.d.ts",
   "exports": {
     ".": {
+      "types": "./dist/index.d.ts",
       "require": "./dist/index.js",
-      "import": "./dist/index.mjs",
-      "types": "./dist/index.d.ts"
+      "import": "./dist/index.mjs"
     },
     "./interfaces": {
-      "import": "./dist/interfaces.mjs",
-      "types": "./dist/interfaces.d.ts"
+      "types": "./dist/interfaces.d.ts",
+      "import": "./dist/interfaces.mjs"
     },
     "./resource/resource": {
-      "import": "./dist/resource/resource.mjs",
-      "types": "./dist/resource/resource.d.ts"
+      "types": "./dist/resource/resource.d.ts",
+      "import": "./dist/resource/resource.mjs"
     },
     "./resource/solidDataset": {
-      "import": "./dist/resource/solidDataset.mjs",
-      "types": "./dist/resource/solidDataset.d.ts"
+      "types": "./dist/resource/solidDataset.d.ts",
+      "import": "./dist/resource/solidDataset.mjs"
     },
     "./resource/file": {
-      "import": "./dist/resource/file.mjs",
-      "types": "./dist/resource/file.d.ts"
+      "types": "./dist/resource/file.d.ts",
+      "import": "./dist/resource/file.mjs"
     },
     "./resource/mock": {
-      "import": "./dist/resource/mock.mjs",
-      "types": "./dist/resource/mock.d.ts"
+      "types": "./dist/resource/mock.d.ts",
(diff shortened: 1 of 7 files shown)


src/index.ts
```diff
@@ -243,11 +245,15 @@ export {
...
-  solidDatasetAsTurtle,
-} from "./formats/index";
+
+// Export the different formats methods from @inrupt/solid-client import.
+//
+// NOTE: We have to export like this, otherwise rollup does some sort of
+// tree-shaking and breaks the build where `dist/formats/index.mjs` is no longer
+// built, and this causes the `@inrupt/solid-client/formats` subpath export to
+// no longer work.
+export const { getJsonLdParser, getTurtleParser, solidDatasetAsTurtle } =
```
NSeydoux, line 255: These were left out of the public API on purpose, as it is all still quite experimental/not something we want to commit to. Is it possible to add a note to this intent here, in addition to this not being documented?

package.json
```diff
@@ -36,121 +36,121 @@
   "types": "dist/index.d.ts",
   "exports": {
     ".": {
+      "types": "./dist/index.d.ts",
       "require": "./dist/index.js",
-      "import": "./dist/index.mjs",
-      "types": "./dist/index.d.ts"
+      "import": "./dist/index.mjs"
```
NSeydoux, line 41: Interesting!

package.json
```diff
@@ -36,121 +36,121 @@
...
     "./profile/jwks": {
-      "import": "./dist/profile/jwks.mjs",
-      "types": "./dist/profile/jwks.d.ts"
+      "types": "./dist/profile/jwks.d.ts",
+      "import": "./dist/profile/jwks.mjs"
     },
     "./profile/webId": {
-      "import": "./dist/profile/webId.mjs",
-      "types": "./dist/profile/webId.d.ts"
+      "types": "./dist/profile/webid.d.ts",
```
solid-akb, line 144: 🙌

package.json
```diff
@@ -36,121 +36,121 @@
...
+      "import": "./dist/profile/webid.mjs"
     },
     "./formats": {
-      "import": "./dist/formats/index.mjs",
-      "types": "./dist/formats/index.d.ts"
+      "types": "./dist/formats/index.d.ts",
+      "import": "./dist/formats/index.mjs"
```
solid-akb, line 149: Great find 👏🎉

src/index.ts
ThisIsMissEm: The parsers are still not documented (only solidDatasetAsTurtle is currently), so there's no change in public API, but we need to export them such that our other packages can use them

NSeydoux (approved):

src/index.ts
NSeydoux: Them being exported, making it extra explicit that they aren't public API wouldn't hurt I think.

ThisIsMissEm: So like, just add a comment above saying `// Non-public APIs needed by the other SDKs, these are not part of our public supported API` ?

ThisIsMissEm:
@VirginiaBalseiro I've got one small thing to fix up on this, but let's cut the release after this is merged.

src/index.ts
ThisIsMissEm: Have improved the comment.

package.json
ThisIsMissEm: @NSeydoux I've also added in a subpath export for `"./profile/webid"` in lowercase to encourage that usage instead of the camelCase one.

PR #1820, opened 2023-01-12
Chore: change target environment to ES2018

VirginiaBalseiro:
Change target environment to ES2018.

PR #1828, opened 2023-01-17
docs: fix typo in README.md and docs related to environments supported

ThisIsMissEm:
Tiny follow up to #1820 and #1827 — note to self: acceptance criteria should really have the correct text in there in the future.

### DIFF
--- a/docs/api/source/index.rst
+++ b/docs/api/source/index.rst
@@ -30,8 +30,8 @@ Changelog
 See `the release
 notes <https://github.com/inrupt/solid-client-js/blob/main/CHANGELOG.md>`__.
 
-Supported environment
-~~~~~~~~~~~~~~~~~~~~~
+Supported environments
+~~~~~~~~~~~~~~~~~~~~~~
 
 Our JavaScript Client Libraries use relatively modern JavaScript, aligned with
 the `ES2018 <https://262.ecma-international.org/9.0/>`__ Specification features, we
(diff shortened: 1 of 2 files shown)


docs/api/source/index.rst
```diff
@@ -39,16 +39,16 @@ ship both `ESM <https://nodejs.org/docs/latest-v16.x/api/esm.html>`__ and
 `CommonJS <https://nodejs.org/docs/latest-v16.x/api/modules.html>`__, with type
 definitions for TypeScript alongside.
 
-This mean that we only support environments (browsers or runtimes) that were
-released after mid-2018 out of the box, if you wish to target these
-environments, then you will need to cross-compile our SDKs via the use of
-`Babel <https://babeljs.io>`__, `webpack <https://webpack.js.org/>`__,
-`SWC <https://swc.rs/>`__, or similar.
+This means that out of the box, we only support environments (browsers or
+runtimes) that were released after mid-2018, if you wish to target other (older)
```
chelseapinka, line 43: I would almost just say "older" instead of but other and older

chelseapinka (approved):

docs/api/source/index.rst
ThisIsMissEm: mm, I wanted to emphasise that's it's anything that's not this, not just "older" environments. Maybe `(e.g., older)` ?

chelseapinka: but what other environments other than the older ones actually apply?

ThisIsMissEm: I'm not sure of top of head, tbh. I just wanted to be more broad to cover all basis.

PR #1832, opened 2023-01-20
Set recursive VC access

NSeydoux:
This introduces a new `options.inherit` flag on the `setVcAccess` function. If set to `true`, in addition to setting the VC access on the resource's own Access Control, it also sets it on its Member Access Control so that access is cascaded to contained resources. For backwards compatibility, `options.inherit` defaults to `false`.

This feature is added to be used in conjunction with recursive Access Grants. In the case of a non-recursive Access Grant, it is recommanded to leave the flag to the default `false` value, so that only the target resource has its access changed.

Note: this has been end-to-end tested (successfully) with the Access Grant library.

# Checklist

### DIFF
--- a/src/acp/util/setVcAccess.test.ts
+++ b/src/acp/util/setVcAccess.test.ts
@@ -198,4 +198,98 @@ describe("setVcAccess", () => {
       ]).graphs.default
     );
   });
+
+  it("should set the given access mode on the member access control for recursive VC access", () => {
+    const resource = mockAccessControlledResource();
+
+    expect(resource.internal_acp.acr.graphs).toStrictEqual({ default: {} });
+    const updatedResource = setVcAccess(
+      resource,
+      { read: true },
+      { inherit: true }
+    );
+    expect(updatedResource.internal_acp.acr.graphs.default).toStrictEqual(
+      createDatasetFromSubjects([
+        [
+          DEFAULT_ACCESS_CONTROL_RESOURCE_URL,
+          [
+            [ACP.accessControl, [TEST_URL.defaultAccessControl]],
+            [
+              ACP.memberAccessControl,
+              [
+                // Setting recursive access creates a member access control.
+                TEST_URL.defaultMemberAccessControl,
+              ],
+            ],
+          ],
+        ],
+        [
+          TEST_URL.defaultAccessControl,
+          [[ACP.apply, [TEST_URL.defaultAccessControlVcPolicy]]],
+        ],
+        [
+          TEST_URL.defaultMemberAccessControl,
+          // The member access control should enforce the VC access policy.
+          [[ACP.apply, [TEST_URL.defaultAccessControlVcPolicy]]],
+        ],
+        [
+          TEST_URL.defaultAccessControlVcPolicy,
+          [
+            [rdf.type, [ACP.Policy]],
+            [ACP.anyOf, [TEST_URL.defaultAccessControlVcMatcher]],
+            [ACP.allow, [ACL.Read]],
+          ],
+        ],
+        [
+          TEST_URL.defaultAccessControlVcMatcher,
+          [
+            [rdf.type, [ACP.Matcher]],
+            [ACP.vc, [VC_ACCESS_GRANT]],
+          ],
+        ],
+      ]).graphs.default
+    );
+  });
+
+  it("should create a default member Access Control if not present", () => {
+    const resource = mockAccessControlledResource(
+      createDatasetFromSubjects([
+        [
+          DEFAULT_ACCESS_CONTROL_RESOURCE_URL,
+          [[rdf.type, [ACP.AccessControlResource]]],
+        ],
+        // Note that the default Access Control does not exist
+      ])
+    );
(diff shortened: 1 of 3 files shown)


src/acp/util/setVcAccess.test.ts
```diff
@@ -198,4 +198,98 @@ describe("setVcAccess", () => {
...
+          ],
+        ],
+        [
+          TEST_URL.defaultAccessControlVcMatcher,
+          [
+            [rdf.type, [ACP.Matcher]],
+            [ACP.vc, [VC_ACCESS_GRANT]],
+          ],
+        ],
+      ]).graphs.default
```
matthieubosquet, line 250: This allows read on the resource controlled by `DEFAULT_ACCESS_CONTROL_RESOURCE_URL` and its decendents.

src/acp/util/setVcAccess.ts
```diff
@@ -63,56 +96,63 @@ export const DEFAULT_VC_MATCHER_NAME = "defaultVcMatcher";
...
-  acr = [acrThing, accessControl, vcPolicy, vcMatcher].reduce(setThing, acr);
+  acr = [acrThing, accessControl, policy, matcher].reduce(setThing, acr);
+
+  if (options.inherit) {
+    // Add triples to the member access control and link it to the ACR only
+    // if the VC access is recursive.
+    memberAccessControl = addIri(memberAccessControl, ACP.apply, policy);
+    acrThing = addIri(acrThing, ACP.memberAccessControl, memberAccessControl);
+    acr = [acrThing, memberAccessControl].reduce(setThing, acr);
+  }
```
matthieubosquet, line 155: The inherit property defaults to true, so maybe you want to mimic that here:
- set allow member all the time if the resource is a container
- do not set allow member if inherit is false

src/acp/util/setVcAccess.ts
```diff
@@ -63,56 +96,63 @@ export const DEFAULT_VC_MATCHER_NAME = "defaultVcMatcher";
...
+  );
+
   let accessControl = getDefaultAccessControlThing(
     resourceWithAcr,
     "defaultAccessControl"
   );
+  if (!getIriAll(accessControl, ACP.apply).includes(asIri(policy))) {
+    // Case when the ACR Thing existed, but did not include a link to the default Access Control.
+    accessControl = addIri(accessControl, ACP.apply, policy);
+  }
```
matthieubosquet, line 130: Maybe you want the same condition on the member access setter?

matthieubosquet (commented):
I think the most important question is whether you want to mimic the default behaviour of access grant inheritance.

From an ACP perspective, it looks like what you're trying to do.

src/acp/util/setVcAccess.ts
NSeydoux: The `inherit` property defaults to `true` in Access Grants, so I preferred setting the default behaviour at this layer rather than at the lower ACP layer: https://github.com/inrupt/solid-client-access-grants-js/pull/477/files#diff-ab8c49f7948de39654745d99bd32d00025c021d76c1c9dfdedbbeeda66d48ce1.

src/acp/util/setVcAccess.ts
NSeydoux: My assumption was that potentially duplicating triples wasn't a big deal, because that would result in equivalent RDF graphs, and would get normalized when converted to RDF. Actually I think I'll even remove the condition on the non-member access control, it makes the code a bit easier to read.

NSeydoux: But in addition I'll add a devX ticket to have some way to specify you want a triple added if it isn't already present.

NSeydoux: Actually my initial proposal is probably a bad idea, let's show intent first and do as you suggested

src/acp/util/setVcAccess.ts
matthieubosquet: When looking at the function's signature, I see:

```javascript
  options: {
    inherit: boolean;
  } = { inherit: false }
```

Which means that when this function is called without explicitly saying `inherit: true`, it will default to not setting member access, which is the oposite of what access grant will do. I guess it really depends on what you want and who calls it ^^,

Maybe if we're sure that the caller will always determine whether inheritance must apply or not, the parameter should not be optional?

NSeydoux: Absolutely: it being recursive by default is something that is a piece of logic specific to access grants. That is why the default behaviour in the Access Grant client is the following: 
```
const updatedResource = acp_ess_2.setVcAccess(resourceInfo, accessMode, {
  inherit: options?.inherit ?? true,
});
```
which results in the VC matcher being applied to the member Access Control. However, my assumption was that we want to keep this default recursive behaviour to the Access Grant layer, and have the lower-level ACP layer to default to a more closed access for VCs unless specified otherwise.

This also keeps the breaking change to the Access Grant layer, where it is aligned with the server behaviour.

VirginiaBalseiro (approved):

PR #1845, opened 2023-01-25
Replace custom random number generator with UUID

solid-akb:
This PR replaces our custom random number generator with `uuid`. 

Note: This PR introduces a new package `uuid` to the codebase, which we have audited and is already used by our other SDKs


src/access/acp_v1.ts
```diff
@@ -18,25 +18,18 @@
 // OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
 // SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 //
-
+import { v4 as uuidv4 } from "uuid";
 import { acp } from "../constants";
-import {
```
solid-akb, line 23: This is unrelated to the PR but general cleanup of unused imports

src/access/acp_v1.ts
```diff
@@ -46,11 +39,9 @@ import {
   setResourcePolicy,
 } from "../acp/policy";
 import {
-  createRule,
```
solid-akb, line 49: Cleanup of unused imports

src/access/acp_v1.ts
```diff
@@ -18,25 +18,18 @@
 // OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
 // SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 //
-
+import { v4 as uuidv4 } from "uuid";
```
ThisIsMissEm, line 21: We normally just write: `import { v4 } from "uuid";` but I think renaming here provides more clarity, so 👍

src/thing/thing.ts
```diff
@@ -471,7 +471,5 @@ export class ValidThingUrlExpectedError extends SolidClientError {
  * @returns An string that's likely to be unique
  */
 const generateName = () => {
-  return (
-    Date.now().toString() + Math.random().toString().substring("0.".length)
-  );
+  return uuidv4().toString().substring("0.".length);
```
ThisIsMissEm, line 474: Screenshot 2023-01-25 at 20 20 56

```suggestion
  return uuidv4();
```

The toString / substring thing was due to the behaviour of Math.random(), see the screenshot above.

src/access/acp_v2.ts
```diff
@@ -867,7 +867,7 @@ function copyPolicyExcludingActor(
   const newIriSuffix =
     "_copy_without" +
     `_${encodeURIComponent(actorRelationToExclude)}_${actorToExclude}` +
-    `_${Date.now()}_${Math.random()}`;
+    `_${uuidv4()}`;
```
ThisIsMissEm, line 870: I think all these identifiers for `newIriSuffix` can be removed, and we can just always generate a `uuidv4()` in each of `createResourceMatcherFor` and `createResourcePolicyFor` probably should just be the matcher IRI with the hash replaced with a uuid, rather than appending to the IRI.

e.g., the current implementation may be leaking information about the actor & the relationship in the policy or matcher IRI, when that value carries no significance, and it's really just a "makes the policy nice to read/look at"

Though that may be a wider refactor, so deferring to @NSeydoux

ThisIsMissEm (changes requested):
Looks good, the only change to make is the comment about `return uuidv4().toString().substring("0.".length);`

Though I would like to maybe wait a moment until Nic A.S. or @matthieubosquet can respond about the policy / matcher IRIs and why the format they are in is at all significant. I think ESS is actually even hashing these IRIs (I wasn't able to find anything in my pod that had an ACR that contained the strings we use in this code)

src/thing/thing.ts
solid-akb: Ahh roger that!

src/access/acp_v2.ts
solid-akb: So if I understand this comment better, we can replace lines 867-870 with 

`const newIriSuffix = uuidv4();`

Is that correct? 

Will wait for @NSeydoux response before pushing updates. 

Beyond just changing the value to the code mentioned above, would we need to change anything else? Specifically, do we have code in other places that relies on having the "copy_without", "actorRelationToExclude" or "actorTo Exclude" in the IriSuffix?

ThisIsMissEm: > Beyond just changing the value to the code mentioned above, would we need to change anything else? Specifically, do we have code in other places that relies on having the "copy_without", "actorRelationToExclude" or "actorTo Exclude" in the IriSuffix?

as far as I can tell, no, and in fact, ESS will make those IRIs for subject/object opaque, so they won't be returned back as-is anyway

NSeydoux: There shouldn't be any hard-coded dependencies on the resource names, or reliance on any in-IRI semantics in the low-level API (we know there is some in the high-level API, but that's irrelevant here). I don't think ESS changes these IRIs, for instance when using the permissions app you can give a human-friendly name to a policy and it is persisted as-is. Finally, I think there isn't any information leak here: any user authorized to see the ACR in the first place would be able to see any actor mentioned in it, so it being embedded in an IRI doesn't add any information to what the resource already contains. It is really just for debug purpose.

All that said, it is a good practice to not have semantics in IRIs, so provided there isn't any reliance on this piece of logic, I think we can go forward with Emelia's suggestion and just use the `uuid`.

ThisIsMissEm (approved):
LGTM

src/access/acp_v2.ts
ThisIsMissEm: I've also subsequently removed the rather brittle test for encoding hashes in the ACP IRIs, as we don't do that any more, and the only place this may come in is through existing IRIs maybe.

solid-akb:
This looks good now with the irrelevant tests removed. 👍

PR #1846, opened 2023-01-25
Updated how we create object

solid-akb:
This PR updates how we create the initial object that tracks the dataset changelog. 

This is important because it protects against [prototype pollution](https://learn.snyk.io/lessons/prototype-pollution/javascript/).

### DIFF
--- a/src/resource/solidDataset.ts
+++ b/src/resource/solidDataset.ts
@@ -941,7 +941,7 @@ function sortChangeLogByThingAndProperty(
     }
     const thingUrl = internal_toIriString(subjectNode);
     const propertyUrl = internal_toIriString(deletion.predicate);
-    changeLogsByThingAndProperty[thingUrl] ??= {};
+    changeLogsByThingAndProperty[thingUrl] ??= Object.create(null);
     changeLogsByThingAndProperty[thingUrl][propertyUrl] ??= {
       added: [],
       deleted: [],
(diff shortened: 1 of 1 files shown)


ThisIsMissEm (changes requested):
On the following lines there's a bunch of `??= {}` which also need to be changed too to `Object.create(null)`

solid-akb:
@ThisIsMissEm Thanks for the comment above. This begs the question, is there a way to check that the problem is fixed in SonarCloud before the PR is created or at least before merge? This PR passed all SonarCloud Github actions, but perhaps it probably shouldn't have.

ThisIsMissEm:
> @ThisIsMissEm Thanks for the comment above. This begs the question, is there a way to check that the problem is fixed in SonarCloud before the PR is created or at least before merge? This PR passed all SonarCloud Github actions, but perhaps it probably shouldn't have.

Honestly, I'm not sure! I would've expected SonarCloud to still be failing here, as the vulnerability wasn't entirely fixed?

solid-akb:
@ThisIsMissEm I made the fixes requested above. 

Also did a little dive in SonarCloud world. We probably don't want to make SonarCloud involved at the PR level because then it will flag everything. For now we can make the changes and then see the updates in SonarCloud after the merge, but I am still on the lookout with my work with setting up SonarQube in AMI if we can find a happy medium (like turning off certain flags that we know we will experience and are ok to push through).

ThisIsMissEm (approved):

PR #1869, opened 2023-02-13
Improve error message

NSeydoux:
Currently, the error message displayed when receiving an unsuccessful response when reading or writing a resource only includes status code and text. Additional useful information may be found in the response body, which should be displayed too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,11 +6,11 @@ The following changes have been implemented but not released yet:
 
 ## [Unreleased]
 
-### Patch
+### Bugfixes
 
 - Add the response body along the status code and text in the error messages thrown
   on unsuccessful response from the server. The response body may contain additional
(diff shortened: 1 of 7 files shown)


solid-akb (approved):
Awesome PR and great proactive changes. 

One question I had that is not a blocking question, but is moving from quotes to escaping characters a style choice or does it impact the implementation?

NSeydoux:
It does impact the implementation: 
```
expect(() => myFunc()).rejects.toThrow("Some error message");
```
matches exactly on the error message, so if the thrown error was `Some error message: some additional context` that would fail. On the other hand, `/.../` is a regex, so it matches in a different way. In particular, it can be used to verify that some pieces of the expected message are present, without matching on the entire message, which is why I was using this syntax here. This makes the tests a bit less strict, which isn't necessarily a bad thing when it comes to error messages.

Issue #1862, opened 2023-02-07
AccessModes not exported from universalAccess module or main solid-client module


ianconsolata:
### Search terms you've used

AccessModes

### Bug description

The AccessModes type is not exported from the `universalAccess` module. This means when using TypeScript, there is no clear way to get the associated types while using that module.

### Expected result

I can import the AccessModes type.

### Actual result

I cannot import the AccessModes type from either the universalAccess module or the main solid-client module.

NSeydoux:
Hi @ianconsolata, thanks for reporting this! We'll patch this soon, and I'll update this issue accordingly.

NSeydoux:
https://github.com/inrupt/solid-client-js/releases/tag/v1.26.0 should resolve this. Thanks again for reporting.

jeswr:
Closing as completed - @ianconsolata feel free to re-open the issue if problems persist.

PR #1870, opened 2023-02-14
Fix/export accessmodes type

NSeydoux:
AccessModes is a type exposed as part of the universal access API and the low-level ACP API. It was missing from the packages exports.
This PR adds it to the publicly exported types, as well as aligning it with how our other public API types are handled, meaning it is exported from the `interfaces` module rather than being defined as part of another submodule.

- [ ] I've added a unit test to test for potential regressions of this bug. **N/A**


NSeydoux:
This should fix https://github.com/inrupt/solid-client-js/issues/1862

solid-akb (approved):
Awesome! Looks good 🙌

PR #1882, opened 2023-02-20
Convert jes configuration to TS

NSeydoux:
This aligns the jest configuration with our other repositories. It includes:
- Having TS configuration file, which allows to statically check for correctness of some configuration values
- Running the unit tests in Node and jsdom as subprojects
- Running the node end-to-end tests as a subproject

Running the tests in Node is the reason why all references to `window.fetch` have been replaced by `fetch`.


package.json
```diff
@@ -17,11 +17,13 @@
     "lint:prettier": "prettier \"{src,e2e}/**/*.{ts,tsx,js,jsx,css}\" \"**/*.{md,mdx,yml}\"",
     "list-licenses": "license-checker --production --csv --out LICENSE_DEPENDENCIES_ALL",
     "prepublishOnly": "npm run build",
-    "test": "jest",
+    "test": "jest --ignoreProjects e2e-node",
```
chelseapinka, line 20: oh so we dont want to run the e2e test cases for the test command?

NSeydoux: No, mostly because our CI uses the `test` command, it is run by dependabot. Dependabot being unable to run the end-to-end tests because it doesn't have access to the credentials. We can change the CI script to explicitly run the unit tests, but I think I'd rather keep the current behavior (which this change makes explicit), which is to only run unit tests on `npm run test`, because the end-to-end test require some environment configuration, and can be a bit slower, so I'd rather run them separately.

VirginiaBalseiro (approved):

PR #1884, opened 2023-02-21
Exp/universal fetch

NSeydoux:
Migrate `@inrupt/solid-client` to `universal-fetch`, replacing the previous dependency `cross-fetch`. This also centralizes `mockResponse` into a single file instead of it being duplicated across tests, and prevents `fetch` to be rebound so that unauthenticated fetch still works in the browser.


src/acl/acl.test.ts
```diff
@@ -792,7 +850,9 @@ describe("getSolidDatasetWithAcl", () => {
     const mockFetch = jest
       .fn<typeof fetch>()
       .mockReturnValue(
-        Promise.resolve(new Response("Not allowed", { status: 403 }))
+        Promise.resolve(
+          new Response("Not allowed", { status: 403, statusText: "Forbidden" })
```
jeswr, line 854: Why has new status text been added?

src/acl/acl.test.ts
```diff
@@ -808,7 +868,9 @@ describe("getSolidDatasetWithAcl", () => {
     const mockFetch = jest
       .fn<typeof fetch>()
       .mockReturnValue(
-        Promise.resolve(new Response("Not found", { status: 404 }))
+        Promise.resolve(
+          new Response("Not found", { status: 404, statusText: "Not Found" })
```
jeswr, line 872: Why has new status text been added?

src/acl/acl.test.ts
```diff
@@ -944,11 +1006,9 @@ describe("getFileWithAcl", () => {
           : url === "https://some.pod/"
           ? { Link: '<.acl>; rel="acl"' }
           : { "Content-Type": "text/turtle" };
-      const init: ResponseInit & { url: string } = {
-        headers,
-        url: url as string,
-      };
-      return Promise.resolve(new Response(undefined, init));
+      const response = new Response(undefined, { headers });
+      jest.spyOn(response, "url", "get").mockReturnValue(url as string);
```
jeswr, line 1010: Why are we spying on the `get` method throughout this series of tests; we don't seem to ever be testing the spyed upon function?

src/acl/acl.test.ts
```diff
@@ -997,7 +1057,9 @@ describe("getFileWithAcl", () => {
     const mockFetch = jest
       .fn<typeof fetch>()
       .mockReturnValue(
-        Promise.resolve(new Response("Not allowed", { status: 403 }))
+        Promise.resolve(
+          new Response("Not allowed", { status: 403, statusText: "Forbidden" })
```
jeswr, line 1061: new status text

src/acl/acl.test.ts
```diff
@@ -1013,7 +1075,9 @@ describe("getFileWithAcl", () => {
     const mockFetch = jest
       .fn<typeof fetch>()
       .mockReturnValue(
-        Promise.resolve(new Response("Not found", { status: 404 }))
+        Promise.resolve(
+          new Response("Not found", { status: 404, statusText: "Not Found" })
```
jeswr, line 1079: new status text

src/acl/acl.test.ts
```diff
@@ -1206,13 +1279,18 @@ describe("getResourceInfoWithAcl", () => {
...
-    const mockFetch = jest
-      .fn<typeof fetch>()
-      .mockReturnValue(Promise.resolve(mockResponse));
+    const mockFetch = jest.fn<typeof fetch>().mockReturnValue(
+      Promise.resolve(
+        mockResponse(
+          "Not allowed",
+          {
+            status: 403,
+            statusText: "Forbidden",
```
jeswr, line 1288: new status text

src/acl/acl.test.ts
```diff
@@ -1224,13 +1302,18 @@ describe("getResourceInfoWithAcl", () => {
...
-    const mockFetch = jest
-      .fn<typeof fetch>()
-      .mockReturnValue(Promise.resolve(mockResponse));
+    const mockFetch = jest.fn<typeof fetch>().mockReturnValue(
+      Promise.resolve(
+        mockResponse(
+          "Not found",
+          {
+            status: 404,
+            statusText: "Not Found",
```
jeswr, line 1311: new status text

src/acp/acp.test.ts
```diff
@@ -58,6 +57,15 @@ jest.mock("../fetcher.ts", () => ({
...
 
+function mockResponseFrom(
+  body?: BodyInit | null,
+  init?: ResponseInit,
+  url?: string
+): Response {
+  const response = new Response(body, init);
+  jest.spyOn(response, "url", "get").mockReturnValue(url ?? "");
+  return response;
+}
```
jeswr, line 68: Why can't the `mockResponse` from `../test.internal.ts` be used. I'm also still unconvinced that this is required at all; shouldn't `new Response(body, init)` suffices since I can't see any tests against the spyed on object.

src/fetcher.ts
jeswr, line 1: Is this file still required - I don't see why we cannot just replace imports from "../fetcher" with imports from `"@inrupt/universal-fetch"`. I also don't follow on why we are overriding the global fetch object as that would mean this library suddenly has sideEffects.

src/profile/webid.test.ts
```diff
@@ -170,44 +176,49 @@ describe("getProfileAll", () => {
...
+      .mockResolvedValueOnce(mockedProfileResponse)
+      .mockResolvedValueOnce(mockedAltProfileResponse);
+
     await getProfileAll(MOCK_WEBID);
-    // The embedded fetch should have been used to fetch the alt profile.
-    expect(mockedAuthFetcher.fetch).toHaveBeenCalledTimes(1);
-    expect(mockedAuthFetcher.fetch).toHaveBeenCalledWith(
-      "https://some.profile",
+    // The embedded fetch should have been used to fetch the WebID Profile first,
+    // and then the alt profile.
```
jeswr, line 195: Why was this not happening before?

src/profile/webid.test.ts
```diff
@@ -219,16 +230,19 @@ describe("getProfileAll", () => {
       profileContent
     );
     // Mock the webid unauthenticated fetch
-    const mockedUnauthFetch = jest.requireMock("cross-fetch") as {
+    const mockedUnauthFetch = jest.requireMock("../fetcher") as {
```
jeswr, line 233: Can we mock `@inrupt/universal-fetch` instead of `../fetcher`?

src/profile/webid.test.ts
```diff
@@ -425,21 +457,24 @@ const mockProfileDoc = (
 
 describe("getPodUrlAll", () => {
   it("does not use the provided fetch to dereference the WebID", async () => {
-    const mockedFetcher = jest.requireMock("cross-fetch") as {
+    const { fetch: mockedFetcher } = jest.requireMock("../fetcher") as {
```
jeswr, line 460: Can we mock `@inrupt/universal-fetch` instead of `../fetcher` (here and in all the tests that follow)

src/resource/mock.test.ts
```diff
@@ -133,14 +137,31 @@ describe("mockFetchError", () => {
...
+      "https://some.pod/resource",
+      599,
+      "Unknown error"
+    );
 
     expect(error).toBeInstanceOf(Error);
-    expect(error.statusText).toBeUndefined();
-    expect(error.message).toBe(
-      "Fetching the Resource at [https://some.pod/resource] failed: [1337] [undefined]."
+    expect(error.statusCode).toBe(599);
```
jeswr, line 147: Why change the error code - seems kind of arbitrary?

src/resource/solidDataset.test.ts
```diff
@@ -1583,7 +1603,9 @@ describe("saveSolidDatasetAt", () => {
       const mockFetch = jest
         .fn<typeof fetch>()
         .mockReturnValue(
-          Promise.resolve(new Response("Not found", { status: 404 }))
+          Promise.resolve(
+            new Response("Not found", { status: 404, statusText: "Not Found" })
```
jeswr, line 1607: This is what I would think most of our tests to look like rather than using `mockResponse`.

That said, why is the addition of `statusText` being done?

src/resource/solidDataset.ts
```diff
@@ -301,7 +302,13 @@ export async function getSolidDataset(
...
     ...options,
   };
-
+  // config.fetch = config.fetch.bind(window);
+  // await config
+  //   .fetch(
+  //     "https://openid.dev-next.inrupt.com/.well-known/openid-configuration"
+  //   )
+  //   .then((response) => response.text())
+  //   .then(console.log);
```
jeswr, line 311: Can this be deleted?

src/resource/solidDataset.ts
```diff
@@ -1160,11 +1167,9 @@ export async function getWellKnownSolid(
       new URL(urlString).origin
     ).href;
 
-    // Technically, the request here should be public and shouldn't require an
-    // authenticated fetch, however, in some environments, fetcher.ts fails to
-    // load cross-fetch sometimes, which results in this call failing if we
-    // don't pass the fetch method through:
-    return await getSolidDataset(wellKnownSolidUrl, { fetch: options.fetch });
+    return await getSolidDataset(wellKnownSolidUrl, {
```
jeswr, line 1170: Why is the comment deleted?

src/tests.internal.ts
jeswr, line 1: I'm not convinced that these utilities add value - see comments above.

jeswr (changes requested):

src/acl/acl.test.ts
NSeydoux: This should actually be just a call to `mockResponse`, fixed in daa9984c. You'll note that we don't track the reference to the spied function, here the intent is simply to mock its return value. a95495869002c2eaf54cfd3035d7c93743dc798c should clarify a bit there.

src/access/wac.test.ts
```diff
@@ -95,27 +89,36 @@ describe("getAgentAccess", () => {
...
-          status: 404,
-          url: "https://some.pod/resource.acl",
-        })
+        mockResponse(
+          "",
+          {
+            status: 404,
+          },
+          "https://some.pod/resource.acl"
+        )
```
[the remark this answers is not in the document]
NSeydoux: I've added a comment in a95495869002c2eaf54cfd3035d7c93743dc798c, hope it clears things up.

src/acl/acl.test.ts
NSeydoux: Ditto

src/acl/acl.test.ts
NSeydoux: Ditto

src/acl/acl.test.ts
NSeydoux: Ditto

src/acl/acl.test.ts
NSeydoux: Ditto

src/acp/acp.test.ts
NSeydoux: It absolutely can, I just decided to have `mockResponse` in a shared file after I wrote this and forgot to come back and use it here. Fixed in d48e1dff.

src/fetcher.ts
NSeydoux: (Nice use of the new GH feature to comment on files ^^)

I don't think this overrides the global fetch, at least that's not the intention: `uniFetch` is the global fetch, but `unifetch.bind(globalThis)` creates a new function that is then exported by this module.

The reason why we want this rebinding is because it appears the browser fetch has internal logic enforcing the `this` to implement the `window` interface. If you run `fetch("https://login.inrupt.com/.well-known/openid-configuration").then(response => response.text()).then(console.log)`  in the browser console, it works fine, but
```
const test = { fetch: fetch }
test.fetch("https://login.inrupt.com/.well-known/openid-configuration").then(response => response.text()).then(console.log)
```
fails, because the `test.fetch` `this` gets bound to `test`, while it is bound to `globalThis` in the default `fetch`.

We inject `fetch` in a lot of our functions, and if said injected fetch is the global one, it can get rebound like this, creating issues. That's the reason for the change in 6adb47bb2bf6f16e18018ca5e517edbb5f215e37 or 3420add2238afd308e38d52f28bf64e32dbdcbcf: using the global fetch failed the browser-based tests because `getSolidDataset` was used with the global fetch to get the WebID for instance.

Since we want to prevent this rebinding wherever we want to explicitly call an unauthenticated fetch, it makes sense to me to keep this file separate just to have a single source of a fetch.

src/profile/webid.test.ts
```diff
@@ -44,28 +44,33 @@ import {
...
+//       new Response(undefined, {
+//         headers: {
+//           Location: "https://arbitrary.pod/resource",
+//           "Content-Type": "text/turtle",
+//         },
+//       })
+//     )
+//   ),
+// }));
+
```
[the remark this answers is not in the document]
NSeydoux: Yup: 0fe2db29

src/profile/webid.test.ts
NSeydoux: The implementation changed a bit to move from using the global fetch to using the bound default fetch, so the mocks needed to be slightly changed.

src/profile/webid.test.ts
NSeydoux: See the comment about the `fetcher` file

src/profile/webid.test.ts
NSeydoux: See the comment about the `fetcher` file

src/resource/mock.test.ts
NSeydoux: That was a fun one: as per https://httpwg.org/specs/rfc9110.html#rfc.section.15, status codes go from 100 to 599. The former `Response` constructor didn't complain, but the new one did.

src/resource/solidDataset.test.ts
NSeydoux: Ditto for the status text, and in this case we don't need the `response.url` value being mocked.

src/resource/solidDataset.ts
NSeydoux: Woops yes 0004885b

src/resource/solidDataset.ts
NSeydoux: It no longer applies: `options.fetch` is potentially authenticated, hence the comment explaining why an authenticated fetch could be passed even though it shouldn't be necessary, but now we just pass the default fetch, which we know to not be authenticated.

package.json
```diff
@@ -182,6 +182,7 @@
     "@types/http-link-header": "^1.0.1",
     "@types/jest": "^29.2.3",
     "@types/n3": "^1.10.4",
+    "buffer": "^6.0.3",
```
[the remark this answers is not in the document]
NSeydoux: It looks unnecessary indeed, removed it.

src/access/universal_v1.test.ts
```diff
@@ -889,16 +891,21 @@ describe("setPublicAccess", () => {
...
     getResourceInfoWithAcr.mockResolvedValueOnce(mockedResourceWithAcr);
     jest.spyOn(acpModule, "internal_setPublicAccess");
 
-    await setPublicAccess("https://arbitrary.pod/resource", { read: true });
-
+    await setPublicAccess(
+      "https://arbitrary.pod/resource",
+      { read: true },
+      { fetch: jest.fn<typeof fetch>() }
+    );
```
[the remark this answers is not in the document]
NSeydoux: This has been added because of what I think are weird `jsdom` issues in 3420add2238afd308e38d52f28bf64e32dbdcbcf, and these specific tests are actually very shallow, they just check that the appropriate module is called. The actual tests are in each module, so I wouldn't be worried here about backwards compatibility.

jeswr:
I suggest we land https://github.com/inrupt/typescript-sdk-tools/pull/140/files before merging this

jest.config.ts
```diff
@@ -45,13 +45,13 @@ export default {
   projects: [
     {
       ...baseConfig,
-      displayName: "browser [Timezone: UTC]",
```
jeswr, line 48: Why the name change (change is also unrelated to the main contents of this PR).

jeswr (approved):

jeswr:
LGTM barring the minor comment.

jest.config.ts
NSeydoux: https://github.com/inrupt/solid-client-js/blob/a3558586109e2374653121e1073db97a72f16a61/package.json#L21 resulted in `npm run test:unit:browser` failing, because there was no project matching the `browser` name. This went unnoticed previously, because `npm run test` ran all unit test suites by simply excluding the e2e ones, but the change to `.github/workflows/ci.yml` resulted in each test suite running independantly. That was causing CI to fail.

PR #1901, opened 2023-03-07
chore(jsdoc): mark file apis as stable

jeswr:


chelseapinka (approved):
Looks good. I was going to ask if we needed to update the changelog about transitioning from experimental but since its just the annotation thats probably not needed.

jeswr:
I've added a brief note to the changelog - this should automerge when tests complete.

PR #1914, opened 2023-03-22
chore/migrate jsonld parser

jeswr:
Migrates from the jsonld library to jsonld-streaming-parser


src/formats/jsonLd.test.ts
```diff
@@ -20,11 +20,31 @@
...
 import { jest, describe, it, expect } from "@jest/globals";
+import { Response } from "cross-fetch";
 import { foaf, rdf } from "rdf-namespaces";
 import { DataFactory } from "n3";
-import jsonld from "jsonld";
+
+// FIXME: Move this to jest.config.ts when merging with the universal fetch PR
+import type * as RDF from "@rdfjs/types";
+import type { IQuadTerms } from "jest-rdf/lib/matchers/toBeRdfDatasetMatching";
+import matchers from "jest-rdf/lib/matchers";
```
NSeydoux, line 30: There isn't a submodule export map in the `jest-rdf` library `package.json`, so this is relying on internals and non-public API, so I would very much prefer not to introduce this. I see you've opened a related issue on the upstream repo, let's see if we can get that merged, and in the meantime let's just work around the matcher.

src/formats/jsonLd.test.ts
```diff
@@ -20,11 +20,31 @@
...
 import { getJsonLdParser } from "./jsonLd";
 
+interface Matchers<R> {
```
NSeydoux, line 33: I am not sure this is necessary. It looks in the upstream that declaring it in the jest namespace makes it available to the jest globals, but that's not the case here.

src/formats/jsonLd.ts
```diff
@@ -74,56 +48,23 @@ export const getJsonLdParser = (): Parser => {
...
-    parse: async (source, resourceInfo) => {
-      const jsonld = (await import("jsonld")).default;
+    parse: async (source, resourceInfo) =>
+      new Promise((res) => {
+        const parser = new JsonLdParser({
+          baseIRI: getSourceUrl(resourceInfo),
+          documentLoader: new FetchDocumentLoader(fetch),
+        });
+        parser.on("end", () => res());
+        parser.on("error", () => res());
```
NSeydoux, line 58: Would it make sense to have the parsing promise reject if there is a parsing error?

NSeydoux (changes requested):

src/formats/jsonLd.ts
jeswr: Usually I would say yes - but in this instance it would be a breaking change because the original parser implementation using `jsonld` also resolved regardless of whether or not there was a parsing error.

src/formats/jsonLd.test.ts
jeswr: Yes you're right - was originally doing type casting with this but replaced it with a `@ts-ignore` in the end.

src/formats/jsonLd.ts
NSeydoux: I think it would be okay to reject honestly, it feels like a bug to me that there isn't an error in case of parsing issue. What do you think?

jeswr: Well if you look at the `Parser` type that this is meant to be implementing it actually is meant to return a `void` rather than a `Promise<void>` (which is what is actually returned before and after the migration). So the expected use of this function would appear to be something like the following where the `parse` method is not even awaited

```ts
const parser = getJsonLdParser();

const quadArr = [];

await new Promise((res, rej) => {
 parser.onQuad(quadArr.push);
 parser.onError(rej);
 parser.onComplete(res)
parser.parse("My string");
});

console.log(quadArr)
```

In this case if we reject then an error is going to resolve in an uncaught reject which will end node process (in most cases unexpectedly) after 1 second then an error will be thrown

jeswr: So imagine I that this will be a breaking change to anyone who does use the parser

jeswr: But this raises another good point; we should add a test to make sure that the `onError` callbacks get triggered before the `onComplete` callbacks for cases like this.

NSeydoux: Oh right I hadn't realized `parse` wasn't an `async`. Yeah in that case rejecting isn't an option, any rejection should actually happen in a wrapping promise like https://github.com/inrupt/solid-client-js/blob/cf381af7eb0ca77853d90aa2ef49013986837eed/src/resource/solidDataset.ts#L205 does. And having a test to make sure things are called in the appropriate order definitely makes sense.

jeswr: Those tests have been added :)

src/formats/jsonLd.test.ts
```diff
@@ -86,58 +128,7 @@ describe("The Parser", () => {
...
-        },
-        object: {
-          termType: "NamedNode",
-          value: "http://xmlns.com/foaf/0.1/Person",
-        },
-        graph: { termType: "DefaultGraph", value: "" },
-      },
-    ]);
-    await parser.parse(jsonLd, {
+    await parser.parse(jsonLdInvalidLiteral, {
```
NSeydoux, line 131: ditto

src/formats/jsonLd.ts
```diff
@@ -74,56 +48,31 @@ export const getJsonLdParser = (): Parser => {
     onComplete: (callback) => {
       onCompleteCallbacks.push(callback);
     },
-    parse: async (source, resourceInfo) => {
-      const jsonld = (await import("jsonld")).default;
+    parse: async (source, resourceInfo) =>
+      new Promise((res) => {
```
NSeydoux, line 52:
```suggestion
parse: (source, resourceInfo) => {
      // eslint-disable-next-line no-void
      void new Promise<void>((res) => {
```
`parse` should be a void function, so not return a Promise. Feel free to add an appropriate comment on the linting exception.

NSeydoux (changes requested):

src/formats/jsonLd.test.ts
```diff
@@ -39,43 +97,27 @@ describe("The Parser", () => {
...
-    {
-      "@id":"https://example.com/some-path#someSubject",
-      "@type":"http://xmlns.com/foaf/0.1/Person",
-      "http://xmlns.com/foaf/0.1/name":"Some name"
-    }
-  `;
-
-    await parser.parse(jsonLd, {
+
+    await parser.parse(jsonLdPersonData, {
```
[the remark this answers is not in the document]
jeswr: The checks cannot be applied until after all the `onData` callbacks have been triggered which is why the await was originally put here.

The alternative would be to `await` the `onComplete` event but since we already test that behavior using the `stringToArray` function I think it is best to keep the test as-is for now. In particular the current setup ensures that we don't accidentally create regressions in which `onQuad` events occur after the promise resolves (even though consumers shouldn't be relying on that assumption as it is a non-documented part of the API).

src/formats/jsonLd.ts
NSeydoux: As per our offline conversation, this will not be changed in the current version, and will only be implemented later as it could be a breaking change for developers relying on undocumented behavior.

NSeydoux (approved):

PR #1934, opened 2023-04-12
 chore: remove push trigger for CodeQL

jeswr:
Resolves this workflow error https://github.com/inrupt/solid-client-js/actions/runs/4666633261/jobs/8261438119#step:7:47

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -6,9 +6,6 @@
 name: "Static Application security Testing (CodeQL)"
 
 on:
-  push:
-    branches:
-      - "*"
   pull_request:
     # The branches below must be a subset of the branches above
     branches:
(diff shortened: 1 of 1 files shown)


NSeydoux (dismissed):

.github/workflows/codeql-analysis.yml
```diff
@@ -6,9 +6,6 @@
 name: "Static Application security Testing (CodeQL)"
 
 on:
-  push:
-    branches:
-      - "*"
```
NSeydoux, line 11: Instead of deleting this, let's just ignore push events to Dependabot PRs: 
```
push:
  branches-ignore:
    - "dependabot/**"
```
(based on https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#on)

NSeydoux (changes requested):

chelseapinka (approved):

NSeydoux (approved):

PR #1942, opened 2023-04-14
Fix: ignore not found error when fetching ACL so that a new one is initialized

### LINKED ISSUE #1549: access/universal functions throw error instead of creating .acl when no .acl exists
### Search terms you've used
setAccess, setAccessFor

### Bug description
The documentation states
> If the Resource did not have an ACL yet, a new one will be initialised. This means that changes to the ACL of a parent Container can no longer affect access people have to this Resource, although existing access will be preserved.
However, in practice, when trying to setAccessFor a newly created resource with no .acl file the function throws an error because it receives a 404.

### To Reproduce

1. Call `setAccessFor(reousourceUrl, "public", { read: true }, options)` on a resource with no .acl

### Expected result

The .acl file should be initialized, and set with the requested access.

### Actual result

An error message is thrown with the following message: 
> Error: Fetching the metadata of the Resource at [https://ian.staging.mysilio.me/spaces/home/public.ttl.acl](https://ian.staging.mysilio.me/spaces/home/public.ttl.acl) failed: [404] [Not Found].

The error bubbles up from from `responseToResourceInfo` : https://github.com/inrupt/solid-client-js/blob/4b9960014a5d5e6ccaab79988feba79407167fdc/src/resource/resource.ts#L84, which seems to throw on all not `ok` responses (including 404s).

That function is used in the `getReourceInfoWithAcr` call that happens on the first line of `setPublicAccess` (called by `setAccessFor`: https://github.com/inrupt/solid-client-js/blob/4b9960014a5d5e6ccaab79988feba79407167fdc/src/access/universal_v1.ts#L457

`getResourceInfoWithAcr` calls `getResourceInfo` which calls `responseToResourceInfo` which throws. This happens on the first line of all the access setters, so I am not sure if any of them correctly implement the .acl creation described in the documentation. 

### Environment
```
System:
    OS: macOS 11.6.5
    CPU: (16) x64 Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
    Memory: 7.10 GB / 32.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.14.0 - ~/.nvm/versions/node/v16.14.0/bin/node
    npm: 8.3.1 - ~/.nvm/versions/node/v16.14.0/bin/npm
  Browsers:
    Brave Browser: 96.1.33.106
    Chrome: 99.0.4844.84
    Firefox: 98.0.2
    Safari: 15.3
  npmPackages:
    @babel/core: ^7.12.3 => 7.14.3
    @inrupt/solid-client: ^1.19.0 => 1.19.0
    @inrupt/solid-client-authn-browser: ^1.11.2 => 1.11.2
    @inrupt/vocab-common-rdf: ^1.0.3 => 1.0.3
    @inrupt/vocab-solid-common: ^0.7.5 => 0.7.5
    @size-limit/preset-small-lib: ^4.8.0 => 4.10.3
    @storybook/addon-essentials: ^6.0.28 => 6.2.9
    @storybook/addon-info: ^5.3.21 => 5.3.21
    @storybook/addon-links: ^6.0.28 => 6.1.1
    @storybook/addons: ^6.0.28 => 6.1.1
    @storybook/react: ^6.0.28 => 6.2.9
    @testing-library/react-hooks: ^3.4.2 => 3.4.2
    @types/jest: ^25.2.3 => 25.2.3
    @types/react: ^17.0.30 => 17.0.30
    @types/react-dom: ^17.0.9 => 17.0.9
    @types/url-parse: ^1.4.3 => 1.4.3
    babel-loader: ^8.2.1 => 8.2.2
    dequal: ^2.0.2 => 2.0.2
    molid: ^0.3.0 => 0.3.0
    react: ^17.0.1 => 17.0.2
    react-dom: ^17.0.2 => 17.0.2
    react-is: ^17.0.2 => 17.0.2
    react-test-renderer: ^17.0.2 => 17.0.2
    size-limit: ^4.8.0 => 4.10.3
    swr: ^1.0.1 => 1.0.1
    ts-jest: ^25.5.1 => 25.5.1
    tsdx: ^0.14.1 => 0.14.1
    tslib: ^2.3.1 => 2.3.1
    typedoc: ^0.22.6 => 0.22.6
    typescript: ^4.4.4 => 4.4.4
    url-parse: ^1.4.7 => 1.5.1
    whatwg-fetch: ^3.5.0 => 3.5.0
  npmGlobalPackages:
    corepack: 0.10.0
    npm: 8.3.1
    yalc: 1.0.0-pre.53
    ```

### PR DESCRIPTION
**Pull request by VirginiaBalseiro**
This PR fixes [#1549](https://github.com/inrupt/solid-client-js/issues/1549).

A 404 error was throwing, which was blocking the new ACL from being initialized when one was not found. In this PR, we ignore this error so that the function can resume and the ACL initialization can be handled. 

// tests pending work on setting up CSS tests

### DIFF
--- a/src/acp/acp.ts
+++ b/src/acp/acp.ts
@@ -333,11 +333,17 @@ async function fetchAcr(
     // an ACL is advertised, we can still fetch its metadata — if that indicates
     // that it's actually an ACP Access Control Resource, then we can fetch that
     // instead.
-    const aclResourceInfo = await getResourceInfo(
-      resource.internal_resourceInfo.aclUrl,
-      options
-    );
-    if (isAcr(aclResourceInfo)) {
+    let aclResourceInfo;
+    try {
+      aclResourceInfo = await getResourceInfo(
+        resource.internal_resourceInfo.aclUrl,
+        options
+      );
+    } catch (e) {
+      // if the above fails, an ACL cannot be found and we proceed.
+    }
+
+    if (aclResourceInfo && isAcr(aclResourceInfo)) {
       acrUrl = getSourceUrl(aclResourceInfo);
     }
   }
--- a/src/acp/acp.ts
+++ b/src/acp/acp.ts
@@ -340,7 +340,10 @@ async function fetchAcr(
         options
       );
     } catch (e) {
-      // if the above fails, an ACL cannot be found and we proceed.
+      // Since both ACL and ACR will be discovered through the same header, we
+      // need to ignore errors here so that in the case of ACL not found, the
+      // code can resume and a new ACL can be initialized. The case for ACR is
+      // covered in the code below, since in this case the ACR is always present
     }
 
     if (aclResourceInfo && isAcr(aclResourceInfo)) {

### CODE REVIEW

File: src/acp/acp.ts [hunk 1]
```diff
@@ -333,11 +333,17 @@ async function fetchAcr(
...
-    );
-    if (isAcr(aclResourceInfo)) {
+    let aclResourceInfo;
+    try {
+      aclResourceInfo = await getResourceInfo(
+        resource.internal_resourceInfo.aclUrl,
+        options
+      );
+    } catch (e) {
+      // if the above fails, an ACL cannot be found and we proceed.
```
NSeydoux, line 343: I think it would be helpful to expand this comment a bit, to recontextualize why that's a correct behavior, rather than describing what the code does. For instance, remind that both ACL and ACR will be discovered through the same header, and that an additional lookup is required for checking the type of the resource, and that the absence of the resource is only possible in the case of an ACL, because in the ACR system the ACR is always present.

NSeydoux (changes requested):
The code is fine, but I think more context in the comment would be helpful, to be kind to future us :p

NSeydoux (approved):
Can you break down the comment in multiple lines before merging, so that this is roughly aligned with our linting config?

VirginiaBalseiro:
> Can you break down the comment in multiple lines before merging, so that this is roughly aligned with our linting config?

Oops, thought I did. Fixed.

PR #1954, opened 2023-04-19
chore: enable node 20 in ci

jeswr:


NSeydoux (approved):
You can update the branch protection rules so that the browser e2e tests are required on node 20, and not node 18.

PR #1979, opened 2023-05-08
fix: use @types/node buffer type rather than lib.dom.ts

jeswr:
This PR fixes #<issue ID>.

As discussed in https://github.com/microsoft/TypeScript/issues/53668
the @types/node definition of a buffer is looser and hence we should
use that in order to be compaible with web buffer types and node
buffer types

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ The following changes have been implemented but not released yet:
 ### Bugfixes
 
 - `Buffer` type: As discussed in microsoft/TypeScript#53668 the
-  @types/node definition of a buffer is looser and hence we now
+  @types/node definition of a buffer is looser than the DOM one (the latter being TS' default), and hence we now
   use that in order to be compatible with web buffer types and node
   buffer types.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,13 @@ The following changes are pending, and will be applied on the next major release
 
 The following changes have been implemented but not released yet:
 
+### Bugfixes
+
+- `Buffer` type: As discussed in microsoft/TypeScript#53668 the
+  @types/node definition of a buffer is looser and hence we now
+  use that in order to be compatible with web buffer types and node
+  buffer types.
+
 ### New feature
 
 - Node 20 support
--- a/e2e/node/resource.test.ts
+++ b/e2e/node/resource.test.ts
@@ -39,6 +39,7 @@ import {
   getPodRoot,
   createFetch,
 } from "@inrupt/internal-test-env";
+import { Buffer as NodeBuffer } from "buffer";
 import {
   getSolidDataset,
   setThing,
@@ -58,7 +59,6 @@ import {
   deleteSolidDataset,
   getWellKnownSolid,
 } from "../../src/index";
-import { Buffer as NodeBuffer } from "buffer";
 
 const env = getNodeTestingEnvironment();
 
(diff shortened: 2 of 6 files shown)


src/resource/file.test.ts
```diff
@@ -376,339 +377,380 @@ describe("Write non-RDF data into a folder", () => {
...
+          fetch: jest.Mocked<typeof window.fetch>;
+        };
+
+        fetcher.fetch = setMockOnFetch(fetcher.fetch);
+
+        const mockTextBlob = new Blob(["mock blob data"]);
+        const savedFile = await saveFileInContainer(
+          "https://some.url",
+          mockTextBlob
+        );
```
NSeydoux, line 524: Should the default `mockBlob` have no `type`, and this be replaced with:
```suggestion
        const savedFile = await saveFileInContainer(
          "https://some.url",
          data
        );
```
?

If so, we should have a separate test for the `instanceof` check, to make sure the generic typing works properly.

src/resource/file.test.ts
```diff
@@ -376,339 +377,380 @@ describe("Write non-RDF data into a folder", () => {
...
+  const mockBuffer = Buffer.from("mock blob data");
+  const mockNodeBuffer = NodeBuffer.from("mock blob data");
+
+  describe.each([
+    ["blob", mockBlob],
+    ["buffer", mockBuffer],
+    ["nodeBuffer", mockNodeBuffer],
+  ] as [string, Blob | Buffer][])(
+    "blob, buffer and nodeBuffer test",
+    (_, data) => {
```
NSeydoux, line 611: ditto

e2e/node/resource.test.ts
```diff
@@ -151,6 +152,49 @@ describe("Authenticated end-to-end", () => {
...
+        new Blob(["test"]),
+        fetchOptions
+      );
+      const sessionDataset = await getSolidDataset(
+        sessionResource,
+        fetchOptions
+      );
+
+      // Eslint isn't detecting the fact that this is inside an it statement
+      // for an unknown reason
```
NSeydoux, line 188:
```suggestion
      // because of the conditional.
```

CHANGELOG.md
```diff
@@ -16,6 +16,13 @@ The following changes are pending, and will be applied on the next major release
 
 The following changes have been implemented but not released yet:
 
+### Bugfixes
+
+- `Buffer` type: As discussed in microsoft/TypeScript#53668 the
+  @types/node definition of a buffer is looser and hence we now
```
NSeydoux, line 22:
```suggestion
  @types/node definition of a buffer is looser than the DOM one (the latter being TS' default), and hence we now
```

jeswr:
Do not merge until https://github.com/inrupt/solid-client-js/pull/1982 is released

src/resource/file.test.ts
```diff
@@ -376,81 +377,164 @@ describe("Write non-RDF data into a folder", () => {
...
-      "Content-Type": "binary",
-      Slug: "someFileName",
-    });
-    expect(mockCall[1]?.body).toEqual(mockBlob);
-  });
+  describe.each([
+    ["blob", mockBlob],
+    ["buffer", mockBuffer],
+    ["nodeBuffer", mockNodeBuffer],
+  ] as [string, Blob | Buffer][])(
```
NSeydoux, line 384: I don't think the type assertion is necessary here

NSeydoux (approved):

PR #2049, opened 2023-06-26
Feat: follow rdfs:seeAlso when discovering extended WebID profiles

VirginiaBalseiro:
- update getAltProfileUrlAllFrom to prioritize following rdfs:seeAlso

- update tests

# Checklist

### DIFF
--- a/src/profile/webid.test.ts
+++ b/src/profile/webid.test.ts
@@ -33,7 +33,7 @@ import {
   setStringNoLocale,
   setThing,
 } from "..";
-import { foaf, pim, rdf } from "../constants";
+import { foaf, pim, rdf, rdfs } from "../constants";
 import { triplesToTurtle } from "../formats/turtle";
 import { toRdfJsQuads } from "../rdfjs.internal";
 import {
@@ -70,7 +70,7 @@ const MOCK_PROFILE = setThing(
 );
 
 describe("getAltProfileUrlAllFrom", () => {
-  it("returns no alt profiles if the WebID profile contains no triples with the foaf:primaryTopic/foaf:isPrimaryTopicOf predicate", async () => {
+  it("returns no alt profiles if the WebID profile contains no triples with the rdfs:seeAlso or foaf:primaryTopic/foaf:isPrimaryTopicOf predicate", async () => {
     const webIdProfile = mockSolidDatasetFrom(MOCK_WEBID);
     await expect(
       getProfileAll(MOCK_WEBID, { webIdProfile })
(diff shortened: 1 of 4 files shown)


src/profile/webid.ts
```diff
@@ -58,6 +58,7 @@ export function getAltProfileUrlAllFrom(
   const altProfileUrlAll = getThingAll(webIdProfile)
     .filter((thing) => getIriAll(thing, foaf.primaryTopic).length > 0)
     .map(asIri)
+    .concat(webIdThing ? getIriAll(webIdThing, rdfs.seeAlso) : [])
```
jeswr, line 61: Should we have some de-duplication logic here for profiles that use both `rdfs:seeAlso` and `foaf:isPrimaryTopicOf` to refer to another WebId?

src/profile/webid.test.ts
```diff
@@ -302,6 +318,53 @@ describe("getProfileAll", () => {
...
+      webIdProfile,
+    });
+    expect(result.altProfileAll).toHaveLength(2);
+    expect(getThingAll(result.altProfileAll[0])).toStrictEqual(
+      getThingAll(MOCK_PROFILE)
+    );
+    expect(getThingAll(result.altProfileAll[1])).toStrictEqual(
+      getThingAll(MOCK_PROFILE)
+    );
+  });
```
jeswr, line 366: Perhaps add another test where both `seeAlso` and `isPrimaryTopicOf` are both used on the same WebId; and check the results array is returned in the expected order?

jeswr (approved):

NSeydoux (approved):
Can you add a changelog entry describing this change?

src/profile/webid.ts
NSeydoux: Sounds like a good idea, e.g. reducing into a Set.

VirginiaBalseiro: If I am not misunderstanding the comment, that is already done in L66.

src/profile/webid.test.ts
VirginiaBalseiro: Do you mean checking that duplicates are removed?

jeswr: Yep :)

jeswr: Ignore previous reply - I thought I was responding to the above discussion.

My understanding of the ticket is that `rdfs:seeAlso` should be prioritized over `foaf:isPrimaryTopicOf` and so this should be reflected in the order of the objects in the array returned. We should have a couple of checks for this in the tests to avoid regressions.

src/profile/webid.ts
jeswr: Ah - my bad!

PR #2059, opened 2023-06-30
Enforce URL path semantics in containment lookup

NSeydoux:
When listing contained resources with `getContainedResourcesAll`, in addition to using `ldp:contains` claims, we apply a filter so that we only return resources that are child resources of the target container from a URL path semantics perspective as well. This means `[https://pod.example.org/foo/bar/moo`](https://pod.example.org/foo/bar/moo%60) cannot be considered a child resource of `[https://pod.example.org/foo/`](https://pod.example.org/foo/%60). Resources from a different origin will also be excluded.

This prevents malicious Pods from claiming containment across domains, which could lead to incorrect behavior of the clients.

### DIFF
--- a/src/resource/solidDataset.ts
+++ b/src/resource/solidDataset.ts
@@ -869,7 +869,7 @@ export function getContainedResourceUrlAll(
     // See https://solidproject.org/TR/protocol#resource-containment
     if (!isChildResource(childUrl, containerUrl)) {
       throw new Error(
-        `Container [${containerUrl}] claims containment of [${childUrl}], but this is not possible according to slash semantics. The containment triple should be removed from [${containerUrl}].`
+        `Container [${containerUrl}] claims containment of [${childUrl}], but this is not possible according to slash semantics (https://solidproject.org/TR/protocol#resource-containment). The containment triple should be removed from [${containerUrl}].`
       );
     }
   });
--- a/src/resource/solidDataset.ts
+++ b/src/resource/solidDataset.ts
@@ -834,7 +834,7 @@ function isChildResource(a: string, b: string): boolean {
   const urlDiff = child.href.substring(parent.href.length, child.href.length);
   // The child path component that isn't present in the parent should only
   // potentially include slashes at the end (if it is a container).
-  const isDirectChild = urlDiff.includes(urlDiff.replace("/", " ").trim());
+  const isDirectChild = urlDiff.includes(urlDiff.replace(/\//g, " ").trim());
   return isAncestor && isDirectChild;
 }
 
--- a/src/resource/solidDataset.ts
+++ b/src/resource/solidDataset.ts
@@ -832,15 +832,12 @@ function isChildResource(a: string, b: string): boolean {
   const child = new URL(a);
   // Explicitly test on the whole URL to enforce similar origins.
   const isAncestor = child.href.startsWith(parent.href);
-  const urlDiff = child.pathname.substring(
-    parent.pathname.length,
-    child.pathname.length
-  );
+  const relativePath = child.pathname
+    .substring(parent.pathname.length, child.pathname.length)
+    .replace(/(^\/)|(\/$)/g, "");
   // The child path component that isn't present in the parent should only
   // potentially include slashes at the end (if it is a container).
-  const isDirectChild =
-    urlDiff !== "" && urlDiff.includes(urlDiff.replace(/\//g, " ").trim());
-  return isAncestor && isDirectChild;
+  return isAncestor && relativePath.length >= 1 && !relativePath.includes("/");
 }
 
 /**
--- a/src/resource/solidDataset.ts
+++ b/src/resource/solidDataset.ts
@@ -844,6 +844,8 @@ function isChildResource(a: string, b: string): boolean {
  * Given a [[SolidDataset]] representing a Container (see [[isContainer]]), fetch the URLs of all
  * contained resources.
  * If the solidDataset given is not a container, or is missing resourceInfo, throw an error.
+ * If the containment of some resources is invalid (see {@link validateContainedResourceAll}),
+ * they are not included in the result.
  *
  * @param solidDataset The container from which to fetch all contained Resource URLs.
  * @returns A list of URLs, each of which points to a contained Resource of the given SolidDataset.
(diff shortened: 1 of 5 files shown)


src/resource/solidDataset.ts
```diff
@@ -844,7 +855,11 @@ export function getContainedResourceUrlAll(
   // See https://www.w3.org/TR/2015/REC-ldp-20150226/#h-ldpc-http_post:
   // > a containment triple MUST be added to the state of the LDPC whose subject is the LDPC URI,
   // > whose predicate is ldp:contains and whose object is the URI for the newly created document
-  return container !== null ? getIriAll(container, ldp.contains) : [];
+  return container !== null
+    ? getIriAll(container, ldp.contains).filter((child) =>
+        isChildResource(child, getSourceUrl(solidDataset))
+      )
+    : [];
```
jeswr, line 862: To me it seems more sensible to do the following (factor out the `getSrcUrl` for performance; and error rather than filtering so that users receive errors to alert them to a malicious Pod rather than results just going missing).

```suggestion
  if (container === null) return [];
  const containment = getIriAll(container, ldp.contains);
  // Make a const outside the loop so we only compute this once
  const srcUrl = getSourceUrl(solidDataset);
  
  for (const child of containment) {
    if (!isChildResource(child, srcUrl))
      // Error rather than just hiding the resource
      throw new Error(`Child resource [${child}] not in container [${srcUrl}] according to slash semantics`);
  }
```

src/resource/solidDataset.ts
```diff
@@ -827,6 +827,17 @@ export async function deleteContainer(
...
 
+function isChildResource(a: string, b: string): boolean {
+  const parent = new URL(b);
+  const child = new URL(a);
+  const isAncestor = child.href.startsWith(parent.href);
+  const urlDiff = child.href.substring(parent.href.length, child.href.length);
+  // The child path component that isn't present in the parent should only
+  // potentially include slashes at the end (if it is a container).
+  const isDirectChild = urlDiff.includes(urlDiff.replace("/", " ").trim());
+  return isAncestor && isDirectChild;
```
jeswr, line 838: The following returns `true` even though I wouldn't expect it to `isChildResource('http://a/b//c/d/e', 'http://a/b/')`.

To fix this I would do something equivalent to
```suggestion
  const stripped = urlDiff.replace(/(^\/)|(\/$)/g, "");
  return isAncestor && stripped.length >= 1 && !stripped.includes("/");
```

jeswr (approved):

jeswr (changes requested):
Meant to request changes rather than approve

src/resource/solidDataset.ts
jeswr: I'd also be inclined to do this all based off `.path` rather than `.href` and check that all the other `URL` attributes are the same in order to prevent things like `http://example.org/a?q1=a/a` being considered to be contained in `http://example.org/a?q1=a/`

NSeydoux: b55a497ff4775001106d5eceebb1a4b3a770f833 addresses your first point (I didn't realize `.replace` only replaced the first occurrence for non-regexes), and 15dc7f566ece765ff36e87f40ab2bef9762142ee your second point, both very good observations.

src/resource/solidDataset.ts
NSeydoux: Agreed, it sounds like throwing here would be a good idea, in particular since there is a direct action the user can take to mitigate the issue: 2e7bebf27d68f188b43ef873cc42bcb874c9c37a

src/resource/solidDataset.ts
```diff
@@ -840,11 +856,24 @@ export async function deleteContainer(
...
   // See https://www.w3.org/TR/2015/REC-ldp-20150226/#h-ldpc-http_post:
   // > a containment triple MUST be added to the state of the LDPC whose subject is the LDPC URI,
   // > whose predicate is ldp:contains and whose object is the URI for the newly created document
-  return container !== null ? getIriAll(container, ldp.contains) : [];
+  const childrenUrlAll = getIriAll(container, ldp.contains);
+  childrenUrlAll.forEach((childUrl) => {
+    // See https://solidproject.org/TR/protocol#resource-containment
+    if (!isChildResource(childUrl, containerUrl)) {
+      throw new Error(
+        `Container [${containerUrl}] claims containment of [${childUrl}], but this is not possible according to slash semantics. The containment triple should be removed from [${containerUrl}].`
```
jeswr, line 872:
```suggestion
        `Container [${containerUrl}] claims containment of [${childUrl}], but this is not possible according to slash semantics (https://solidproject.org/TR/protocol#resource-containment). The containment triple should be removed from [${containerUrl}].`
```

src/resource/solidDataset.ts
```diff
@@ -827,6 +827,22 @@ export async function deleteContainer(
...
+  // Explicitly test on the whole URL to enforce similar origins.
+  const isAncestor = child.href.startsWith(parent.href);
+  const urlDiff = child.pathname.substring(
+    parent.pathname.length,
+    child.pathname.length
+  );
+  // The child path component that isn't present in the parent should only
+  // potentially include slashes at the end (if it is a container).
+  const isDirectChild =
+    urlDiff !== "" && urlDiff.includes(urlDiff.replace(/\//g, " ").trim());
```
jeswr, line 842: I still find this less intuitive and more buggy than https://github.com/inrupt/solid-client-js/pull/2059#discussion_r1248579674.

In particular; it looks like `isDirectChild` will incorrectly return true when `urlDiff` is the following values `/`, `//`, `///` etc; `/ /a//`. You could even do

```ts
if (urlDiff.startsWith('/')) urlDiff = urlDiff.slice(1)
if (urlDiff.endsWith('/')) urlDiff = urlDiff.slice(0, -1)
isDirectChild = urlDiff !== "" && !urlDiff.includes("/")
```

Secondly should we check to make sure that `search` attribute of both `URL`s is the empty string.

NSeydoux: You're right, it is cleaner, implemented in 3b5aa3c26786c23715c614c509c303fb08ad98f7.

To your second point, I don't know if the specification says anything about search params. There probably won't be search params in most cases, but if the spec doesn't explicitly prevent it let's not make assumptions.

NSeydoux:
Note that there is a discussion in https://github.com/inrupt/solid-client-java/pull/571 about whether we should throw or not, I'd like to have consistency across SDKs for such behaviors.

jeswr (approved):

PR #2097, opened 2023-07-26
chore: extend validateContainedResourceAll docs to reference slash semantics

jeswr:


src/resource/solidDataset.ts
```diff
@@ -844,8 +844,9 @@ function isChildResource(a: string, b: string): boolean {
  * Given a [[SolidDataset]] representing a Container (see [[isContainer]]), fetch the URLs of all
  * contained resources.
  * If the solidDataset given is not a container, or is missing resourceInfo, throw an error.
- * If the containment of some resources is invalid (see {@link validateContainedResourceAll}),
- * they are not included in the result.
+ * If the containment of some resources is invalid according to
```
kay-kim, line 847: Can we move this up closer to the first sentence?
So ... 

> Given {@link SolidDataset} representing a Container (see  {@link isContainer}), fetch the URLs of all contained resources with the following exception: if the containment of some resources is invalid according to [slash semantics](https://solidproject.org/TR/protocol#uri-slash-semantics) (see {@link validateContainedResourceAll}), they are not included in the result.

src/resource/solidDataset.ts
```diff
@@ -871,14 +872,31 @@ export function getContainedResourceUrlAll(
...
  * Resources for which containment is invalid are not included in the result set returned by
  * {@link getContainedResourceUrlAll}.
  *
+ * The ESS will always include containment triples that respect
+ * [slash semantics](https://solidproject.org/TR/protocol#uri-slash-semantics); and so
+ * this function will always return true for containers fetched from the ESS.
+ *
+ * It is recommended that this function *always* be used before reading the containment triples
+ * from a Pod using {@link getContainedResourceUrlAll} in order to detect unexpected behaviour,
+ * including malicious containment triples, when interacting with other Solid servers.
```
kay-kim, line 898: I'd flip this sentence to lead with When interacting with other Solid servers, ...

src/resource/solidDataset.ts
kay-kim: Alternatively, you can go

> Given {@link SolidDataset} representing a Container (see {@link isContainer}), fetch the URLs of all contained resources that respect slash semantics (see {@link validateContainedResourceAll}).

src/resource/solidDataset.ts
```diff
@@ -871,14 +872,31 @@ export function getContainedResourceUrlAll(
...
+ *
+ * The following resources are invalid:
+ *  - https://example.org/container/resource/invalid (not a direct child resource)
+ *  - https://example.org/container2 (not a child resource)
+ *  - https://domain2.example.org/container/resource (not a direct child resource)
  *
  * Resources for which containment is invalid are not included in the result set returned by
  * {@link getContainedResourceUrlAll}.
  *
+ * The ESS will always include containment triples that respect
```
kay-kim, line 892: Do we want to finesse this to make it less like we're just  ESS-centric vs we're supporting Solid-compliant servers?  Like

> ESS, conforming to the Solid Protocol,  respects [slash semantics](https://solidproject.org/TR/protocol#uri-slash-semantics) for its containment triples.  As such, the function returns true for containers fetched from ESS.

src/resource/solidDataset.ts
jeswr: Ah - this isn't quite what I meant; I'm saying *always* use it (unconditionally), is the following clearer?
```
 * It is recommended that this function *always* be used before reading the containment triples
 * from a Pod using {@link getContainedResourceUrlAll}. This will allow you do detect unexpected behaviour from servers,
 * including malicious containment triples that could appear when interacting with other Solid servers.
 ```

kay-kim: Ah ... if that's the case, I'd move up that up a bit so that we have 

> Given a SolidDataset ..., verify that all its containment claims are valid. Containment claims are valid if it respects slash semantics. 
> 
> For example, given container at https://example.org/container/:
> 
> - The following resources are valid:
> 
> - The following resources are invalid:
> 
> If a component claim is invalid, validateContainedResourceAll returns the invalid component's URL as part of its return object.
> 
> Note: It is recommended that this function always be used before calling
> {@link getContainedResourceUrlAll} since {@link getContainedResourceUrlAll} does not
> return Resources for which containment is invalid. Using the function in conjunction with {@link getContainedResourceUrlAll} allows for the detection of unexpected behaviour from servers,
> including malicious containment triples that could appear. Because ESS conforms to the Solid Protocol,i.e., respects slash semantics for its containment triples, `validateContainedResourceAll` returns true for containers fetched from ESS.


This way, we state

- What this function does, including explanation of what's invalid, and what it does return
- Recommendation: Use it with getContainedResourceUrlAll and why
- and the ESS statement becomes more of a throwaway.

kay-kim (approved):
LGTM!

chelseapinka (approved):

PR #2117, opened 2023-08-04
test: getContainedResourceUrlAll ignores path traversal children

Otto-AA:
Adds a check for path traversal with `getContainedResourceUrlAll` to ensure future modifications don't miss this edge case.


NSeydoux:
Superseded by  #2119

PR #2283, opened 2023-12-13
E2E test application-defined request metadata

NSeydoux:
This adds end-to-end tests for application-defined request metadata. The tests scenarios are a combination of authenticated and unauthenticated requests, with success and error responses.

### DIFF
--- a/e2e/node/request-metadata.test.ts
+++ b/e2e/node/request-metadata.test.ts
@@ -40,13 +40,6 @@ import { getSolidDataset } from "../../src";
 const { Headers } = global;
 
 const env = getNodeTestingEnvironment();
-if (
-  env.features?.APPLICATION_DEFINED_REQUEST_METADATA === "false" ||
-  env.features?.APPLICATION_DEFINED_REQUEST_METADATA === ""
-) {
-  // eslint-disable-next-line jest/no-focused-tests, jest/expect-expect
-  test.only(`Skipping unsupported application-defined request metadata tests in ${env.environment}`, () => {});
-}
 
 if (
   env.features?.APPLICATION_DEFINED_REQUEST_METADATA === "true" &&
@@ -264,4 +257,8 @@ if (
       });
     },
   );
+} else {
+  // eslint-disable-next-line jest/expect-expect, jest/no-focused-tests
+  test.only(`Skipping unsupported application-defined request metadata tests in ${env.environment}`, () => {});
+
 }
--- a/e2e/node/request-metadata.test.ts
+++ b/e2e/node/request-metadata.test.ts
@@ -1,0 +1,265 @@
+//
+// Copyright Inrupt Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal in
+// the Software without restriction, including without limitation the rights to use,
+// copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the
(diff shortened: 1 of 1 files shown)


e2e/node/request-metadata.test.ts
```diff
@@ -0,0 +1,265 @@
...
+        ...init?.headers,
+        ...requestHeaders,
+      },
+    });
+    responseHeaders(response.headers);
+    return response;
+  };
+};
+
+describe.each(headers)(
```
jeswr, line 102: Jest expects an array of arrays here where the nth element of each element array corresponds to the nth argument of the callback function

```suggestion
describe.each(headers.map(header => [header]))(
```

e2e/node/request-metadata.test.ts
```diff
@@ -0,0 +1,265 @@
...
+          responseHeaders = headers;
+        };
+        const spiedFetch = jest.spyOn(global, "fetch");
+        const customFetch = customizeHeaders(
+          fetch,
+          requestHeaders,
+          readHeaders,
+        );
+
+        // The response will be a 401
```
chelseapinka, line 208: if this is a `resolves.not.toThrow` is that actually a 401? this comment seems right but the validator seems wrong.

e2e/node/request-metadata.test.ts
NSeydoux: The mapping is implicit for 1-D arrays according to the docs: https://jestjs.io/docs/api#1-describeeachtablename-fn-timeout

e2e/node/request-metadata.test.ts
NSeydoux: `solid-client` acts as a high-level client, and as such it will throw on 401 or 403.

e2e/node/request-metadata.test.ts
```diff
@@ -0,0 +1,263 @@
...
+
+import type { Session } from "@inrupt/solid-client-authn-node";
+import {
+  getNodeTestingEnvironment,
+  getPodRoot,
+  getAuthenticatedSession,
+} from "@inrupt/internal-test-env";
+import { getSolidDataset } from "../../src";
+
+const { Headers } = global;
```
jeswr, line 40:
```suggestion
```

e2e/node/request-metadata.test.ts
```diff
@@ -0,0 +1,263 @@
...
+          ).toContain(Object.values(returnedHeaders)[0]);
+          expect(responseHeaders.get("non-returned-header")).toBeNull();
+        });
+
+        it("can read back predefined headers set by the client on a failed request", async () => {
+          let responseHeaders: Headers = new Headers();
+          const readHeaders = (headers: Headers) => {
+            responseHeaders = headers;
+          };
+          const spiedFetch = jest.spyOn(global, "fetch");
```
jeswr, line 229:
```suggestion
          const spiedFetch = jest.spyOn(globalThis, "fetch");
```
I suggest we replace all instances of global with globalThis (which works in node and browser)

NSeydoux: ecf655c

e2e/node/request-metadata.test.ts
NSeydoux: Weird, for some reason when I was testing I had a `Headers is not defined` error, which was fixed by this, but I can't reproduce now even with this gone. Removed in ecf655c.

jeswr (approved):

PR #2292, opened 2023-12-19
Breaking: drop buffer support

jeswr:
- Drop support for `Buffer`s as input to `saveFileInContainer` and `overwriteFile`. Use the file name of inputs to `saveFileInContainer` as the `Slug` name recommendation if no slug is explicitly provided.

### DIFF
--- a/e2e/node/resource.test.ts
+++ b/e2e/node/resource.test.ts
@@ -163,7 +163,6 @@ describe("Authenticated end-to-end", () => {
     await deleteFile(fileUrl, fetchOptions);
   });
 
-  // Blob is only available globally Node 18 and above
   it("can create, delete, and differentiate between RDF and non-RDF Resources using a Blob", async () => {
     const fileUrl = `${sessionResource}.txt`;
 
--- a/e2e/node/resource.test.ts
+++ b/e2e/node/resource.test.ts
@@ -150,6 +150,8 @@ describe("Authenticated end-to-end", () => {
       // We need to type cast because the buffer definition
       // of Blob does not have the prototype property expected
       // by the lib.dom.ts
+      // See https://github.com/microsoft/TypeScript/issues/53668
+      // and https://github.com/microsoft/TypeScript/issues/52166
       new NodeBlob(["test"], {
         type: "text/plain",
       }) as unknown as globalThis.Blob,
--- a/e2e/node/resource.test.ts
+++ b/e2e/node/resource.test.ts
@@ -72,6 +72,9 @@ if (env.environment === "NSS") {
 
 const TEST_SLUG = "solid-client-test-e2e-resource";
 
+const nodeVersion = process.versions.node.split(".");
+const nodeMajor = Number(nodeVersion[0]);
+
 describe("Authenticated end-to-end", () => {
   let fetchOptions: { fetch: typeof fetch };
   let session: Session;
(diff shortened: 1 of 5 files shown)


e2e/node/resource.test.ts
```diff
@@ -146,28 +142,17 @@ describe("Authenticated end-to-end", () => {
...
-  it("can create, delete, and differentiate between RDF and non-RDF Resources using a node Buffer", async () => {
+  it("can create, delete, and differentiate between RDF and non-RDF Resources using a Blob from the node Buffer package", async () => {
     const fileUrl = `${sessionResource}.txt`;
 
     const sessionFile = await overwriteFile(
       fileUrl,
-      NodeBuffer.from("test"),
+      // We need to type cast because the buffer definition
+      // of Blob does not have the prototype property expected
+      // by the lib.dom.ts
```
NSeydoux, line 152: If I remember correctly, you had opened an issue about this a while ago. Do you happen to have the link to it?

e2e/node/resource.test.ts
```diff
@@ -178,59 +163,32 @@ describe("Authenticated end-to-end", () => {
     await deleteFile(fileUrl, fetchOptions);
   });
 
-  it("can create, delete, and differentiate between RDF and non-RDF Resources using a Blob from the node Buffer package", async () => {
+  // Blob is only available globally Node 18 and above
```
NSeydoux, line 166:
```suggestion
```

e2e/node/resource.test.ts
```diff
@@ -359,18 +317,15 @@ describe("Authenticated end-to-end", () => {
...
+      const response = await fetchOptions.fetch(args[0], {
+        ...args[1],
         headers: {
-          ...init?.headers,
+          ...args[1]?.headers,
           "User-Agent": "some-user-agent",
         },
       });
-      if (info.toString() === sessionResource) {
+      if (args[0].toString() === sessionResource) {
```
NSeydoux, line 328: Why this change? I find the previous version to be more readable.

src/resource/file.ts
```diff
@@ -19,9 +19,9 @@
 // SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 //
 
-import type { Buffer, File as NodeFile } from "buffer";
+import type { File as NodeFile } from "buffer";
 import type {
-  File,
+  File as BlobFile,
```
NSeydoux, line 24: We have to remove that alias eventually, it's really confusing ^^. When Node 18 goes away, that's going with it.

src/resource/file.ts
```diff
@@ -200,23 +201,8 @@ type SaveFileOptions = WriteFileOptions & {
...
-export async function saveFileInContainer<
-  FileExt extends File | NodeFile | Buffer,
->(
-  folderUrl: Url | UrlString,
-  file: FileExt,
-  options?: Partial<SaveFileOptions>,
-): Promise<FileExt & WithResourceInfo>;
 export async function saveFileInContainer<
-  FileExt extends File | NodeFile | Buffer,
+  FileExt extends File | BlobFile | NodeFile,
```
NSeydoux, line 205: What is the first `File` here?

jeswr: That's the one globally defined by typescript

jeswr: Even though the constructor doesn't work till mode 20, the type seems to exist in node 18

e2e/node/resource.test.ts
jeswr:
```suggestion
      // We need to type cast because the buffer definition
      // of Blob does not have the prototype property expected
      // by the lib.dom.ts
      // See https://github.com/microsoft/TypeScript/issues/53668
      // and https://github.com/microsoft/TypeScript/issues/52166
```

@NSeydoux could you leave a comment in https://github.com/microsoft/TypeScript/issues/53668 to see if we can get this re-opened?

NSeydoux (approved):

e2e/node/resource.test.ts
jeswr: My editor was complaining (see below) but this doesn't actually seem to be an issue in CI so I've reverted the changes.

image

PR #2294, opened 2023-12-19
chore: remove deprecated APIs

jeswr:
- Removes deprecated APIs; and all ESS 1.1 APIs (even if not marked as deprecated - see https://github.com/inrupt/solid-client-js/pull/2294/commits/23ab6f5ab2cbb1de072b5d6ada38273c2e8f4f15).
 - Makes parsers return `void` rather than `Promise<void>` in order to comply with type definition.

There are still a few deprecated functions in [src/acp/matcher](https://github.com/inrupt/solid-client-js/blob/breaking/remove-deprecated-apis/src/acp/matcher.ts) and [src/acp/rule](https://github.com/inrupt/solid-client-js/blob/breaking/remove-deprecated-apis/src/acp/rule.ts) that I am not sure whether should be kept or removed at the moment.


e2e/node/acp.test.ts
```diff
@@ -129,12 +124,8 @@ describe("An ACP Solid server", () => {
     );
   });
 
-  it("advertises its ACLs as ACP AccessControlResources", async () => {
-    expect(await acp.isAcpControlled(sessionResource, fetchOptions)).toBe(true);
-  });
```
NSeydoux, line 134: Is this something that doesn't have an equivalent in the latest API?

NSeydoux (approved):

e2e/node/acp.test.ts
jeswr: Indeed it does; added back in in https://github.com/inrupt/solid-client-js/pull/2294/commits/60d976a2ca5598c0b2b1772557a6ac49ffbb4d9b.

PR #2307, opened 2024-01-16
Move tsconfig to Node16 module interop

NSeydoux:


NSeydoux:
This was an experiment, will roll out this change in all libraries simultaneously.

Issue #2339, opened 2024-02-25
Policy under unnamed node is not returned from `getResourcePolicyAll()`


renyuneyun:
### Search terms you've used

- ACP
- getResourcePolicyAll
- getPolicyUrlAll

### Bug description

When following [the document](https://docs.inrupt.com/developer-tools/javascript/client-libraries/tutorial/manage-acp/#view-policies-and-matchers-for-a-resource) to get policies, it seems only policies with named nodes are returned from `acp_ess_2.getResourcePolicyAll()`. This looks like a bug rather than intentional choice, as apparently `.acr` documents can contain unnamed `acp:Policy` nodes.

### To Reproduce

1. Create a `.acr` with unnamed policy nodes
2. Call `acp_ess_2.getResourcePolicyAll()` for that resource

**Minimal reproduction**

Example `.acr` file: https://gist.github.com/renyuneyun/834eb5ee542e06a2cc3ee1a57712ca3f

### Expected result

All `acp:Policy` nodes are returned in some form.

### Actual result

Only named nodes (of `acp:Policy`) are returned.

### Environment

```
  System:
    OS: Linux 6.6 Arch Linux
    CPU: (8) x64 11th Gen Intel(R) Core(TM) i5-1135G7 @ 2.40GHz
    Memory: 1.90 GB / 15.40 GB
    Container: Yes
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 21.6.1 - ~/node_modules/.bin/node
    Yarn: 1.22.21 - /usr/bin/yarn
    npm: 10.4.0 - /usr/bin/npm
  Browsers:
    Chromium: 122.0.6261.57
  npmPackages:
    @comunica/query-sparql: ^2.8.1 => 2.10.0 
    @inrupt/solid-client: ^2.0.0 => 2.0.0 
    @inrupt/solid-client-authn-browser: ^2.0.0 => 2.0.0 
    @inrupt/vocab-common-rdf: ^1.0.5 => 1.0.5 
    @inrupt/vocab-solid: ^1.0.4 => 1.0.4 
    @mdi/font: ^7.1.96 => 7.2.96 
    @renyuneyun/solid-helper: ^0.0.3 => 0.0.3 
    @rushstack/eslint-patch: ^1.1.4 => 1.3.2 
    @tsconfig/node20: ^20.1.0 => 20.1.0 
    @types/n3: ^1.10.4 => 1.16.0 
[...]

PR #2368, opened 2024-03-27
Stop nesting blank nodes in internal structure

NSeydoux:
This changes the internal representation of RDF datasets so that Blank Nodes subjects are no longer nested, even if they are part of a single predicate chain. This approach was initially chosen to avoid generating unstable identifiers for these Blank Nodes, but it prevents browsing the potential Blank Nodes subjects that are part of a single chain when querying a Dataset. In particular, the library relying on immutability, a BN without an identifier cannot be used as the subject of a new quad.

This is work that will help resolving #2339. 

# Checklist

- [ ] Relevant documentation, if any, has been written/updated. N/A, this is an internal change


src/formats/solidDatasetAsTurtle.test.ts
```diff
@@ -33,10 +33,11 @@ async function getDataset(ttl: string): Promise<SolidDataset> {
...
 const ttl = `
-  prefix : <#>
-  prefix ex: <https://example.org/>
-  prefix foaf: <http://xmlns.com/foaf/0.1/>
-  prefix vcard: <http://www.w3.org/2006/vcard/ns#>
+  @prefix : <#> .
+  @prefix ex: <https://example.org/> .
+  @prefix foaf: <http://xmlns.com/foaf/0.1/> .
+  @prefix vcard: <http://www.w3.org/2006/vcard/ns#> .
+  @base <https://example.org/> .
```
NSeydoux, line 40: This change introduces the `@base` directive, because the URl resolution was incorrect.

src/rdf.test.ts
```diff
@@ -212,7 +214,7 @@ describe("fromRdfJsDataset", () => {
     expect(fromRdfJsDataset(rdfJsDataset)).toStrictEqual({
       type: "Dataset",
       graphs: {
-        default: {
+        default: expect.objectContaining({
```
NSeydoux, line 217: We no longer do exact matches on the internal representation, because of the unstable blank nodes identifiers.

src/rdf.test.ts
```diff
@@ -231,41 +233,35 @@ describe("fromRdfJsDataset", () => {
...
+      { scope: acrGraphIriString },
+    );
+    const subjectsIncludingBlankNodes = getThingAll(
+      fromRdfJsDataset(rdfJsDataset),
+      { scope: acrGraphIriString, acceptBlankNodes: true },
+    );
+    // There should be two blank nodes in the resulting dataset.
+    expect(
+      subjectsIncludingBlankNodes.length - subjectsExcludingBlankNodes.length,
+    ).toBe(2);
```
NSeydoux, line 264: This replaces the previous exact match of the internal structure

src/rdfjs.internal.ts
NSeydoux, line 1: The changes in this file are centered around removing the code that nested blank nodes.

src/resource/solidDataset.ts
NSeydoux, line 1: There used to be two separate implementations of the conversion of an RDFJS dataset into a `SolidDataset`. This refactors the code so that it is implemented only once.

src/rdf.test.ts
```diff
@@ -231,41 +233,35 @@ describe("fromRdfJsDataset", () => {
...
+                  expect.stringMatching(/_:/),
                 ],
               },
             },
           },
-        },
+        }),
       },
     });
+    const subjectsExcludingBlankNodes = getThingAll(
```
chelseapinka, line 253: so am I understanding correct that the representation would have have the blank nodes linked off other legit nodes, if you have `acceptBlankNodes: false` would you still expect those Blank nodes linked or would they just be gone entirely?

chelseapinka: nvm answered belowed :point_down:

chelseapinka (approved):

PR #2457, opened 2024-08-14
Adjust dependabot configuration

acoburn:
The dependabot configuration is [currently invalid](https://github.com/inrupt/solid-client-js/runs/28673463034), showing the following error:

> Update configs must have a unique combination of 'package-ecosystem', 'directory', and 'target-branch'. Ecosystem 'npm' has overlapping directories.

This removes the overlapping configuration, though it does mean that dependabot will no longer distinguish between dev and prod dependencies.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,6 +1,6 @@
 version: 2
 updates:
-  # Enable weekly version updates for prod dependencies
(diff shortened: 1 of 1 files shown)


edwardsph (approved):
LGTM

PR #2526, opened 2024-10-14
SDK-3368: Feature flag problem details tests

NSeydoux:
The end-to-end tests need to run against older supported versions of ESS, not all of which support problem details.

By default, only the response status and title are checked, which are provided by default against any HTTP response. Problem details and instance are only checked if the environment variable E2E_TEST_FEATURE_PROBLEM_DETAILS is set to true.

### DIFF
--- a/e2e/node/resource.test.ts
+++ b/e2e/node/resource.test.ts
@@ -19,6 +19,10 @@
 // SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 //
 
+// Assertions are made conditionally on problem details responses because not all
+// servers support this feature.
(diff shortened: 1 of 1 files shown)


davidboweninrupt:
Don't you want this to be the other way round @NSeydoux ? If you demand that it's set to true where the feature exists you are going to need to carry this feature flag for ever more and callers will always have to set it.

ofc this is tempered by the opposing force that we should not break existing code when adding something.

NSeydoux:
Exactly, I went with the non-breaking approach. When we stop supporting versions that do not have the problem details feature, we can remove the check altogether.

PR #2530, opened 2024-10-15
fix: expose parserOptions more

### LINKED ISSUE #2529: Expose ParserOption on more places
### Search terms you've used

- ParserOption
- getSolidDatasetWithAcl

### Bug description

It is currently not possible to add parser options when getting a solid dataset with acl.
This is problematic when the resource is not text/turtle but something like application/trig.

### Expected result

`getSolidDatasetWithAcl` returns the expected dataset with acl options without crashing. 

### Actual result

`getSolidDatasetWithAcl` crashes with a parse error.

## Additional information

Note that I don't expect the solid-client to handle graphs correctly, as I understand it no graphs are introduced anywhere.
My use case is to use this function to get and update the acl files for a trig resource.

ajuvercr:
This PR fixes #2529.

### DIFF
--- a/src/acl/acl.internal.ts
+++ b/src/acl/acl.internal.ts
@@ -23,14 +23,14 @@ import type { Quad } from "@rdfjs/types";
 import { getSolidDataset } from "../resource/solidDataset";
 import type {
   IriString,
-  WithChangeLog,
   Thing,
+  WithChangeLog,
   WithServerResourceInfo,
 } from "../interfaces";
 import {
-  getSourceUrl,
   getResourceInfo,
   getSourceIri,
+  getSourceUrl,
 } from "../resource/resource";
 import { acl, rdf } from "../constants";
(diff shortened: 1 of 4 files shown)


NSeydoux:
As far as I understand, this does not achieve the intended effect: the introduced `ParserOptions` from `n3` is different from the `ParseOptions` in the `getSolidDataset` options (see https://github.com/inrupt/solid-client-js/blob/main/src/resource/solidDataset.ts#L153). This PR adds a new type to the interface, but the passed option will never actually be used, it's just not incompatible with the existing API because it is an additive change.

I think switching to the `solid-client` `ParseOption`, and adding that to the API where you were adding `ParserOption` should get you where you in the right direction.

ajuvercr:
Thanks @NSeydoux for having a look!
I totally didn't see the difference between ParseOptions and ParserOptions, it should be all fixed and correct now

ajuvercr:
Could you have another look @NSeydoux ?

